PR #4, opened 2015-10-27
fix sql filter by column error handling

marcin-ptaszynski:
makeColumn on nil property can cause panic.


nati:
We need followings for this commit
- Detailed commit message
- Github issue
- UT fix

nati:
I'll fix this by myself

Issue #13, opened 2015-10-31
Support to set config parameter from env variable 


nati:
Support to set config parameter from env variable

Apply go template with environment parameter.
like this.

address: {{ .GOHAN_IP}}:{{ .GOHAN_PORT}}

PR #23, opened 2015-11-05
Support to use Environment value in configuraion

nati:
In this commit, we support to use environment value in the configruaion
so that we can resue base configraion for each environment.

You can apply go based template for configuraion by this commit
address: {{ .GOHAN_IP}}:{{ .GOHAN_PORT}}

Implements issue: #13


yudai:
LGTM

Issue #14, opened 2015-10-31
Support PATCH


nati:
Some applications such as Keystone uses PATCH request. 

We should be able to support PATCH request.
https://tools.ietf.org/html/rfc5789

PR #26, opened 2015-11-06
Support PATCH request

nati:
Some REST API are using PATCH request.
In this commit, we support PATCH.

Implements issue: #14


yudai:
I feel uncomfortable that PUT and PATCH behave in the same manner. PUT should accept creating a resource?

nati:
Yes. we should accept to create resource when it isn't exist according to RFC.
I put new issue -> https://github.com/cloudwan/gohan/issues/27

yudai:
got it. Then LGTM :+1:

Issue #30, opened 2015-11-09
etcd watch process should handle 401 error


nati:
etcd watch process should handle 401 error.

What's happening in our env.

(1) resources more than 500 rows
(2) boot gohan
(3) state watch process will lock for each
(4) etcd stores up to 1000 history, so we lost some history and got 401 error

PR #31, opened 2015-11-09
Fix issue when we got 401 from etcd

nati:
Closes: issue #30

Etcd can hold up to 1000 history, so we need to handle this case
correctly.
We have following issues.

(1) We should clone env for each extension run by sync
(2) sync.Watch API keep blocking even if we get 401.
(3) Stop lock per resource
  In case we have more than 1000 resources, sync process do infinite
  loop because it increases etcd index by lock.

In this commit, we fix these issues,  and we added simple integration
test configuraion file.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -128,6 +128,10 @@
 			"Rev": "1aedb1f9e106db5d7403222313a0504a06def5b3"
 		},
 		{
+			"ImportPath": "github.com/streamrail/concurrent-map",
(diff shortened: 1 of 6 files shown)


yudai:
LGTM

MaciekPytel:
In case of index cleared error shouldn't there be a diff between sync and db state versions to run extensions for resources with changed state version (but not for other resources)?

nati:
I believe we can't ensure anything in case of index cleared error because we may face 
 out-of-sync + index cleared same time.
We need full sync in that case also

PR #36, opened 2015-11-11
Update wercker.yml to install gb

leifmadsen:


leifmadsen:
A little early :)  I can only test wercker when I make a pull request here, but it failed. Still tweaking :)

PR #39, opened 2015-11-13
Update README and remove unneeded Godeps folder

leifmadsen:
Update README documentation, and remove Godeps folder.

This does not update the information in the development section. I will update that in a separate branch once everything is completed for issue #28.


nati:
Thanks!
it is super nice to have more good writings in README also :+1:

Issue #45, opened 2015-11-19
Parent relationship order matters


leifmadsen:
When defining a schema, if you wish to add a parent relationship later, the order matters when the GUI is being created.

After creating a new schema `registration` where the parent object was `device`, the ability to add a new `registration` entry to the `device was not possible.

When the parent object was added, it was added as the last object in the object list for that schema. Once I used the schema builder to drag the `device_id` value to the top of the list, then I had the ability to add entries to the `registration` schema.

It appears the `propertiesOrder` is what changed to cause the GUI to provide the input mechanisms.

*Working*
```json
properties:
  auth_pass:
    description: Authorization Password
    permission:
      - create
      - update
    title: Auth Pass
    type: string
  auth_user:
    description: Authorization User
    permission:
      - create
      - update
    title: Auth User
    type: string
  callerid_name:
    description: CallerID Name
    permission:
      - create
      - update
    title: CID Name
    type: string
  callerid_num:
    description: CallerID Number
    permission:
      - create
      - update
    title: CID Number
    type: string
  device_id:
    description: parent object
    permission:
      - create
      - update
    relation: device
    title: device
    type: string
    unique: false
  id:
    description: ID
    permission:
      - create
    title: ID
    type: string
    view:
      - detail
  label:
    description: Line Label Display
    permission:
      - create
      - update
    title: Label
    type: string
  name:
    description: Name
    permission:
      - create
      - update
    title: Name
    type: string
  tenant_id:
    description: Tenant ID
    permission:
      - create
    title: Tenant ID
    type: string
    view:
      - detail
propertiesOrder:
  - device_id
  - tenant_id
  - id
  - name
  - auth_user
  - auth_pass
  - label
  - callerid_name
  - callerid_num
required:
  - device_id
type: object
```

nati:
I think this is related with #44 but we need order schema based on dependency relationship

leifmadsen:
@nati ah interesting. I'm not sure I totally follow. I think the ordering is definitely needed overall. It seems like there might be a code enhancement that would allow stepping through looking for relationships at a higher level, and then sort them for display or something.

Unfortunately I'm not that familiar with the code to know if what I'm saying is even a valid thought :)

nati:
yes. 
I think we should reorder schemas when Gohan loads it based on the dependency.
also, we should detect circular dependencies.

leifmadsen:
:+1:

Issue #44, opened 2015-11-19
Deleting schema with data linked to a parent causes crash


leifmadsen:
When deleting a schema that has data associated with it, which has a parent object, the FOREIGN KEY constraint fails, and causes gohan to crash.

Log:

```
13:08:07.777 gohan.schema INFO  Loading schema ./provisioning_schema.yaml ...
13:08:07.779 gohan.schema WARNING  Overwriting schema device_type
13:08:07.779 gohan.schema WARNING  Overwriting schema server
13:08:07.780 gohan.schema WARNING  Overwriting schema customer
13:08:07.781 gohan.schema WARNING  Overwriting schema device
13:08:07.782 gohan.db CRITICAL  Error during deleting table:FOREIGN KEY constraint failed
```

It does seem to delete the schema and remove it from the YAML file, however the database constraint fails, and causes the crash to happen.

nati:
@leifmadsen you are using drop_on_create, right?

leifmadsen:
@nati yes sir, currently using `drop_on_create` (I'm working on schema creation right now in the dev environment)

nati:
@leifmadsen  Thanks! I'll take a look

leifmadsen:
Thanks!

nati:
your welcome!

PR #48, opened 2015-11-24
Reorder schemas by dependency on load time if it is DAG

nati:
In this commit, we sort schemas by dependency using Tarjan's algorithm,
so that users don't need to sort it manually.

Fixes bug: #44
Fixes bug: #45

### DIFF
--- a/schema/manager.go
+++ b/schema/manager.go
@@ -136,6 +136,68 @@ func (manager *Manager) OrderedSchemas() []*Schema {
 	return res
 }
 
+//reorderSchema tries reorder schemas using Tarjan's algorithm
+type state int
+
+const (
+	notVisited state = iota
+	visited
+	temporaryVisited
+)
+
+func visitSchema(schemas []*Schema, schemaList []string, schema *Schema, visitedSchema map[string]state) ([]string, error) {
+	if visitedSchema[schema.ID] == temporaryVisited {
+		return nil, fmt.Errorf("Schemas aren't DAG. We can't reorder automatically")
+	}
+	if visitedSchema[schema.ID] == visited {
+		return schemaList, nil
+	}
+	visitedSchema[schema.ID] = temporaryVisited
+	relatedSchemas := schema.relatedSchemas()
+	var err error
+	for _, relatedSchemaID := range relatedSchemas {
+		for _, candidate := range schemas {
+			if candidate.ID == relatedSchemaID {
+				schemaList, err = visitSchema(schemas, schemaList, candidate, visitedSchema)
+				if err != nil {
+					return nil, err
+				}
+				break
+			}
+		}
+		if err != nil {
+			return nil, err
+		}
+	}
+	visitedSchema[schema.ID] = visited
+	schemaList = append(schemaList, schema.ID)
+	return schemaList, nil
+}
+
+func reorderSchemas(schemas []*Schema) ([]string, error) {
+	var err error
+	schemaList := []string{}
+	visitedSchema := map[string]state{}
+	for _, schema := range schemas {
+		visitedSchema[schema.ID] = notVisited
+	}
+	for {
+		var schema *Schema
+		for _, s := range schemas {
+			if visitedSchema[s.ID] == notVisited {
+				schema = s
+			}
+		}
+		if schema == nil {
+			return schemaList, nil
+		}
+		schemaList, err = visitSchema(schemas, schemaList, schema, visitedSchema)
+		if err != nil {
+			return nil, err
+		}
+	}
+}
+
 //Policies gets policies from manager
 func (manager *Manager) Policies() []*Policy {
 	return manager.policies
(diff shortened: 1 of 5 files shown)


schema/manager.go
```diff
@@ -136,6 +136,68 @@ func (manager *Manager) OrderedSchemas() []*Schema {
...
+		return nil, fmt.Errorf("Schemas aren't DAG. We can't reorder automatically")
+	}
+	if visitedSchema[schema.ID] == visited {
+		return schemaList, nil
+	}
+	visitedSchema[schema.ID] = temporaryVisited
+	relatedSchemas := schema.relatedSchemas()
+	var err error
+	for _, relatedSchemaID := range relatedSchemas {
+		for _, relatedSchema := range schemas {
```
yudai: Naming `relatedSchema`  here is a bit confusing, because there is "relatedSchamas" just above.
Simply `schema` could be better.

schema/manager.go
```diff
@@ -136,6 +136,68 @@ func (manager *Manager) OrderedSchemas() []*Schema {
...
+		}
+	}
+	visitedSchema[schema.ID] = visited
+	schemaList = append(schemaList, schema.ID)
+	return schemaList, nil
+}
+
+func reorderSchemas(schemas []*Schema) ([]string, error) {
+	var err error
+	schemaList := []string{}
```
yudai: `make([]string, 0, len(schemas))` could be better to save cost of append() called later.

schema/manager.go
```diff
@@ -136,6 +136,68 @@ func (manager *Manager) OrderedSchemas() []*Schema {
...
+		for _, relatedSchema := range schemas {
+			if relatedSchema.ID == relatedSchemaID {
+				schemaList, err = visitSchema(schemas, schemaList, relatedSchema, visitedSchema)
+				if err != nil {
+					return nil, err
+				}
+				break
+			}
+		}
+		if err != nil {
```
yudai: Is this needed? no function call that retunes err above in the for loop

schema/manager.go
```diff
@@ -136,6 +136,68 @@ func (manager *Manager) OrderedSchemas() []*Schema {
...
+	visited
+	temporaryVisited
+)
+
+func visitSchema(schemas []*Schema, schemaList []string, schema *Schema, visitedSchema map[string]state) ([]string, error) {
+	if visitedSchema[schema.ID] == temporaryVisited {
+		return nil, fmt.Errorf("Schemas aren't DAG. We can't reorder automatically")
+	}
+	if visitedSchema[schema.ID] == visited {
+		return schemaList, nil
```
yudai: I'm not sure, but to reject loops, this should return an error?

schema/manager.go
```diff
@@ -136,6 +136,68 @@ func (manager *Manager) OrderedSchemas() []*Schema {
...
+			if visitedSchema[s.ID] == notVisited {
+				schema = s
+			}
+		}
+		if schema == nil {
+			return schemaList, nil
+		}
+		schemaList, err = visitSchema(schemas, schemaList, schema, visitedSchema)
+		if err != nil {
+			return nil, err
```
yudai: Let me make it clear that we don't care the situation that a subset of schemas is a DAG?

schema/manager.go
nati: yeah. but we have already "schema" on argument of this func.

schema/manager.go
nati: done

schema/manager.go
nati: done

schema/manager.go
nati: This is from Tarjan algorithm
https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm

schema/manager.go
nati: Not this time. 
But it's interesting topic.
We may decouple schemas by subgraph using flood fill algorithm. then we can do topological sort for each sub-graph if it don't have loop.
i'll add new issue. ( but low priority )

schema/manager.go
yudai: ah, I see, then `candidate` or something might better.

schema/manager.go
yudai: :+1:

schema/manager.go
yudai: :+1:

schema/manager.go
yudai: ok

schema/manager.go
yudai: The page looks nothing for alrady-maked nodes. I'm interesting in what would happen if we have a loop in a schema. Could you add a test case for detecting loops?

nati: It will be reached.
Code is here
```
function visit(node n)
    if n has a temporary mark then stop (not a DAG)
    if n is not marked (i.e. has not been visited yet) then
```

We will mark each node in three color In Tarjan algorithm.

white: not visited
gray: temporary visited 
black: visited

Note that gray color will be only painted for node where we are checking.
so it is not a DAG if you meet that color again.
If you meet black color, it is no problem because that means that node is already checked

https://commons.wikimedia.org/wiki/File:Tarjan%27s_Algorithm_Animation.gif

schema/manager.go
nati: done

schema/manager.go
yudai: ok, now I got it :+1:

yudai:
:+1:

Issue #8, opened 2015-10-31
Support conditions by property value in policy


nati:
Support conditions by property status value in policy

``` yaml
policy:
  - action: 'read'
    condition:
    - property:
        status:
        - ACTIVE
        - CREATE_IN_PROGRESS
        - UPDATE_IN_PROGRESS
        - DELETE_IN_PROGRESS
        - ERROR
    effect: allow
    id: member
    principal: Member
  - action: 'update'
    condition:
    - property:
        status:  
           ACTIVE: ERROR
    effect: allow
    id: member
    principal: Member
  - action: 'reboot'
    condition:
    - property:
        status: ACTIVE
    effect: allow
    id: member
    principal: Member
  - action: 'delete'
    condition:
    - property:
        status: 
        - ACTIVE
        - ERROR
    effect: allow
    id: member
    principal: Member
```

PR #53, opened 2015-12-03
Implement property based condition

nati:
We implement property based condition in this commit.
You can limit action based on current resource property value per each Role.
see more doc update

Closes: #8

### DIFF
--- a/docs/source/policy.rst
+++ b/docs/source/policy.rst
@@ -20,7 +20,7 @@ Policy has following properties.
 Conditions
 ----------
 
-Gohan supports two types of conditions
+Gohan supports several types of conditions
 
 - :code:`is_owner` - Gohan will enforce access privileges for the resources
   specified in the policy. By default access to resources of all other tenants
@@ -85,3 +85,51 @@ Example policy
     resource:
       path: /v2.0/server/?$
 
+- :code: type `property` - You can add condition based on resource value.
+  You can specify allowed values in match.
+  if it is a value, we check exact match.
+  if it is a list, we check if the value is in the list
+  if it is a dict, we check if we have a key for this value and, updated value matches it.
+  Note that this is only valid for update action.
+
+.. code-block:: yaml
+    policy:
+      - action: 'read'
+        condition:
+        - type: property:
+          match:
+            status:
+              - ACTIVE
+              - CREATE_IN_PROGRESS
+              - UPDATE_IN_PROGRESS
+              - DELETE_IN_PROGRESS
+              - ERROR
+        effect: allow
+        id: member
+        principal: Member
+      - action: 'update'
+        condition:
+        - property:
+            status:
+              ACTIVE:
+              - UPDATE_IN_PROGRESS
+              - ERROR
+        effect: allow
+        id: member
+        principal: Member
+      - action: 'reboot'
+        condition:
+        - property:
+            status: ACTIVE
+        effect: allow
+        id: member
+        principal: Member
+      - action: 'delete'
+        condition:
+        - property:
+            status:
+            - ACTIVE
+            - ERROR
+        effect: allow
+        id: member
+        principal: Member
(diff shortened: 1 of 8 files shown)


docs/source/policy.rst
```diff
@@ -85,3 +85,49 @@ Example policy
...
+              - DELETE_IN_PROGRESS
+              - ERROR
+        effect: allow
+        id: member
+        principal: Member
+      - action: 'update'
+        condition:
+        - property:
+            status:  
+              ACTIVE: ERROR # you can update a resource with ACTIVE state & you can can change to ERROR value only
```
pedro-r-marques: nit: use "and" rather than &

schema/policy_test.go
```diff
@@ -308,19 +308,102 @@ var _ = Describe("Policies", func() {
...
+					},
+				}
+				policy, _ = NewPolicy(testPolicy)
+				err := policy.Check("create", &authorization, data)
+				Expect(err).ToNot(HaveOccurred())
+			})
+		})
+
+		Describe("Property based condition", func() {
+			It("should work with string condition based on propery", func() {
```
pedro-r-marques: spelling: property.

schema/policy.go
```diff
@@ -413,6 +430,59 @@ func (p *Policy) AddTenantToFilter(action string, tenant Tenant) {
 	p.actionTenantFilter[action] = append(p.actionTenantFilter[action], tenant)
 }
 
+// AddPropertyConditionFilter adds property based filter for action
+func (p *Policy) AddPropertyConditionFilter(action string, match map[string]interface{}) {
+	p.actionPropertyConditionFilter[action] = append(p.actionPropertyConditionFilter[action], match)
+}
+
+// ApplyPropertyConditionFilter applies filter based on Property
+func (p *Policy) ApplyPropertyConditionFilter(action string, data map[string]interface{}, next map[string]interface{}) error {
```
pedro-r-marques: It is unclear to me what the `next` argument represents and how it is used. It seems that most of the times it is not used... only on update ? could you have just called this same method for the map of properties that you are going to update ? why is next applied to map but not to list or item cases.

docs/source/policy.rst
nati: done

schema/policy_test.go
nati: done

schema/policy.go
nati: I've changed var name and added comment around here.


// ApplyPropertyConditionFilter applies filter based on Property
// You need to pass candidate update value in updateCandidateData on update API, so
// that we can limit allowed update value.
// Let's say we would like to only allow to update from ACTIVE to ERROR on an API.
// We can define this policy like this.
//
//   - action: 'update'
//     condition:
//     - property:
//         status:
//            ACTIVE: ERROR
//     effect: allow
//     id: member
//     principal: Member
//
// This policy check error in case of followings
// - Original value isn't ACTIVE
// - Update candidate value isn't ERROR

server/resources/resource_management.go
```diff
@@ -632,11 +656,26 @@ func DeleteResourceInTransaction(context middleware.Context, resourceSchema *sch
...
+	if err != nil {
+		return err
+	}
+	if resource != nil {
+		context["resource"] = resource.Data()
+	}
+	// apply property filter
+	err = policy.ApplyPropertyConditionFilter(schema.ActionUpdate, resource.Data(), nil)
+	if err != nil {
+		return ResourceError{err, "", Unauthorized}
```
pedro-r-marques: Does anything need to happen to "resource" in case of error ? Should you use a "defer" statement after resource is assigned ? I don't have enough context to know whether the Fetch operation starts a transaction or just gets you a handle to a transaction that is in progress...

nati: DB connection will be closed in another defer.  We return HTTP error after this function returns ResourceError, and the context vars are subject to be GC.

pedro-r-marques:
LGTM

PR #65, opened 2015-12-17
Reduce docker consumption of go.

jbking:
There is a bit large layer in docker repository of just moving go. This can be easier to reduce.

```
% docker history jbking/gohan
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
...
9483a3b0eb46        About an hour ago   /bin/sh -c mv /go /usr/local/go                 229.4 MB
a698a39ac63f        About an hour ago   /bin/sh -c tar xzvf go$GO_VERSION.linux-amd64   229.4 MB
...
```


nati:
Thanks! 
May be, we can also use base docker image for go too

Issue #64, opened 2015-12-12
Broken schema handling


pax0r:
When I've made a schema where in one of the fields `type` is missing Gohan exit with error:
`
panic: runtime error: invalid memory address or nil pointer dereference
`
which is not very helpful.

I think Gohan should print some more meaningful error like "property type is missing from field x on schema y". In current situation on complicated schema is may take quite a while to find where the bug was.

nati:
Hi pax0r

I supposed this should be checked by Json schema validation.
Could you share your schema ?

pax0r:
I fixed it already (I'll try to reproduce this later). Basically it was a schema with one resource which missed "type" on one of the fields.

nati:
fixed

PR #71, opened 2016-01-06
Fix error handling on loading schema

nati:
Close: issue #64


famao:
It looks fine for me,
Outputs from gohan contain useful information like below
~~~
00:05:12.521 gohan.server CRITICAL  invalid schema: Json validation error:
	type: type is required,
~~~

PR #77, opened 2016-01-28
Add environment variables for verbosity and output-format; fix crash in environment parsing

### LINKED ISSUE #78: client crashes on empty/unset environment variables
Errors from environment variable parsing are ignored, causing the client to crash when no/invalid values are provided

Problematic cases:
  One of `GOHAN_{REGION,SCHEMA_URL,SERVICE_NAME}` unset or empty
  One of `GOHAN_{OUTPUT_FORMAT,VERBOSITY}` (from #77) having an invalid value

Testing for the problem:
```bash
cmd=(gohan client event list)
for i in GOHAN_{CACHE_{PATH,SCHEMAS,TIMEOUT},REGION,SCHEMA_URL,SERVICE_NAME}
do
    echo "--------- Unset $i:"
    env -u $i $cmd
    echo
    echo "--------- $i"='"":'
    env $i="" $cmd
    echo
done

# Variables introduced in #77
for i in GOHAN_{VERBOSITY,OUTPUT_FORMAT}
do
    echo "--------- Unset $i:"
    env -u $i $cmd
    echo
    echo "--------- $i"='<null>:'
    env $i='<null>' $cmd
    echo
    echo "--------- $i=wrong:"
    env $i=wrong $cmd
    echo
done
```

vozhyk-:
The second commit fixes #78.


cli/client/opts.go
```diff
@@ -129,5 +137,41 @@ func NewOptsFromEnv() (*GohanClientCLIOpts, error) {
...
+	return "", fmt.Errorf(incorrectOutputFormat, outputFormats)
+}
+
+func parseLogLevel(verbosityOpt interface{}) (logging.Level, error) {
+	for i, logLevel := range logLevels {
+		if fmt.Sprint(i) == verbosityOpt {
+			return logLevel, nil
+		}
+	}
+	return 0, fmt.Errorf(incorrectVerbosityLevel, 0, len(logLevels)-1)
```
kwapik: I think that instead of `0` we should return `logging.WARNING`

docs/source/commands.rst
```diff
@@ -62,7 +62,7 @@ Configuration
 Gohan CLI Client is configured though environment variables. Configuration can be divided into few sections.
 
 You can use default configuraion for client with:
-:code:`source ./etc/gohan_client.rc`
+:code:`source ./etc/gohan_cli.rc`
```
kwapik: Instead of changing docs we should rename the file, because Gohan CLI is not only client but also server, schema validation etc.

cli/client/opts.go
vozhyk-: I'll change this, but it shouldn't matter, since callers shouldn't use the return value in case of an error. But if they should be able to, shouldn't `findOutputFormat()` return a default value (e.g. "table") as well?

vozhyk-:
Updated - addressed kwapik's comments.

nati:
Could you rebase 3 commit for single one?

vozhyk-:
@nati Done.

nati:
Thanks! LGTM

PR #83, opened 2016-02-10
test commit

nati:


nati:
For testing CI & coverall. Don't review this.

PR #92, opened 2016-02-26
support text/plain data

yokoi-h:
- send text/plain data from gohan_http if the content-type is "text/plain"


yokoi-h:
thank you for your comment. I fixed it.

nati:
LGTM but could you rebase multiple commit for one?

nati:
could you also add unit test for this?
I think we can run web server in the test.

yokoi-h:
Thank you.
I'll add test case and rebase.

PR #94, opened 2016-03-07
support text/plain data

yokoi-h:
added following code
- to post text/plain data
- test cases for posting json object and plain text

This PR was renewed from https://github.com/cloudwan/gohan/pull/92 because it had multiple commits and didn't have test cases.


nati:
thanks

Issue #32, opened 2015-11-09
export swagger api def


nati:
export swagger api def
http://swagger.io/

PR #96, opened 2016-03-10
Support swagger file generation from schema

nati:
Closes: issue #32


jbking:
Does this support schema actions?

nati:
Hi 
I've added action support

docs/source/schema.rst
```diff
@@ -663,4 +663,31 @@ The default for unspecified action is repeatble read.
...
+You can generate OpenAPI / Swagger file from gohan schema, so that you can
+afford swagger utility tools.
+
+.. code-block:: shell
+
+    gohan template --config-file etc/gohan.yaml
+
+    # or you can customize template file using
+
+    gohan template --config-file etc/gohan.yaml --template etc/templates/swagger.tmpl
```
jbking: Found a bug https://github.com/cloudwan/gohan/issues/97

etc/templates/swagger.tmpl
```diff
@@ -0,0 +1,229 @@
...
+                        "description": "unexpected error",
+                        "schema": {
+                        "$ref": "#/definitions/errorModel"
+                        }
+                    }
+                }
+            }
+        }{% for action in schema.Actions %},
+        "{{ schema.GetPluralURL() }}{{ action.Path | swagger_path }}": {
+            "post" : {
```
jbking: Do you think `"{{ action.Method|lower }}"`?

docs/source/schema.rst
nati: ok i'll fix this in different patch.

etc/templates/swagger.tmpl
nati: got it

nati:
ok it looks like there is no more comment on this.
I'll merge this.

PR #106, opened 2016-03-24
Example and doc fix about gohan script fibonacci function

morrisson:


nati:
sorry I've fixed it in another branch..

PR #107, opened 2016-03-24
make loopvar configurable

nati:


morrisson:
LGTM.

I have one comment unrelated to this PR itself. Is there any constraint for a variable name? For example, can variable name start with character other than alphabet [^a-zA-Z]?

nati:
Thanks  it has no restriction for now.

PR #128, opened 2016-03-30
context.schema should be available in state/monitoring update

morrisson:


nati:
LGTM waiting CI get ok

PR #136, opened 2016-04-01
Separate schema files should work

morrisson:
When abstract schema is written in a separate file, reading schemas
process terminates in failure. Fix bug and tests to try read
multiple schema files.


nati:
LGTM Thanks!

PR #151, opened 2016-04-12
FakeKeystone should return internalURL and publicURL

morrisson:
Keystone response of getting token should have internalURL and
publicURL values in serviceCatalog/endpoints, but before this commit
FakeKeystone did not support them.

TODO if someone wants: admin/internal/publicURL to be configurable


nati:
Thanks I agree with it

PR #172, opened 2016-04-18
Fix bug that config version in sync is always zero

### LINKED ISSUE #171: Fail to update config_version to sync (etcd)
syncEvent function in gohan/server/sync.go should update "version" for config if that resource is set as 'state_versioning'. Config_version in DB is updated properly, but "version" value in sync layer is always 0. 

I investigated this issue, then I found that casting "version" property in int64 type failed, therefore version is assigned as 0 (default value). I found that this error can be fixed by changing cast type to 'int'. 'int' type might be different bit-size according to server's arch, but I think this is allowable because it is hard to imagine config version becomes more than 2^32.

morrisson:
fix #171

### DIFF
--- a/server/sync.go
+++ b/server/sync.go
@@ -189,7 +189,10 @@ func (server *Server) syncEvent(resource *schema.Resource) error {
 	path := resource.Get("path").(string)
 	path = configPrefix + path
 	body := resource.Get("body").(string)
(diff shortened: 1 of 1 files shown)


morrisson:
Sorry, this code does not pass test cases in ubuntu env, so I will check what happens now.

PR #179, opened 2016-04-25
Enhance gohan validate to handle multiple schema files

### LINKED ISSUE #178: gohan validate cannot validate schema with multiple files
gohan validate -d <schema_file> can only one schema file. Therefore, if schema files are separated into A and B, and some resource in a file B depends on other in a file A, validation of a file B fails.

One of the problem is current 'd' flag of gohan validate is StringFlag, therefore only one value is accepted. I propose this should be changed to StringSliceFlag to accept multiple values for -d.

morrisson:
fix #178


nati:
Code looks good, but could you rebase this branch for master?

PR #186, opened 2016-04-27
Hide implementation detail of return value of sync.Fetch()

yudai:
sync.Fetch() returns interface{} that directly has implementation
specific struct of the sync layer (*etcd.Response). This forces the
users of this method to
handle the return value in the implementation specific way and makes it
hard to support other
sync layers.
This commit introduces a new struct `Node` that abstracts the data in
sync layers. sync.Fetch() returns abstracted data using this struct.


nati:
LGTM

PR #187, opened 2016-04-27
Manage versioning of monitoring updates

yudai:
Add `version` field to the JSON format of monitoring update so that
Gohan can ignore updates from monitoring agents which still have old
configuration values.
This situation can happen when the monitoring agents receive requests
for stopping/starting monitoring in an async manner.


nati:
Could you add unit tests?

yudai:
Oops, sorry. Added.

nati:
Thanks LGTM

PR #196, opened 2016-05-02
Allow empty input for custom actions

yudai:
When you want a simple GET action without any parameters, you should be able to omit `input` field of the schema.

Btw, I could not found test for custom actions. They are missing now?


nati:
OK here is test. could you add UT for this case ? https://github.com/cloudwan/gohan/blob/master/server/server_test.go#L954

yudai:
Thanks. Updated.

PR #199, opened 2016-05-04
Remove post_in_transaction and post events from custom actions

yudai:
There are inconsistency between the document and the actual
implementation. In the document, custom actions are described as supporting
pre, pre_in_transaction, post_in_transaction and post events.
However, the actual implementation has only post_in_transaction
and post events. In addition to that, post_in_transaction is a bit broken because the main
event (action itself) is not included in the same transaction.
This commit simply removes those events anyway and move the main event
into the transaction. We can re-implement the other events in a proper way when requested.

### DIFF
--- a/docs/source/extension.rst
+++ b/docs/source/extension.rst
@@ -131,6 +131,11 @@ update data in db without informing etcd
 
 delete data in db
 
+- gohan_db_transaction()
+
+start a new DB transaction. You are responsible for managing tranansactions created by this function. Call .Close() or .Commit() after using the return value.
+
+
 - gohan_model_list(context, schema_id, filter)
 
   Retrieve data through Gohan.
--- a/docs/source/schema.rst
+++ b/docs/source/schema.rst
@@ -623,23 +623,11 @@ Then, register extension to handle it, e.g.
 
 .. code-block:: javascript
 
-  gohan_register_handler("pre_action_reboot_in_transaction", function(context){
-    // retrieve necessary details regarding vm from db
-  });
-
   gohan_register_handler("action_reboot", function(context){
     // handle reboot in southbound
   });
 
 In order to query above action, POST to /v1.0/servers/:id/action with
-
-Available events
-
-- pre_$action
-- pre_$action_in_transaction
-- $action
-- post_$action_in_transaction
-- post_$action
 
 .. code-block:: json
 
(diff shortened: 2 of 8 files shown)


nati:
Do you mind if I ask fix this part correct way?

yudai:
Not at all! Please do it !

yudai:
Oh, btw, how much time do you need? If it will be the next week, I'd appreciate if I could merge my changes for now.

nati:
sorry, may be my English wasn't so clear.. 
Could you fix correct way instead of partial patch?

yudai:
ah, sorry. I didn't read your comment properly.

I have a question on this functionality. Do we need the events around the main event for custom actions? Because custom actions do not have "default" behaviors provided by Gohan.
I think that the benefit of events like pre|post, pre|post_in_transaction is that we can wrap around the default Gohan procedure like show, list, create and delete. However, for custom actions, all logic will be implemented in extensions on each resource type, so I'm wondering if it makes sense to implement this functionality.

nati:
your comment makes sense for me. 
I think we can have two solution.

(1) Use one event
so action name will be event name. 
In this case, we need support for starting new transaction from extension. Gohanscript has already support this, so we need this for Javascript.

(2) Use pre_$action, $action_in_transaction, post_$action

I prefer fix (1) since not all of custom action need transaction..

yudai:
The commit makes the main event handler wrapped by `InTransaction`. So you can use tx in the main event handler by default. Do you want make tx disabled by default?

nati:
yes. I think transaction should be started in the extension in this case.

yudai:
I see. In that case, writing tests with transactions could be a bit hard. Because we cannot pass `MockTransaction` by putting it to `context.transaction` and call `GohanTrigger()` with it.
Providing a transaction by default could be a better solution...?

yudai:
If we provide something like `gohan_transaction()`, users could use it in other places like `post_transaction` and make thing messed up. I'm getting feeling like extension initiated transactions are bit dangerous.

nati:
Yeah, that why I didn't added it. but now I changed mind that is needed because we can use it in single script code.  so I would say it's programmers responsibility now.

For testing, we can use 
https://github.com/DATA-DOG/go-sqlmock.

yudai:
OK, added a new commit with a new function `gohan_db_transaction()`. You can inject a mock transaction by `gohan_db_transaction.Expect()` like `gohan_http`.

nati:
Awesome. GJ

PR #202, opened 2016-05-04
Add tests for allowing any json type for output

yudai:
Removing `response_code = 200`, because it generates CustomException and
prevent the events following getting invoked.

(Feel like if `200` is set to `response_code`, we should skip generating CustomException?)


nati:
LGTM waiting CI

Yes. I agree... using CustomException for setting response code is ugly..

PR #206, opened 2016-05-05
Exit with proper exit code

yudai:
The exit code returned by `go test` was simply ignored and `exit` in the
script was used without the parameter.
This commit makes the script to exit immediately with the exit code of
`go test` when it's is larger than `0`.


yudai:
The test of this PR should fail.

yudai:
It failed. Looks good.

nati:
Good catch!

PR #207, opened 2016-05-06
Allow gohan_db_list to order, limit and offset

yudai:
Adding optional parameters that set order, limit and offset values.
The change is backward compatible


nati:
could you rebase branch?

nati:
DOC need to be updated also

yudai:
Done

nati:
coverage down?

PR #226, opened 2016-05-19
Fix bugs of SQL JOIN and column generation

### LINKED ISSUE #215: Resource cannot have two relations to the same resource if it has another relational resource
Example schema:
```
schemas:

- id: gakko
  description: Gakko
  singular: gakko
  plural: gakkos
  title: Gakko
  prefix: /v1.0
  schema:
    properties:
      id:
        description: The ID of Gakko
        format: uuid
        title: ID
        type: string
      name:
        description: Name
        title: Name
        type: string

- id: kodomo
  description: Kodomo
  singular: kodomo
  plural: kodomos
  title: Kodomo
  prefix: /v1.0
  schema:
    properties:
      id:
        description: The ID of Kodomo
        format: uuid
        title: ID
        type: string
      gakko_id:
        description: Gakko
        title: Gakko
        type: string
        relation: gakko
        relation_property: gakko

- id: oya
  description: Oya
  singular: oya
  plural: oyas
  title: Oya
  prefix: /v1.0
  schema:
    properties:
      id:
        description: The ID of Oya
        format: uuid
        title: ID
        type: string
      otoko_id:
        description: Otoko no ko
        title: Otoko
        type: string
        relation: kodomo
        relation_property: otoko_kodomo
      onna_id:
        description: Onna no ko
        title: Onna
        type: string
        relation: kodomo
        relation_property: onna_kodomo
```

Example initial values.
```
gakkos:
- id: gakko1
  name: Gakko 1
kodomos:
- id: otoko_id
  gakko_id: gakko1
- id: onna_id
  gakko_id: gakko1
oyas:
- id: oya_id
  otoko_id: otoko_id
  onna_id: onna_id
```

In this setting, if you want to get the list of oyas like `curl http://localhost:9091/v1.0/oyas`, following errors show:
```
18:13:49.638 gohan.db.sql DEBUG  Executing SQL query 'SELECT oyas.`id` as `oyas__id`, oyas.`otoko_id` as `oyas__otoko_id`, otoko_kodomo.`id` as `otoko_kodomo__id`, otoko_kodomo.`gakko_id` as `otoko_kodomo__gakko_id`, gakko.`id` as `gakko__id`, gakko.`name` as `gakko__name`, oyas.`onna_id` as `oyas__onna_id`, onna_kodomo.`id` as `onna_kodomo__id`, onna_kodomo.`gakko_id` as `onna_kodomo__gakko_id`, gakko.`id` as `gakko__id`, gakko.`name` as `gakko__name` FROM `oyas` LEFT JOIN `kodomos` as `otoko_kodomo` on `oyas`.`otoko_id` = `otoko_kodomo`.id LEFT JOIN `gakkos` as `gakko` on `otoko_kodomo`.`gakko_id` = `gakko`.id LEFT JOIN `kodomos` as `onna_kodomo` on `oyas`.`onna_id` = `onna_kodomo`.id LEFT JOIN `gakkos` as `gakko` on `onna_kodomo`.`gakko_id` = `gakko`.id ORDER BY oyas.`id` asc'
18:13:49.638 gohan.server.middleware ERROR  Error 1066: Not unique table/alias: 'gakko'
```

The root cause is gohan produced two same 'gakko' aliases.

morrisson:
If some resource has two or more relation properties to the same
resource which has some other relation properties, SQL join
generation can generate ambiguous alias. To avoid this problem,
table and column name should be generalized in the recursive manner.

fixes #215

### DIFF
--- a/db/sql/sql.go
+++ b/db/sql/sql.go
@@ -504,6 +504,10 @@ func makeColumn(tableName string, property schema.Property) string {
 	return fmt.Sprintf("%s.%s", tableName, quote(property.ID))
 }
 
+func makeAliasTableName(tableName string, property schema.Property) string {
+	return fmt.Sprintf("%s__%s", tableName, property.RelationProperty)
+}
+
 // MakeColumns generates an array that has Gohan style colmun names
 func MakeColumns(s *schema.Schema, tableName string, join bool) []string {
 	var cols []string
@@ -512,7 +516,8 @@ func MakeColumns(s *schema.Schema, tableName string, join bool) []string {
 		cols = append(cols, makeColumn(tableName, property)+" as "+quote(makeColumnID(tableName, property)))
 		if property.RelationProperty != "" && join {
 			relatedSchema, _ := manager.Schema(property.Relation)
-			cols = append(cols, MakeColumns(relatedSchema, property.RelationProperty, true)...)
+			aliasTableName := makeAliasTableName(tableName, property)
+			cols = append(cols, MakeColumns(relatedSchema, aliasTableName, true)...)
 		}
 	}
 	return cols
@@ -535,10 +540,11 @@ func makeJoin(s *schema.Schema, tableName string, q sq.SelectBuilder) sq.SelectB
 			continue
 		}
 		relatedSchema, _ := manager.Schema(property.Relation)
+		aliasTableName := makeAliasTableName(tableName, property)
 		q = q.LeftJoin(
-			fmt.Sprintf("%s as %s on %s.%s = %s.id", quote(relatedSchema.GetDbTableName()), quote(property.RelationProperty),
-				quote(tableName), quote(property.ID), quote(property.RelationProperty)))
-		q = makeJoin(relatedSchema, property.RelationProperty, q)
+			fmt.Sprintf("%s as %s on %s.%s = %s.id", quote(relatedSchema.GetDbTableName()), quote(aliasTableName),
+				quote(tableName), quote(property.ID), quote(aliasTableName)))
+		q = makeJoin(relatedSchema, aliasTableName, q)
 	}
 	return q
 }
@@ -559,7 +565,8 @@ func (tx *Transaction) decode(s *schema.Schema, tableName string, data map[strin
 		if property.RelationProperty != "" {
 			relatedSchema, _ := manager.Schema(property.Relation)
 			resourceData := map[string]interface{}{}
-			tx.decode(relatedSchema, property.RelationProperty, data, resourceData)
+			aliasTableName := makeAliasTableName(tableName, property)
+			tx.decode(relatedSchema, aliasTableName, data, resourceData)
 			resource[property.RelationProperty] = resourceData
 		}
 	}
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -38,13 +38,17 @@ import (
 )
 
 var (
-	server           *srv.Server
-	baseURL          = "http://localhost:19090"
-	schemaURL        = baseURL + "/gohan/v0.1/schemas"
-	networkPluralURL = baseURL + "/v2.0/networks"
-	subnetPluralURL  = baseURL + "/v2.0/subnets"
-	serverPluralURL  = baseURL + "/v2.0/servers"
-	testPluralURL    = baseURL + "/v2.0/tests"
+	server            *srv.Server
+	baseURL           = "http://localhost:19090"
+	schemaURL         = baseURL + "/gohan/v0.1/schemas"
+	networkPluralURL  = baseURL + "/v2.0/networks"
+	subnetPluralURL   = baseURL + "/v2.0/subnets"
+	serverPluralURL   = baseURL + "/v2.0/servers"
+	testPluralURL     = baseURL + "/v2.0/tests"
+	parentsPluralURL  = baseURL + "/v1.0/parents"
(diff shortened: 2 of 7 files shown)


db/sql/sql.go
```diff
@@ -505,14 +505,24 @@ func makeColumn(tableName string, property schema.Property) string {
 }
 
 // MakeColumns generates an array that has Gohan style colmun names
-func MakeColumns(s *schema.Schema, tableName string, join bool) []string {
+func MakeColumns(s *schema.Schema, tableName string, join bool, aliasTableName string) []string {
 	var cols []string
 	manager := schema.GetManager()
 	for _, property := range s.Properties {
-		cols = append(cols, makeColumn(tableName, property)+" as "+quote(makeColumnID(tableName, property)))
+        if aliasTableName != "" {
```
nati: how about
if aliasTableName == "" {
  aliasTableName = tableName;
}

server/server_test.go
```diff
@@ -237,6 +240,32 @@ var _ = Describe("Server package test", func() {
 		})
 	})
 
+    Describe("TwoSameResourceRelations", func ()  {
+        It("should work", func ()  {
+			By("creating 2 gakko")
```
nati: could you use english words in test code?

db/sql/sql.go
```diff
@@ -528,17 +538,28 @@ func makeStateColumns(s *schema.Schema) (cols []string) {
 	return cols
 }
 
-func makeJoin(s *schema.Schema, tableName string, q sq.SelectBuilder) sq.SelectBuilder {
+func makeJoin(s *schema.Schema, tableName string, q sq.SelectBuilder, aliasRelationPropertyPrefix string, aliasTableName string) sq.SelectBuilder {
```
nati: I'm not sure we need aliasTableName..
How about just pass newAliasTableName as tableName and omit l543-545

nati:
also rebase needed

db/sql/sql.go
morrisson: aliasTableName is used for if evaluation below, so this value should not be changed here.

db/sql/sql.go
morrisson: Actually, aliasTableName is important when top level resource has two children with the same resources which has two or more depth relations. Without newAliasTableName, current test passes because child resource of top level resource in my test schema has only one depth relation. In the next patch, I will modify test schema to check if this aliasTableName is needed.

server/server_test.go
morrisson: Ok.

db/sql/sql.go
nati: I believe we can do this always.

 newAliasTableName = aliasTableName + "__" + property.RelationProperty

db/sql/sql.go
nati: Why not do 

newAliasTableName = aliasTableName + "__" + property.RelationProperty

for always? if we do this, we can simplify logic.

morrisson:
Thanks for your comment. I can make simpler logic.

db/sql/sql.go
```diff
@@ -505,14 +505,18 @@ func makeColumn(tableName string, property schema.Property) string {
...
-func MakeColumns(s *schema.Schema, tableName string, join bool) []string {
+func MakeColumns(s *schema.Schema, tableName string, join bool, aliasTableName string) []string {
 	var cols []string
 	manager := schema.GetManager()
 	for _, property := range s.Properties {
-		cols = append(cols, makeColumn(tableName, property)+" as "+quote(makeColumnID(tableName, property)))
+		if aliasTableName == "" {
+			aliasTableName = tableName
+		}
+		cols = append(cols, makeColumn(aliasTableName, property)+" as "+quote(makeColumnID(tableName, property)))
```
nati: I don't understand this part.  

quote(makeColumnID(tableName, property)

I guess there would be case in that some properties get shadowed, right?
quote(makeColumnID(aliasTableName, property) works?

In this case, we can omit aliasTableName too

server/server_test.go
```diff
@@ -237,6 +241,38 @@ var _ = Describe("Server package test", func() {
...
+
+			By("creating 1 parent")
+			charlieParent := getParent("charlie", bobChild["id"].(string), aliceChild["id"].(string))
+			testURL("POST", parentsPluralURL, adminTokenID, charlieParent, http.StatusCreated)
+
+			By("assuring 1 parent was returned without error")
+			result := testURL("GET", parentsPluralURL, adminTokenID, nil, http.StatusOK)
+			res := result.(map[string]interface{})
+			parents := res["parents"].([]interface{})
+			Expect(parents).To(HaveLen(1))
```
nati: Could you check this response contains related resources?

morrisson:
I need to fix decode part additionally, but it should be simpler. Thanks for your comments.

server/server_test.go
```diff
@@ -1161,12 +1246,22 @@ func clearTable(tx transaction.Transaction, s *schema.Schema) error {
 	if s.IsAbstract() {
 		return nil
 	}
+	fmt.Println(schema.GetManager().Schemas())
```
nati: debug?

nati:
Thanks! the patch get really clean.
could you also run make in your local env before push? I guess lint tool such as go fmt isn't applied 
for this code.

server/server_test.go
morrisson: Oops. Sorry.

nati:
Thanks LGTM

Issue #33, opened 2015-11-10
Customize sync key per resources


nati:
We will support customizing sync key with template support.

sync_key: "{{.resource.node_id}}/server/{{.resource.id}}"

Usecase: Scheduling

Each node will watch certain path
/node1/servers/server_1
/node2/servers/server_1

In this case we will add this in schema config

```
sync_key: "{{.resource.node_id}}/server/{{.resource.id}}"
```

PR #227, opened 2016-05-19
Allow dynamic sync path in schema metadata

wsikorska-codilime:
In metadata it is allowed to have additional field sync_key_template. It allows dynamic path generation based on properties,

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -19,7 +19,9 @@ import (
 	"fmt"
 
 	"github.com/cloudwan/gohan/util"
+	"github.com/flosch/pongo2"
 	"github.com/xeipuuv/gojsonschema"
+	"strings"
 )
 
 //Schema type for defining data type
@@ -344,6 +346,62 @@ func (schema *Schema) StateVersioning() bool {
 		return false
 	}
 	return stateful
+}
+
+func (schema *Schema) SyncKeyTemplate() (syncKeyTemplate string, ok bool) {
+	syncKeyTemplateRaw, ok := schema.Metadata["sync_key_template"]
+	if !ok {
+		return
+	}
+	syncKeyTemplate, ok = syncKeyTemplateRaw.(string)
+	return
+}
+
+func (schema *Schema) GenerateCustomPath(data map[string]interface{}) (path string, err error) {
+	syncKeyTemplate, ok := schema.SyncKeyTemplate()
+	if !ok {
+		err = fmt.Errorf("Failed to read sync_key_template from schema %v", schema.URL)
+		return
+	}
+	tpl, err := pongo2.FromString(syncKeyTemplate)
+	if err != nil {
+		return
+	}
+	path, err = tpl.Execute(pongo2.Context{}.Update(data))
+	return
+}
+
+func (schema *Schema) GetResourceIdFromPath(schemaPath string) string {
+	syncKeyTemplate, ok := schema.SyncKeyTemplate()
+	if !ok {
+		return strings.TrimPrefix(schemaPath, schema.URL + "/")
+	}
+
+	syncKeyTemplateSplit := strings.Split(syncKeyTemplate, "/")
+	schemaPathSplit := strings.Split(schemaPath, "/")
+	if len(schemaPathSplit) >= len(syncKeyTemplateSplit) {
+		resourceID := ""
+		for k, v := range syncKeyTemplateSplit {
+			if v == "{{id}}" {
+				resourceID = schemaPathSplit[k]
+				break
+			}
+		}
+		return resourceID
+	} else {
+		return strings.TrimPrefix(schemaPath, schema.URL+"/")
+	}
+}
+
+func GetSchemaByPath(path string) *Schema {
+	var schemaByPath *Schema
+	for _, schema := range GetManager().Schemas() {
+		if strings.HasPrefix(path, schema.URL) {
+			schemaByPath = schema
+			break
+		}
+	}
+	return schemaByPath
 }
 
 // FormatParentID ...
--- a/server/sync.go
+++ b/server/sync.go
@@ -186,9 +186,11 @@ func (server *Server) syncEvent(resource *schema.Resource) error {
 	}
 	defer tx.Close()
 	eventType := resource.Get("type").(string)
-	path := resource.Get("path").(string)
-	path = configPrefix + path
+	resourcePath := resource.Get("path").(string)
 	body := resource.Get("body").(string)
+
+	path := generatePath(resourcePath, body)
+
 	version, ok := resource.Get("version").(int)
 	if !ok {
 		log.Debug("cannot cast version value in int for %s", path)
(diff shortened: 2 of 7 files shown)


nati:
Thanks for this.
I have already an issue for this. so please close this in commit message.
https://github.com/cloudwan/gohan/issues/33

schema/schema.go
```diff
@@ -346,6 +347,55 @@ func (schema *Schema) StateVersioning() bool {
...
+		return ""
+	}
+	syncKeyTemplate, ok := syncKeyTemplateRaw.(string)
+	if !ok {
+		return ""
+	}
+	return syncKeyTemplate
+}
+
+func (schema *Schema) GenerateCustomPath(data map[string]interface{}) (string, error) {
```
nati: let's use pongo2 template instead of defining function by ourself.
https://github.com/flosch/pongo2

pongo2 is already used in extension.

marcin-ptaszynski:
We also should support dynamic path when matching paths from sync (/state, /monitoring). Perhaps we need two keys in metadata for it or we can use simple syntax to make it work both ways.

marcin-ptaszynski:
LGTM, please rebase.

schema/schema.go
```diff
@@ -360,29 +361,36 @@ func (schema *Schema) SyncKeyTemplate() string {
...
+	}
+	path, err := tpl.Execute(pongo2.Context{}.Update(data))
+	if err != nil {
+		return "", err
+	}
+	return path, nil
+}
+
+func (schema *Schema) GetResourceIdFromPath(schemaPath string) string {
+	resourceID := ""
```
vozhyk-: This can be a named return argument instead.

schema/schema.go
```diff
@@ -346,6 +348,62 @@ func (schema *Schema) StateVersioning() bool {
 	return stateful
 }
 
+func (schema *Schema) SyncKeyTemplate() string {
+	syncKeyTemplateRaw, ok := schema.Metadata["sync_key_template"]
+	if !ok {
+		return ""
```
vozhyk-: I suggest making `ok` another return argument.

schema/schema.go
```diff
@@ -346,6 +348,62 @@ func (schema *Schema) StateVersioning() bool {
...
+	path, err := tpl.Execute(pongo2.Context{}.Update(data))
+	if err != nil {
+		return "", err
+	}
+	return path, nil
+}
+
+func (schema *Schema) GetResourceIdFromPath(schemaPath string) string {
+	resourceID := ""
+	if schema.SyncKeyTemplate() == "" {
```
vozhyk-: [#comment](https://github.com/cloudwan/gohan/pull/227/files#r64607353) ... and using it here.

schema/schema.go
nati: +1

server/sync.go
```diff
@@ -186,9 +186,29 @@ func (server *Server) syncEvent(resource *schema.Resource) error {
...
-	path := resource.Get("path").(string)
-	path = configPrefix + path
+	resourcePath := resource.Get("path").(string)
 	body := resource.Get("body").(string)
+
+	var curSchema = schema.GetSchemaByPath(resourcePath)
+
+	path := resourcePath
+	_, ok := curSchema.SyncKeyTemplate()
+	if ok {
```
vozhyk-: This can be one line
```go
if _, ok := curSchema.SyncKeyTemplate(); ok {
```

server/sync.go
```diff
@@ -186,9 +186,29 @@ func (server *Server) syncEvent(resource *schema.Resource) error {
...
 	defer tx.Close()
 	eventType := resource.Get("type").(string)
-	path := resource.Get("path").(string)
-	path = configPrefix + path
+	resourcePath := resource.Get("path").(string)
 	body := resource.Get("body").(string)
+
+	var curSchema = schema.GetSchemaByPath(resourcePath)
+
+	path := resourcePath
```
vozhyk-: It seems it'll be good to extract this part (`path` generation) into a function.

schema/schema.go
```diff
@@ -348,36 +348,37 @@ func (schema *Schema) StateVersioning() bool {
...
+	syncKeyTemplate, ok := schema.SyncKeyTemplate()
+	if !ok {
+		err = fmt.Errorf("Failed to read sync_key_template from schema %v", schema.URL)
+		return
 	}
-	path, err := tpl.Execute(pongo2.Context{}.Update(data))
+	tpl, err := pongo2.FromString(syncKeyTemplate)
 	if err != nil {
-		return "", err
+		return
```
vozhyk-: I'm not sure this returns the proper `err`. My understanding is that `tpl, err :=` creates a new `err` variable, but this `return` returns the one declared as a return argument.

schema/schema.go
```diff
@@ -348,36 +348,37 @@ func (schema *Schema) StateVersioning() bool {
...
+
+func (schema *Schema) SyncKeyTemplate() (syncKeyTemplate string, ok bool) {
 	syncKeyTemplateRaw, ok := schema.Metadata["sync_key_template"]
 	if !ok {
-		return ""
-	}
-	syncKeyTemplate, ok := syncKeyTemplateRaw.(string)
-	if !ok {
-		return ""
+		return
```
vozhyk-: [Same](https://github.com/cloudwan/gohan/pull/227#discussion_r64899412) for this one. It might be necessary to use `=` instead of `:=` for this to return the intended variable's value.

schema/schema.go
```diff
@@ -346,6 +348,63 @@ func (schema *Schema) StateVersioning() bool {
...
+	}
+	path, err = tpl.Execute(pongo2.Context{}.Update(data))
+	return
+}
+
+func (schema *Schema) GetResourceIdFromPath(schemaPath string) string {
+	resourceID := ""
+	syncKeyTemplate, ok := schema.SyncKeyTemplate()
+	if !ok {
+		resourceID = strings.TrimPrefix(schemaPath, schema.URL+"/")
```
vozhyk-: This can be returned immediately and the `else` dropped, to be more Go-style.

schema/schema.go
wsikorska-codilime: The err variable will be overridden in this case. Sample code to check:
`package main
 
import "fmt"
 
func main() {
  a, b := checkShadowing()
  fmt.Printf("%s, %s\n", a, b)
}

func checkShadowing() (path string, err error) {
	a, ok := get1()
	if !ok {
		err = fmt.Errorf("error not ok")
		return
	}
	b, err := get2(a)
	if err != nil {
		return
	}
	path, err = get3(b)
	return
}

func get1() (a string, ok bool) {
    return "aaa", true
}

func get2(a string) (b string, err error) {
    return a + "bbb", fmt.Errorf("error in get2")
}

func get3(b string) (c string, err error) {
    return b + c, fmt.Errorf("error in get3")
}` 

It returns ", error in get2", so desired error. This means that err is overridden.

vozhyk-: Great, thanks for checking!

vozhyk-:
LGTM.

nati:
OK LGTM thanks

PR #237, opened 2016-06-01
Fill default values when creating initial data

morrisson:
Just like 899503d , default values are not filled when creating
initial data. Therefore, sqlite3 occurs a SQL error for inserting
data if there are some missing properties in input data. This fix
enables creating initial data for sqlite3.


morrisson:
It might contain some bug, I will check so close this PR at this moment.

morrisson:
I found that is no problem, so reopen this PR.

PR #238, opened 2016-06-02
Add gohan logging built-ins

vozhyk-:
Currently, the only way of logging available for javascript applications is writing directly to stderr using `console.log`.

Add built-ins that use Go logging: `gohan_log_<level>(message)`, `gohan_log(module, level, message)`, and constants for log levels, so it's possible to run something like this in javascript:

```javascript
gohan_log_debug("test")
// or:
gohan_log(LOG_MODULE, DEBUG, "test");
```

and it gets shown in Gohan logs similarly to this:
```
17:57:08.901 gohan.extension.network.post_list_in_transaction DEBUG  test
```

TODO:
- [ ] Format the rest of `docs/source/extension.rst` the way the added documentation is formatted?

### DIFF
--- a/extension/otto/gohan_logging.go
+++ b/extension/otto/gohan_logging.go
@@ -1,0 +1,169 @@
+// Copyright (C) 2016 NTT Innovation Institute, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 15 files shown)


extension/otto/gohan_logging.go
```diff
@@ -0,0 +1,177 @@
...
+		    var old_module = LOG_MODULE;
+		    LOG_MODULE += "." + new_module;
+		    return old_module;
+		}
+
+		function gohan_log_module_restore(old_module){
+		    LOG_MODULE = old_module;
+		}
+
+		function Log(msg) {
```
marcin-ptaszynski: this shouldn't be here

extension/otto/gohan_logging.go
```diff
@@ -0,0 +1,177 @@
...
+		    gohan_log_debug(msg);
+		}
+
+		function logError(msg) {
```
marcin-ptaszynski: ditto

extension/otto/gohan_logging.go
vozhyk-: Removed.

extension/otto/gohan_logging.go
vozhyk-: Removed.

PR #240, opened 2016-06-03
Improve test handling in gohanscript

nati:


server/resources/resource_management.go
```diff
@@ -364,7 +364,7 @@ func CreateResource(
 		dataMap["tenant_id"] = context["tenant_id"]
 	}
 
-	if tenantID, ok := dataMap["tenant_id"]; ok {
+	if tenantID, ok := dataMap["tenant_id"]; ok && tenantID != nil {
```
vozhyk-: The `ok &&` is unnecessary here.

nati: tenantID could be nil even if we have key
https://play.golang.org/p/a8hx_300QZ

vozhyk-: Now I see I was wrong.

Issue #246, opened 2016-06-08
Gohan panics if we use getters or setters in javascript extension code


marcin-ptaszynski:
If we use ECMA5 [getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) or [setters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set) anywhere in javascript code, including 3rd party libraries, we'll get a panic upon any request.

Example code (from MDN):
```
var log = ['test'];
var obj = {
  get latest () {
    if (log.length == 0) return undefined;
    return log[log.length - 1]
  }
}
```

Resulting Traceback:
```
[martini] PANIC: runtime error: invalid memory address or nil pointer dereference
/home/marcin/.gvm/gos/go1.6/src/runtime/panic.go:426 (0x434a89)
        gopanic: reflectcall(nil, unsafe.Pointer(d.fn), deferArgs(d), uint32(d.siz), uint32(d.siz))
/home/marcin/.gvm/gos/go1.6/src/runtime/panic.go:45 (0x433139)
        panicmem: panic(memoryError)
/home/marcin/.gvm/gos/go1.6/src/runtime/sigpanic_unix.go:24 (0x44985a)
        sigpanic: panicmem()
/home/marcin/workspace/gohan/src/github.com/cloudwan/gohan/vendor/github.com/dop251/otto/clone.go:88 (0xae206f)
        (*_clone).object: return in.objectClass.clone(in, out, clone)
/home/marcin/workspace/gohan/src/github.com/cloudwan/gohan/vendor/github.com/dop251/otto/clone.go:151 (0xae2baf)
        (*_clone).property: set := clone.object(value[1])
/home/marcin/workspace/gohan/src/github.com/cloudwan/gohan/vendor/github.com/dop251/otto/object_class.go:471 (0xb54e35)
        objectClone: out.property[index] = clone.property(property)
/home/marcin/workspace/gohan/src/github.com/cloudwan/gohan/vendor/github.com/dop251/otto/clone.go:88 (0xae207c)
        (*_clone).object: return in.objectClass.clone(in, out, clone)
...
```

PR #247, opened 2016-06-08
fix otto clone get/set bug

marcin-ptaszynski:
During cloning, property.value is assumed to be of type Value,
_propertyGetSet type is also a possiblity when using ECMA5.1 getters
or setters.

This commit fixes it by proper handling of _propertyGetSet type


marcin-ptaszynski:
fixes issue #246

PR #249, opened 2016-06-09
Fixed generating sync_key_template path

wsikorska-codilime:
resource.Get("body") sometimes doesn't have all necessary fields to put them into generated path, so the data should be fetched from db.

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -348,6 +348,7 @@ func (schema *Schema) StateVersioning() bool {
 	return stateful
 }
 
+//SyncKeyTemplate - for custom paths in etcd
 func (schema *Schema) SyncKeyTemplate() (syncKeyTemplate string, ok bool) {
 	syncKeyTemplateRaw, ok := schema.Metadata["sync_key_template"]
 	if !ok {
@@ -357,6 +358,7 @@ func (schema *Schema) SyncKeyTemplate() (syncKeyTemplate string, ok bool) {
 	return
 }
 
+//GenerateCustomPath - returns custom path based on sync_key_template
 func (schema *Schema) GenerateCustomPath(data map[string]interface{}) (path string, err error) {
 	syncKeyTemplate, ok := schema.SyncKeyTemplate()
 	if !ok {
@@ -371,7 +373,8 @@ func (schema *Schema) GenerateCustomPath(data map[string]interface{}) (path stri
 	return
 }
 
-func (schema *Schema) GetResourceIdFromPath(schemaPath string) string {
+//GetResourceIDFromPath - parse path and gets resourceID from it
+func (schema *Schema) GetResourceIDFromPath(schemaPath string) string {
 	syncKeyTemplate, ok := schema.SyncKeyTemplate()
 	if !ok {
 		return strings.TrimPrefix(schemaPath, schema.URL+"/")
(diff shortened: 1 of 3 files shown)


nati:
We should do this when we make body in the transaction, otherwise we can't ensure consistency of data.

nati:
Could you make your commit message descriptive? 
I'm not sure what is fixed..

wsikorska-codilime:
During state update the path in etcd was chosen based on schema.url prefix instead of sync_key_template path. If path in etcd (when it's based on sync_key_template) doesn't have the same prefix as resource url in api, then gohan doesn't watch state updates for that resource. The same happens in the case of monitoring update.

nati:
I got it. Could you put the statement in the commit message?

schema/schema.go
```diff
@@ -393,7 +393,8 @@ func (schema *Schema) GetResourceIdFromPath(schemaPath string) string {
 	}
 }
 
-func GetSchemaByPath(path string) *Schema {
+//Get schema by resource path (from API)
```
vozhyk-: The comment should start with the function name. `golint` will likely complain.

schema/schema.go
```diff
@@ -404,6 +405,42 @@ func GetSchemaByPath(path string) *Schema {
 	return schemaByPath
 }
 
+//Get schema by etcd path
```
vozhyk-: Same here.

schema/schema_test.go
```diff
@@ -113,6 +113,36 @@ var _ = Describe("Schema", func() {
...
+			Expect(idExists).To(BeTrue())
+		})
+
+		It("GetSchemaByTemplatePath", func() {
+			data := map[string]interface{}{
+				"id": "1234",
+			}
+			path, err := metadataIdSchema.GenerateCustomPath(data)
+			Expect(err).To(Succeed())
+			Expect(metadataIdSchema).To(Equal(GetSchemaByPath(path)))
```
vozhyk-: If I understand the intention of the test, his should be like
```
Expect(GetSchemaByPath(path)).To(Equal(metadataIdSchema))
```
(`Expect(ACTUAL).To(Equal(EXPECTED))`(https://onsi.github.io/gomega/#making-assertions))

marcin-ptaszynski:
Commit message looks malformatted. Please check https://git-scm.com/book/ch5-2.html

schema/schema.go
wsikorska-codilime: Thanks, I fixed comments.

schema/schema_test.go
wsikorska-codilime: I'm testing GetSchemaByPath, so GetSchemaByPath(path) is an expected value.

wsikorska-codilime: Ok, fixed.

marcin-ptaszynski:
Hi @nati , what is the status of this PR? Can we merge it?

nati:
@marcin-ptaszynski let me review again. Does this look good for you?

nati:
Commit messege need to be updated.
How about this?

During state update the path in etcd was chosen based on schema.url prefix instead of sync_key_template path. If path in etcd (when it's based on sync_key_template) doesn't have the same prefix as resource url in api, then gohan doesn't watch state updates for that resource. The same happens in the case of monitoring update.
In this commit, we apply sync_key_template for sync path.

wsikorska-codilime:
I changed commit message and rebased to current version of master branch.

schema/schema.go
```diff
@@ -404,6 +407,42 @@ func GetSchemaByPath(path string) *Schema {
...
+func GetSchemaByPath(path string) *Schema {
+	var schemaByPath *Schema
+	for _, schema := range GetManager().Schemas() {
+		syncKeyTemplate, ok := schema.SyncKeyTemplate()
+		if ok {
+			if checkIfPathMatchesSyncKeyTemplate(path, syncKeyTemplate) {
+				schemaByPath = schema
+				break
+			}
+		} else {
```
nati: } else if {

schema/schema.go
```diff
@@ -404,6 +407,42 @@ func GetSchemaByPath(path string) *Schema {
...
 
+//GetSchemaByPath - gets schema by sync_key_template path
+func GetSchemaByPath(path string) *Schema {
+	var schemaByPath *Schema
+	for _, schema := range GetManager().Schemas() {
+		syncKeyTemplate, ok := schema.SyncKeyTemplate()
+		if ok {
+			if checkIfPathMatchesSyncKeyTemplate(path, syncKeyTemplate) {
+				schemaByPath = schema
+				break
```
nati: you can return here

schema/schema.go
```diff
@@ -404,6 +407,42 @@ func GetSchemaByPath(path string) *Schema {
...
+			}
+		} else {
+			if strings.HasPrefix(path, schema.URL) {
+				schemaByPath = schema
```
nati: you can return here

schema/schema.go
```diff
@@ -404,6 +407,42 @@ func GetSchemaByPath(path string) *Schema {
 	return schemaByPath
 }
 
+//GetSchemaByPath - gets schema by sync_key_template path
+func GetSchemaByPath(path string) *Schema {
+	var schemaByPath *Schema
```
nati: we don't need to define this

nati:
ok thanks!

PR #251, opened 2016-06-15
test framework: Print only logs for failing tests

vozhyk-:
- Print only logs for failing tests

    Add a command-line flag (`-v`) to print logs for all tests.
    Print test function result immediately
      (only if it fails or `-v` is specified).
- Don't make test file paths absolute
- Add an `All tests have passed.` message
- Print summary only when a test failed

    Print only failed tests in summary.
    Do the above unless `-v` is supplied.


nati:
please rebase commits

vozhyk-:
@nati Done.

PR #218, opened 2016-05-17
Add gohan_config built-in

vozhyk-:
- Refactor built-ins' argument parsing
    This refactor includes:

    1. Strict and consistent checking of types in builtin functions arguments - creating functions for them
    2. Extracting non-otto code from builtin functions to test them better
    3. Making builtin functions code cleaner to easily implement new functions
    4. Fixing bugs:
        - not all functions create transaction when passed null
- Add `gohan_config` built-in

PR #255, opened 2016-06-28
add gohan_raw_http

marcin-ptaszynski:
gohan_raw_http works directly on Round Tripper instead of http.Client,
allowing developer to manually handle redirects and manage sessions

### DIFF
--- a/extension/otto/gohan_util.go
+++ b/extension/otto/gohan_util.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/cloudwan/gohan/schema"
 	"github.com/cloudwan/gohan/util"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


extension/otto/gohan_util.go
```diff
@@ -84,6 +85,63 @@ func init() {
...
+				if err != nil {
+					ThrowOttoException(&call, err.Error())
+				}
+				rawData, err := GetString(call.Argument(3))
+				if err != nil {
+					ThrowOttoException(&call, err.Error())
+				}
+				//TODO: pass Transport options like timeouts
+
+				// prepare request
```
vozhyk-: The rest of the function below could be a separate `GohanRawHTTP()` function, like [pull #218@Refactor builtins:GohanDbQuery](https://github.com/cloudwan/gohan/pull/218/commits/9780d7929938862e144f66b078018787cb3f8a77#diff-ff4e9f745acd7902b3d865025f2589bbR284).

PR #261, opened 2016-07-12
gohan_raw_http: string response body type

marcin-ptaszynski:
Type []byte is converted to array of numbers in otto, which results
in obfuscated logs, requires manual conversion to String type
in order to work with response body and is generally a bad idea.

In this commit, Gohan automatically converts arrived data to string
and adds missing documentation for gohan_raw_http.


nati:
could you rebase with master?

marcin-ptaszynski:
@nati , rebased.

PR #262, opened 2016-07-13
npm support

wsikorska-codilime:
Introduced motto which will allows us to integrate npm packages inside javascript extensions. Now we can use require function to integrate module installed via npm in node_modules at working directory. Migration from dop251/otto to robertkrimen/otto was necessary to make it work. robertkrimen otto seems to be more alive than dop251.


examples/npm_example/README.md
```diff
@@ -0,0 +1,16 @@
+Using NPM modules in extensions
+-------------------------------
+
+In this example, we show how we can
+use modules installed via npm in js extensions
+
+We have installed npm module underscore in
```
marcin-ptaszynski: Registering modules should also be documented.

examples/npm_example/example_schema.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - status
+    required: []
+    type: object
+  singular: pet
+  title: Pet
+extensions:
+- id: schema
+  code_type: javascript
+  code: |
+    gohan_register_handler("pre_list", function (context){
```
marcin-ptaszynski: Could you give an example of registering module too?

docs/source/extension.rst
```diff
@@ -294,7 +294,8 @@ close ssh session
 
 - require(module)
 
-Dynamically load modules
+Dynamically load modules loaded in source code or
+installed via npm in node_modules at working directory
```
marcin-ptaszynski: Is working directory where server is executed? I think it should be configurable path from config.

wsikorska-codilime: Ok, I'll add a configurable path in config. Working directory is a directory of loaded config yaml.

nati:
wow this is pretty neat.
pl rebase code

PR #268, opened 2016-07-21
dont override initial data

marcin-ptaszynski:
When running in server mode, initial data read from config overrides
rows changed by operator. This may not be desired behavior when
running in server mode.

This commit allows CopyDBResources method to set overrideExisting flag.
When set, it won't force db update on existing row.
This flag is set to false when handling initial data for Server mode.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -287,7 +287,7 @@ Useful for development purposes.`,
 				util.ExitFatal(err)
 			}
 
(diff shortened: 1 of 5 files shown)


wsikorska-codilime:
LGTM

nati:
s/dont/don't/  in commit message, otherwise LGTM
please also rebase with master

marcin-ptaszynski:
@nati , done. Thanks

PR #273, opened 2016-07-29
Crash on sync/snmp listener goroutine panic

marcin-ptaszynski:
If sync processing goroutine crashes for any reason, we should not
just log panic and keep on going. In such case, Gohan is rendered
unusable, but it doesn't go down completely, so it's not detectable.

Another problem is that in case of failure in any goroutines mentioned
above logging only error message is not sufficient for debugging.

This commit changes above behavior so in case of listener goroutine
panic, it will log whole stacktrace and crash whole Gohan process.

### DIFF
--- a/server/amqp.go
+++ b/server/amqp.go
@@ -61,7 +61,7 @@ func listenAMQP(server *Server) {
 
 	for _, queue := range queues {
 		go func(queue string) {
-			defer util.LogPanic(log)
+			defer util.LogFatalPanic(log)
 			for server.running {
 				conn, err := amqp.Dial(connection)
 				if err != nil {
--- a/server/snmp.go
+++ b/server/snmp.go
@@ -51,7 +51,7 @@ func startSNMPProcess(server *Server) {
 
 	buf := make([]byte, 1024)
 	go func() {
-		defer util.LogPanic(log)
+		defer util.LogFatalPanic(log)
 		defer conn.Close()
 		for server.running {
 			rlen, remote, err := conn.ReadFromUDP(buf)
--- a/server/sync.go
+++ b/server/sync.go
@@ -284,7 +284,7 @@ func startSyncProcess(server *Server) {
 	pollingTicker := time.Tick(eventPollingTime)
 	committed := transactionCommitInformer()
 	go func() {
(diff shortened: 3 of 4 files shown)


nati:
In a case of we got panic on go routine, the Gohan process gets crash.
Let's say if we have a bug in an extension code, It may fail. But I don't want to get the whole system down.

marcin-ptaszynski:
I agree, thanks. I've added additional LogFatalPanic function to util and call only when listener goroutine crashes. Normal LogPanic logs full stacktrace.

marcin-ptaszynski:
Hi @nati, I've updated PR based on your comment. It should not crash Gohan if extension processing code panics, only listener/dispatcher goroutine.

nati:
May I ask why you want to get Gohan process crashed on sync error?
There is various reason we may get panic.

marcin-ptaszynski:
@nati , consider following scenario:
1. You have 3 Gohan nodes, and 1 is a sync master.
2. Sync processing goroutine [line 483](https://github.com/cloudwan/gohan/pull/273/files#diff-f227c71de6c6b96171aa8b42c56a3c2cR483) crashes for any reason.
3. etcd watching and locking goroutine [line 463](https://github.com/cloudwan/gohan/pull/273/files#diff-f227c71de6c6b96171aa8b42c56a3c2cR463) works fine.

The result is that even though Gohan is running in HA, no events are being processed from sync because locking and watchine part works fine, but processing does not.

Perhaps we should rework it a bit, so that the loop can be recovered, but IMO, in such situation, it's still better to crash than do nothing (this requires manual restart).

nati:
How about restart go routine on panic?
I'm still not sure we need crash everything on such case.

marcin-ptaszynski:
I agree, we could do that. On the other hand, this seems a little complicated, especially if we want to avoid triggering endless restart loops, etc. We should consider defining such policies internally in Gohan. Currently, it's simply easier just to crash.

nati:
fair enough. could you rebase this code? I'll merge

marcin-ptaszynski:
thanks, @nati . rebased.

PR #274, opened 2016-08-01
Compare tenant IDs between parent and child

abonie:
This is in regard to issue 229

A check has been added to functions UpdateResource
and CreateResource that compares tenant_id in parent
and child resources for equality and fails if they differ.

### DIFF
--- a/server/resources/resource_management.go
+++ b/server/resources/resource_management.go
@@ -368,6 +368,70 @@ func CreateOrUpdateResource(
 	return false, UpdateResource(context, dataStore, identityService, resourceSchema, resourceID, dataMap)
 }
 
+func hasProperty(s *schema.Schema, propertyID string) bool {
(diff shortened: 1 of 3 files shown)


server/resources/resource_management.go
```diff
@@ -368,6 +368,70 @@ func CreateOrUpdateResource(
 	return false, UpdateResource(context, dataStore, identityService, resourceSchema, resourceID, dataMap)
 }
 
+func hasProperty(s *schema.Schema, propertyID string) bool {
```
nati: This method should be under schema package.

schema.hasProperty(propertyID string)

nati:
There is also a use-case where tenant_id could be different on parent and child.
Can we make this configurable using policy?

nati:
please reopen this if you still need this

PR #277, opened 2016-08-04
Configurable log levels in extension tests

wsikorska-codilime:
Introduces ability to use config file to set up logging levels for extension tests.


nati:
plz rebase

PR #278, opened 2016-08-09
Configurable log levels in extension tests

wsikorska-codilime:
Introduces ability to use config file to set up logging levels for extension tests. Duplicate pull request closed due to my mistake.

### DIFF
--- a/extension/framework/framework.go
+++ b/extension/framework/framework.go
@@ -24,7 +24,9 @@ import (
 	"github.com/cloudwan/gohan/extension/framework/buflog"
 	"github.com/cloudwan/gohan/extension/framework/runner"
 	"github.com/codegangsta/cli"
+	gohan_log "github.com/cloudwan/gohan/log"
 	logging "github.com/op/go-logging"
+	"github.com/cloudwan/gohan/util"
 )
 
 var log = logging.MustGetLogger("extest")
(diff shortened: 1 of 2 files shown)


extension/framework/framework.go
```diff
@@ -24,7 +24,9 @@ import (
 	"github.com/cloudwan/gohan/extension/framework/buflog"
 	"github.com/cloudwan/gohan/extension/framework/runner"
 	"github.com/codegangsta/cli"
+	logg "github.com/cloudwan/gohan/log"
```
vozhyk-: Maybe import it as `gohan_log`?

extension/framework/framework.go
```diff
@@ -33,9 +35,28 @@ var log = logging.MustGetLogger("extest")
 func TestExtensions(c *cli.Context) {
 	buflog.SetUpDefaultLogging()
 
+	var config *util.Config
```
vozhyk-: Why not just use `config := util.GetConfig()`? As far as I understand, this can't fail.

extension/framework/framework.go
```diff
@@ -33,9 +35,28 @@ var log = logging.MustGetLogger("extest")
...
+			log.Error(fmt.Sprintf("Failed to set up logging: %v", err))
+			os.Exit(1)
+		}
+	}
+
 	testFiles := getTestFiles(c.Args())
 
-	returnCode := RunTests(testFiles, c.Bool("verbose"))
+	//logging from config is a limited printAllLogs option
+	returnCode := RunTests(testFiles, c.Bool("verbose") || config != nil)
```
vozhyk-: Why enable `verbose` when there is a config? I thought they would be independent options (`verbose` meaning "print logs for passing tests too"; and the config specifying the log levels for what gets printed, regardless if it's only logs for failing tests).

A config option for verbosity can be added, if it's needed.

Or, if we don't want `verbose` being disabled by default, this can be changed.

extension/framework/framework.go
```diff
@@ -33,9 +35,28 @@ var log = logging.MustGetLogger("extest")
...
 	buflog.SetUpDefaultLogging()
 
+	var config *util.Config
+	configFilePath := c.String("config-file")
+
+	if configFilePath != "" && !c.Bool("verbose") {
+		config = util.GetConfig()
+		err := config.ReadConfig(configFilePath)
+		if err != nil {
+			log.Error(fmt.Sprintf("Failed to read config from path %s", configFilePath))
```
vozhyk-: Add `%v", <...> err`

extension/framework/framework.go
wsikorska-codilime: It's because I didn't want to add unnecessary logic to this code. We don't want to log config logs via buffer log, because then we will need to rewrite gohan_log (or logg) functionality in it. For me there is no point in using both config and verbose option - you can set in config default logging level to DEBUG and have verbose output.

PR #280, opened 2016-08-09
Use SetIsolationLevel method for sync events

marcin-ptaszynski:
Without it, it will use database default, which is usually set to
REPEATABLE_READ. This isolation level is prone to races.

This commit adds new event types for transaction and uses them to
retrieve default isolation level, which in this case is Serializable.


nati:
could you update doc?

marcin-ptaszynski:
I've updated Custom Isolation section in schema.rst

Issue #257, opened 2016-07-06
Support for custom actions in Gohan Client


marcin-ptaszynski:
Gohan supports custom actions for schema, but `gohan client` supports only basic CRUD. It would've have been good, if `gohan client` could parse those actions and make the request based on action and it's input.

thayashi:
This'd be nice, I've thought same idea.

PR #283, opened 2016-08-18
Add support for custom actions in client

abonie:
Issue #257 

Gohan client will parse custom actions defined in schema
and support them in CLI.

### DIFF
--- a/cli/client/arguments.go
+++ b/cli/client/arguments.go
@@ -25,6 +25,45 @@ import (
 	"github.com/cloudwan/gohan/schema"
 )
 
+func (gohanClientCLI *GohanClientCLI) getCustomArgsAsMap(
+	args []string,
+	actionInput string,
+	action schema.Action,
+) (argsMap map[string]interface{}, err error) {
+	argsMap = map[string]interface{}{}
+	if action.InputSchema != nil {
+		var value interface{}
+		inputType, ok := action.InputSchema["type"].(string)
+		if !ok {
+			return nil, fmt.Errorf("Invalid input schema")
+		}
+		switch inputType {
+		case "integer", "number":
+			value, err = strconv.ParseInt(actionInput, 10, 64)
+		case "boolean":
+			value, err = strconv.ParseBool(actionInput)
+		case "array", "object":
+			err = json.Unmarshal([]byte(actionInput), &value)
+		default:
+			value = actionInput
+		}
+		if err != nil {
+			return nil, fmt.Errorf("Error parsing action input %s:", err)
+		}
+		argsMap[action.ID] = value
+	}
+	for i := 0; i < len(args); i += 2 {
+		key := strings.TrimPrefix(args[i], "--")
+		if _, ok := commonParams[key]; !ok {
+			return nil, fmt.Errorf("Error parsing parameter %s", key)
+		}
+		value := args[i+1]
+		argsMap[key] = value
+	}
+	err = gohanClientCLI.handleCommonArguments(argsMap)
+	return
+}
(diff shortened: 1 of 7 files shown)


marcin-ptaszynski:
@abonie , I think you should update document additional client commands here https://github.com/cloudwan/gohan/blob/master/docs/source/commands.rst#command

cli/client/arguments.go
```diff
@@ -25,6 +25,45 @@ import (
...
+			return nil, fmt.Errorf("Error parsing action input %s:", err)
+		}
+		argsMap[action.ID] = value
+	}
+	for i := 0; i < len(args); i += 2 {
+		key := strings.TrimPrefix(args[i], "--")
+		if _, ok := commonParams[key]; !ok {
+			return nil, fmt.Errorf("Error parsing parameter %s", key)
+		}
+		value := args[i+1]
```
marcin-ptaszynski: isn't it going to crash if we pass odd number of arguments, like `gohan client action --opt1 value1 --opt2`?

schema/schema.go
```diff
@@ -489,6 +489,17 @@ MergeAction:
 
 //JSON returns json format of schema
 func (schema *Schema) JSON() map[string]interface{} {
+	// TODO check if there exist such a function already
+	// if not maybe put this in separate function
+	actions := map[string]interface{}{}
+	for _, a := range schema.Actions {
+		actions[a.ID] = map[string]interface{}{
```
marcin-ptaszynski: I see why you prefer jsonn style over straight GO struct marshal, but it's much easier to achieve with tags, https://golang.org/pkg/encoding/json/#Marshal http://stackoverflow.com/questions/11693865/lower-case-key-names-with-json-marshal-in-go

abonie: Cool, I will add tags to Action definition and use Marshal here

marcin-ptaszynski: I see, I think it's fine as it is.

docs/source/commands.rst
```diff
@@ -246,3 +246,12 @@ it's name or id could be used:
   gohan client subnet create --network "Network Name"
   gohan client subnet create --network network-id
   gohan client subnet create --network_id network-id
+
+Custom commands
+---------------
+
+Custom commands, if specified in schema, are also supported by gohan client and should be invoked as follows:
```
marcin-ptaszynski: there is no reference to Actions.

abonie: I don't get it. Unless you mean, that it should say "Custom actions" instead of "Custom commands"?

marcin-ptaszynski: Exactly that. user may not know that custom commands are generated from Custom Actions defined in https://github.com/cloudwan/gohan/blob/master/docs/source/schema.rst#custom-actions-schema

docs/source/commands.rst
```diff
@@ -246,3 +246,12 @@ it's name or id could be used:
   gohan client subnet create --network "Network Name"
   gohan client subnet create --network network-id
   gohan client subnet create --network_id network-id
+
+Custom commands
+---------------
+
+Any `custom actions <./schema.rst/#custom-actions-schema>`_ specified in schema are also supported as commands by gohan client and should be invoked as follows:
```
abonie: I don't really know anything about creating docs from .rst format, so someone should please verify, that this link doesn't break

marcin-ptaszynski: You can easily verify it with:
```
sudo pip install sphinx
cd ./docs
make html
```
Results can be found in build directory.

PR #290, opened 2016-09-26
Long polling (issue 288).

m-kostrzewa:
~~Early concept. Implemented long polling for /state/ changes.~~

~~edit: [WIP]~~
edit2: no longer WIP

### DIFF
--- a/examples/long_poll_example/README.md
+++ b/examples/long_poll_example/README.md
@@ -1,0 +1,209 @@
+long polling example
+---------------------
+
+In this example, we show how you can perform long polling of resources. The example uses curl.
+
+> **Note**: etcd must be running (on localhost:4001 by default). You can change this in gohan.yaml.
+
+long polling List API
+----------------------
+
+First, perform normal GET request using List API:
+
+```
+curl localhost:9091/v1.0/store/pets/ -i
+```
+
+Example response:
+
+```
+HTTP/1.1 200 OK
+Etag: 7b2f681a074dc89a5c918eb9ddbfec79
+...
+{
+  "pets": []
+}
+```
+
+In the response headers, you will notice ``Etag``:
+
+```
+Etag: 7b2f681a074dc89a5c918eb9ddbfec79
+```
+
+The value of Etag will be different depending on response content.
+Now, in a separate terminal, start a long polling GET request of list of resources, by providing a ``Long-Poll`` header with value of just received ``Etag`` header:
+
+```
+curl localhost:9091/v1.0/store/pets/ -H 'Long-Poll: 7b2f681a074dc89a5c918eb9ddbfec79' -i
+```
+
+This command should block. In another terminal create a new resource:
+
+```
+(in another terminal)
+curl http://localhost:9091/v1.0/store/pets/ -H "Content-Type: application/json" -X POST -d '{"name":"fluffy"}'  -i
+```
+
+Notice that the long polling request returned, and that it contains a new ``Etag``:
+
+```
+HTTP/1.1 200 OK
+Etag: 85d39266ed1e01c70265331848edd5d4
+...
+{
+  "pets": [
+    {
+      "description": null,
+      "id": "881e436d-2edc-4c28-97cb-ee448e0f2cc8",
+      "name": "fluffy",
+      "status": null,
+      "tenant_id": "admin"
+    }
+  ]
+}
+```
+
+This new ``Etag`` value can be used to perform new long polling request:
+
+```
+curl localhost:9091/v1.0/store/pets/ -H 'Long-Poll: 85d39266ed1e01c70265331848edd5d4' -i
+```
+
+Trying to modify an existing resource will also cause the long polling request to return:
+
+```
+(in another terminal)
+curl localhost:9091/v1.0/store/pets/881e436d-2edc-4c28-97cb-ee448e0f2cc8 -H "Content-Type: application/json" -X PUT -d '{"name":"modified_fluffy"}' -i
+```
+
+Example response:
+
+```
+HTTP/1.1 200 OK
+Etag: 323d37c537e500e8a3385a55466653be
+...
+{
+  "pets": [
+    {
+      "description": null,
+      "id": "881e436d-2edc-4c28-97cb-ee448e0f2cc8",
+      "name": "modified_fluffy",
+      "status": null,
+      "tenant_id": "admin"
+    }
+  ]
+}
+```
+
+Deletion of existing resource will also cause Gohan to respond (remember to use the new Etag in the request):
+
+```
+curl localhost:9091/v1.0/store/pets/ -H 'Long-Poll: 323d37c537e500e8a3385a55466653be' -i
+```
+
+```
+(in another terminal)
+curl http://localhost:9091/v1.0/store/pets/881e436d-2edc-4c28-97cb-ee448e0f2cc8 -X DELETE -i
+```
+
+Example response:
+
+```
+HTTP/1.1 200 OK
+Etag: 7b2f681a074dc89a5c918eb9ddbfec79
+...
+{
+  "pets": []
+}
+```
+
+long polling Show API
+----------------------
+
+To long poll a specific resource, it must already exist. So create a new resource:
+
+```
+(in another terminal)
+curl http://localhost:9091/v1.0/store/pets/ -H "Content-Type: application/json" -X POST -d '{"name":"reksio"}' -i
+```
+
+Use "id" field contained in the response to perform curl request of specific resource.
+
+```
+curl localhost:9091/v1.0/store/pets/5596c073-6cd2-448f-b862-1318421025d4 -i
+```
+
+Example response: 
+
+```
+HTTP/1.1 200 OK
+Etag: 564b7733ccc65ce66a6a99dcadf04e70
+...
+{
+  "pet": {
+    "description": null,
+    "id": "5596c073-6cd2-448f-b862-1318421025d4",
+    "name": "reksio",
+    "status": null,
+    "tenant_id": "admin"
+  }
+}
+```
+
+This will return an ``Etag`` associated with that specific resource. Use it to perform a long polling request:
+
+```
+curl localhost:9091/v1.0/store/pets/5596c073-6cd2-448f-b862-1318421025d4 -H 'Long-Poll: 564b7733ccc65ce66a6a99dcadf04e70' -i
+```
+
+Creating a new resource will not cause the request to return:
+
+```
+(in another terminal)
+curl  http://localhost:991/v1.0/store/pets/ -H "Content-Type: application/json" -X POST -d '{"name":"lajka"}' -i
+````
+
+But modyfing the long-polled resource - will:
+
+```
+(in another terminal)
+curl localhost:9091/v1.0/store/pets/5596c073-6cd2-448f-b862-1318421025d4 -H "Content-Type: application/json" -X PUT -d '{"name":"modified_reksio"}' -i
+```
+
+Example response:
+
+```
+HTTP/1.1 200 OK
+Etag: 0b939757e0a3896117449d7c7f502d79
+...
+{
+  "pet": {
+    "description": null,
+    "id": "5596c073-6cd2-448f-b862-1318421025d4",
+    "name": "modified_reksio",
+    "status": null,
+    "tenant_id": "admin"
+  }
+}
+```
+
+Perform a new long polled request using new Etag:
+
+```
+curl localhost:9091/v1.0/store/pets/5596c073-6cd2-448f-b862-1318421025d4 -H 'Long-Poll: 0b939757e0a3896117449d7c7f502d79' -i
+```
+
+Deleting queried resource also causes the request to return:
+
+```
+(in another terminal)
+curl localhost:9091/v1.0/store/pets/5596c073-6cd2-448f-b862-1318421025d4 -X DELETE -i
+```
+
+Example response:
+
+```
+HTTP/1.1 404 Not Found
+{"error":""}
+```
--- a/server/api.go
+++ b/server/api.go
@@ -212,10 +212,29 @@ func MapRouteBySchema(server *Server, dataStore db.DB, s *schema.Schema) {
 	getPluralFunc := func(w http.ResponseWriter, r *http.Request, p martini.Params, identityService middleware.IdentityService, context middleware.Context) {
 		addJSONContentTypeHeader(w)
 		fillInContext(context, dataStore, r, w, s, p, server.sync, identityService, server.queue)
-		if err := resources.GetMultipleResources(context, dataStore, s, r.URL.Query()); err != nil {
+
+		var newEtag string
+		var err error
+		if r.Header.Get(LongPollHeader) != "" {
+			oldEtag := r.Header.Get(LongPollHeader)
+			getResource := func(context middleware.Context) error {
+				return resources.GetMultipleResources(context, dataStore, s, r.URL.Query())
+			}
+			newEtag, err = server.longPoll.GetOrWait(r.URL.Path, oldEtag, context, getResource, calculateResponseEtag)
+		} else {
+			err = resources.GetMultipleResources(context, dataStore, s, r.URL.Query())
+			if err == nil {
+				newEtag = calculateResponseEtag(context)
+			}
+		}
+
+		if err != nil {
 			handleError(w, err)
 			return
 		}
+
+		w.Header().Add("Cache-Control", "no-cache")
+		w.Header().Add(LongPollEtag, newEtag)
 		w.Header().Add("X-Total-Count", fmt.Sprint(context["total"]))
 		routes.ServeJson(w, context["response"])
 	}
@@ -230,10 +249,29 @@ func MapRouteBySchema(server *Server, dataStore db.DB, s *schema.Schema) {
 		addJSONContentTypeHeader(w)
 		fillInContext(context, dataStore, r, w, s, p, server.sync, identityService, server.queue)
 		id := p["id"]
-		if err := resources.GetSingleResource(context, dataStore, s, id); err != nil {
+
+		var newEtag string
+		var err error
+		if r.Header.Get(LongPollHeader) != "" {
+			oldEtag := r.Header.Get(LongPollHeader)
+			getResource := func(context middleware.Context) error {
+				return resources.GetSingleResource(context, dataStore, s, id)
+			}
+			newEtag, err = server.longPoll.GetOrWait(r.URL.Path, oldEtag, context, getResource, calculateResponseEtag)
+		} else {
+			err = resources.GetSingleResource(context, dataStore, s, id)
+			if err == nil {
+				newEtag = calculateResponseEtag(context)
+			}
+		}
+
+		if err != nil {
 			handleError(w, err)
 			return
 		}
+
+		w.Header().Add("Cache-Control", "no-cache, no-store")
+		w.Header().Add(LongPollEtag, newEtag)
 		routes.ServeJson(w, context["response"])
 	}
 	route.Get(singleURL, middleware.Authorization(schema.ActionRead), getSingleFunc)
--- a/server/message_dispatch.go
+++ b/server/message_dispatch.go
@@ -1,0 +1,154 @@
+// Copyright (C) 2015 NTT Innovation Institute, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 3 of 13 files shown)


server/api.go
```diff
@@ -234,6 +242,11 @@ func MapRouteBySchema(server *Server, dataStore db.DB, s *schema.Schema) {
 			handleError(w, err)
 			return
 		}
+
+		if r.Header.Get("Long-Poll") == "true" {
```
marcin-ptaszynski: ditto

server/api.go
```diff
@@ -216,6 +219,11 @@ func MapRouteBySchema(server *Server, dataStore db.DB, s *schema.Schema) {
...
```
marcin-ptaszynski: if it's non standard header, we should prefix it with "X-" or something.

server/api.go
```diff
@@ -388,6 +401,37 @@ func MapRouteBySchema(server *Server, dataStore db.DB, s *schema.Schema) {
 	}
 }
 
+func waitForNewChanges(w http.ResponseWriter, r *http.Request, context middleware.Context) {
+	etag := calculateResponseEtag(context)
+	if etag == r.Header.Get("If-None-Match") {
```
marcin-ptaszynski: Perhaps using If-None-Match is not such a good idea here... See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match . Spec would force us to return 304, while we want to Long Poll on this ETag and eventually get next version. It would also interfere with caches. We need something like: "X-Long-Poll-If-Mach: xxxx".

server/api.go
```diff
@@ -388,6 +401,37 @@ func MapRouteBySchema(server *Server, dataStore db.DB, s *schema.Schema) {
...
+
+func calculateResponseEtag(context middleware.Context) string {
+	hash := md5.New()
+	responseBytes, err := json.Marshal(context["response"])
+	if err == nil {
+		hash.Write(responseBytes)
+	} else {
+		randomBytes := make([]byte, 8)
+		rand.Read(randomBytes)
+		hash.Write(randomBytes)
```
marcin-ptaszynski: why? This doesn't help us.

server/sync.go
```diff
@@ -54,6 +56,54 @@ func transactionCommitInformer() chan int {
...
+
+	url := strings.TrimPrefix(fullKey, "/state")
+	key := strings.Split(url, "/")
+	for i := 0; i < len(key); i++ {
+		subkey := strings.Join(key[:i+1], "/")
+		log.Critical("Notifying key %s", subkey)
+
+		subsListMutex.Lock()
+		subs := keyUpdateSubscribers(subkey)
+		for e := subs.Front(); e != nil; e = e.Next() {
```
marcin-ptaszynski: https://golang.org/pkg/sync/#Cond and Broadcast would fit more.

server/sync.go
```diff
@@ -54,6 +56,54 @@ func transactionCommitInformer() chan int {
...
+		keyUpdated = make(map[string]*list.List)
+	}
+	if _, ok := keyUpdated[key]; !ok {
+		log.Critical("Creating update sub list for new key %s", key)
+		keyUpdated[key] = list.New()
+	}
+	return keyUpdated[key]
+}
+
+func notifyKeyUpdateSubscribers(fullKey string) {
```
marcin-ptaszynski: This will work only for single node. It should be distributed to other HA nodes as well.

server/api.go
m-kostrzewa: using X- is no longer recommended (http://stackoverflow.com/a/3561399) apparently, but sure

nati:
I would rather prefer using etcd directory or just let gohan work as an authentication proxy.
so it is better to be same API with etcd watch API.

nati:
CI test upgraded. could you rebase your patch?

server/message_dispatch.go
```diff
@@ -0,0 +1,115 @@
...
+package server
+
+import (
+	"strings"
+	"sync"
+)
+
+type MessageDispatch struct {
+	input       chan string
+	output      map[string][]chan string
```
marcin-ptaszynski: Not sure, why prefer iterating over array of channels instead of use simpler `Cond` structure.

Keep in mind, that sending by default is not asynchronous and if channel is not buffered (and currently, it's not), this can block dispatcher loop when some client disconnects. See example:
https://play.golang.org/p/-I9eVBimNA
https://play.golang.org/p/1yLFOtu1Cy

server/api.go
```diff
@@ -234,6 +249,19 @@ func MapRouteBySchema(server *Server, dataStore db.DB, s *schema.Schema) {
...
+		if r.Header.Get("Long-Poll") == "true" {
+			waitForNewChanges(w, r, context)
+			delete(context, "response")
+			log.Critical("Getting second time")
+			if err := resources.GetMultipleResources(context, dataStore, s, r.URL.Query()); err != nil {
+				handleError(w, err)
+				return
+			}
+			newEtag := calculateResponseEtag(context)
+			w.Header().Add("Long-Poll-Etag", newEtag)
```
m-kostrzewa: If-None-Match and ETags could mess with standard caching. You actually pointed it out during last review IIRC. That's why we decided to go with non-standard header

server/message_dispatch.go
m-kostrzewa: We actually had a discussion about this and couldn't figure out a good way to do it using conds. It's hard to prevent notification-misses and subscription-misses using Conds without a memory leak - basically, we can't know when to "cleanup" our Conds without multiple mutexes. That's why we implemented our own broadcasting mechanism. Channels not being buffered is definitely a problem though.

marcin-ptaszynski: Perhaps I don't get it, but first subscriber can get Mutex and Insert Cond in map. When event arrives, dispatcher Broadcasts and removes Cond from map. GC removes it after all client goroutines remove it from scope.

koziol: You're right about buffered channels, I'll apply this change. 

As for using Conds, I tend to prefer high-level abstractions (like chan) over low-level primitives (like Cond+Mutex+buffer). As far as I understand, what you're suggesting is actually encapsulating a Cond (plus helper variables) into a chan. IMO, it'll be a better practice to use ready-made library types instead.

What's more, when designing a prototype with manual synchronization we've came up with a code that was hard to understand and analyze. The problem were sequences of interleaved lock-unlock calls on different objects.

marcin-ptaszynski: Hi @koziol,

> As for using Conds, I tend to prefer high-level abstractions (like chan) over low-level primitives (like Cond+Mutex+buffer).

What makes you think, Cond is low level abstraction, while chan is high-level? chan is builtin, while Cond is standard library, which would rather point to the opposite conclusion. In any case, we're talking about:

- Mutex
- map of lists for grouping chans
- push list in wait method
- for loop in broadcast method

vs.

- Mutex
- map of Conds

If you have a look at https://golang.org/src/sync/cond.go?s=1407:1428#L43 , there is actually a similar semantics to what you're proposing, except it's already abstracted.

> As far as I understand, what you're suggesting is actually encapsulating a Cond (plus helper variables) into a chan.

I'm not sure, I understand this sentence. Could you clarify? We're interested in waking up groups of goroutines. What we're doing now, is waiting on a channel created specifically for receiving single notification signal and not caring about it's content. If anything, I think it's exactly the reverse case, where we're trying to force channels to behave like Cond.

> What's more, when designing a prototype with manual synchronization we've came up with a code that was hard to understand and analyze. The problem were sequences of interleaved lock-unlock calls on different objects.

I think I still don't see a problem. I've done small demo, which shows how straight-forward it is to use https://play.golang.org/p/I5LhD7oyjy . Could you check this out and explain what's the problem with this approach?

koziol: Your demo works equally well, but it's slightly less (unit-)testable. I'm talking about TCs were you want to send Broadcast only after a listener called Wait (but this is not a major issue).

Anyway, I've changed the implementation back to Conds like you suggested. Let's keep the discussion about high/low level concurrency primitives offline.

marcin-ptaszynski:
is it still WIP?

m-kostrzewa:
Not WIP anymore.

nati:
Could you rebase and make CI working?

m-kostrzewa:
Added example + doc

examples/long_poll_example/README.md
```diff
@@ -0,0 +1,137 @@
...
+...
+{
+  "pets": []
+}
+```
+
+In the response headers, you will notice ``Etag``:
+
+```
+Etag: "7b2f681a074dc89a5c918eb9ddbfec79"
```
nati: Could you tell me why we need double quote? I think we shouldn't need this

examples/long_poll_example/README.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+This new ``Etag`` value can be used to perform new long polling request.
+
+> Note: long polling requests will return upon any modifying operation on queried resource, which can be creation, deletion or modification.
+
+long polling single resource
+-----------------------------
+
+Notice, how long polling ``/v1.0/store/pets/`` path returns when a child object is created.
```
nati: I don't get meaning of this sentence.

Developers can watch a List API and a Show API. If a developer watch List API, Gohan server returns diff on list request result. If a developer watch Show API, Gohan server returns update on a single resource. 

Is this what you meant?

examples/long_poll_example/README.md
```diff
@@ -0,0 +1,137 @@
...
+    "id": "7d941380-7ca5-43bc-9246-7020ed7efcaf",
+    "name": "new_fluffy",
+    "status": null,
+    "tenant_id": "admin"
+  }
+}
+```
+
+long polling ancestor
+--------------------
```
nati: We should make this configurable list etcd API.

examples/long_poll_example/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+```
+curl  http://localhost:991/v1.0/store/pets/ -H "Content-Type: application/json" -X POST -d '{"name":"lajka"}' -i
+````
+
+But modyfing the queried resource - will:
+
+```
+curl localhost:9091/v1.0/store/pets/7d941380-7ca5-43bc-9246-7020ed7efcaf -H "Content-Type: application/json" -X PUT -d '{"name":"modified_fluffy"}' -i
+```
```
nati: Could you add DELETE example? I've tried delete but it wasn't working..

nati: for both of List and Show API

m-kostrzewa:
Removed double quotes in Etag value. Improved docs + example.

nati:
Hi I'm still thinking replicating same api as etcd is best.
We can't know which element got deleted by current API..

stepek:
From WebUI point sends diff from by long poll request its bad.
Because, if we got table with first 10 elements and we remove the second one. We need to fetch all data from gohan after that. In this implementation of long poll api we get new data from API after action and we don't need to fetch data second time. 
The save will be with adding items to table.

So, maybe will be good to, add this feature and after that add feature with diffs?

@thayashi What you think about that?

m-kostrzewa:
Maybe we should implement etcd-like API as a separate feature then?

nati:
Supporting two API is a burden.. so I think we need only etcd-like API.

thayashi:
I don't get what's the problem of current implementation. Also how  etcd-like API will look like? Is WebUI able to know only there is some update, then needs to fetch the all data?

nati:
@thayashi 

Current API returns diff, so we should check entire local cache when a resource get deleted.

You can know what action happened for the resource using etcd-like api.
https://coreos.com/etcd/docs/latest/api.html

marcin-ptaszynski:
Hi @nati ,
Proxying etcd might seem easy for Gohan, but consider following:
- API won't be stable when Gohan migrates to etcd v3
- It requires to implement completely new event (etcd actions) handling on UI side, while above PR requires addition of 2 headers. Minimal set of changes is more healthy than completely alien API just for handling Long-Poll.
- It won't be REST anymore. Requesting List of resources should return a List, not action diff on KV store. Perhaps it should be separate *event* endpoint API with subscription options, but then we lose simplicity (see above).

nati:
- yes. we need also migrate etcd v3.
- i think having action helps to implement clientside, otherwise UI code should maintain diff
- yes it won't be REST when we support something like watch API.

marcin-ptaszynski:
@nati ,
Additional argument against etcd proxy is that in current implementation it's possible for resource to change for example it's operational state without updating corresponding `/config` path in etcd, since it's designed as config for Worker, not internal state. Internal state is held in DB.

thayashi:
Does checking entire local cache when deleted cause serious performance issue? If not, I think this implementation is more convenient for WebUI.

nati:
We can't support act-act model for watch API without etcd. so using etcd is a mandatory option.
Even if it is possible to generate diff in client side, I have strong opinion we should try best to provide most convenient API for client.

m-kostrzewa:
There are a couple issues that complicate the current implementation if etcd-like API were to be implemented. 

It's not clear how to handle missed updates. Say, a client has resource A1 and doesn't long poll for a while, but during that while, the resource is modified to A2. Now, when the client requests resource A, what should be the returned value? We can't return a diff, we can only return the whole resource. This creates a need to introduce resource versioning... which is there, but: 

a) it requires a separate call to etcd to check resource version,

b) how do we version lists? We would need to add control number to etcd directories?,

c) how do we version paginated lists? 

Alternatively, we can come up with a new action, like "update_missed", and return the whole resource, but this complicates client logic. And at that point, we might as well stick to current implementation.

More on pagination. If a user has paginated resources A, B, C, and during long poll, resource C is deleted and resource D "takes its place" on the list, should Gohan return "modified C->B" as the action? Or should it return "deleted"? Same goes for other events that could mess with paginated content. 

etcd watch API doesn't have to deal with lists, but Gohan does. I don't see a clear way to handle them in watch API manner.

nati:
Etcd supports all except pagination, and I believe we can't support pagination for watch API.
Bascially, etcd maintains globally unique version number called etcd index. 
The case you mentioned, you can get full update history from specified etcd index.
see https://coreos.com/etcd/docs/latest/api.html

or Martin knows details.

m-kostrzewa:
I talked to @stepek and he thinks that current implementation is easy to implement client-side, because it returns the same resource as normal Show/List API does, not diff. So client/WebUI doesn't have to maintain or deal with any diffs. 

We also talked about Gohan acting like etcd proxy and it would require a very different implementation than what we have. Kamil also mentioned that not having support for paginated lists could be problematic. 

Having that in mind, maybe we should mark current API as "unstable" and implement "etcd proxy" in another PR, so that frontend guys can start working on it? Simply switching from short polling is a big step.

nati:
how do you support pagenation by spec?

let's say we have resource R1 to R10

R1 R2 R3 R4 R5 with order by a property X.

then a user updates a resource RA. 
then we got 

R1 R2 RA R3 R4 R5

In this case, we should send an update for a user who is watching resource with limit2, offset2 and order by the property X.
This makes huge computation overhead, so I believe we shouldn't support this practically.

Again, I have a strong preference for sending diff from experiences to implement agents.
Changing implementation later is so hard, so I can't support that too.

m-kostrzewa:
I can see ways in which it could be done using current hash+notification based mechanism, though. It's more flexible and we don't have to rely on etcd indexing mechanism which we have no control over and proves problematic, especially with the 1000 events limit.

Transitioning to etcd based API means abandonning pagination, which is used frequently and I believe short polling using List API was the main reason long polling feature was requested in the first place.

Also, problems arise when using Gohan as etcd authentication proxy with tenant isolation and policy filtering.

nati:
We have the same limitation if we relying on etcd, and we already know how to avoid such limitation.

I don't understand how we are supporting pagination case with ordering. it looks like just broadcasting everything on the specific path.

Note that, I don't saying implementing "etcd-proxy", but suggesting using etcd like API style such as having event type for each event.

marcin-ptaszynski:
@m-kostrzewa , 

> Transitioning to etcd based API means abandonning pagination, which is used frequently and I believe short polling using List API was the main reason long polling feature was requested in the first place.

@thayashi , could you confirm this? Is Long polll on a single resource any use to you?

@nati , @m-kostrzewa Another way to deal with List view of Long Poll is:
- Subscribe on updates for every resource separately (feasibility depends on number of resources shown at) with above implementation.
- List Long Poll has changed semantics and it returns new (refreshed) view only when resource matching filter criteria (usually just tenant_id) is added/removed, but not updated. This would cause database refresh even when change doesn't affect observable window defined by pagination options, but it doesn't need to return to client after comparing the ids on the resulting list.
This allows us to combine simple API with watch for single resource and collection wise events with pagination while retaining simple API and little computational overhead.

nati:
@marcin-ptaszynski 

Subscribe on updates for every resource separately with above implementation. Feasibility depends on number of resources shown on single page.

<- I don't think this is a feasible way on the aspect of connection management. if we show 100 page, then 1 client requests 100 session. and Note that webui isn't only client for UI but also client agents owned by tenants.

marcin-ptaszynski:
Hi @nati ,

> <- I don't think this is a feasible way on the aspect of connection management. if we show 100 page, then 1 client requests 100 session. and Note that webui isn't only client for UI but also client agents owned by tenants.

I'm afraid, you're correct and this is not feasible solution.

I think we can extend current behavior to support additions/updates and removals on List long-poll by:
1. Notifying sleeping long-poll goroutine when any change happens to resource matching all filters (tenant_id, name, parent_resource_id, etc). This can be achieved in lightweight manner by maintaining tree structure with filters as keys and Event objects as values. Updated resource needs to iterate over it's properties and trigger that many events (some may subscribe for tenant_id=x, other to tenant_id=x and parent_id=y). This should limit number of refreshes in step 2.
2. After goroutine is awakened it will refresh db view with additional pagination options (sort, limit, offset) and compare if anything changed by hashing the response.
3. Returns if anything changed in the view.

 It may still be heavy when tenant has many resources of the same type, which exist in flat space, but efficiently handling paginated view would require full db replica in memory. etcd has it easier, since it either uses hierarchical key structure or key range query, which is less than we're expected to do.

What do you think about it?

nati:
I like that idea, but I would like to server side simplified as much as we can.

I see there are UI use-case and Agent use-case.

For the UI use-case, A single notification which tells client there is some update in list request, then UI can simply reissue list API request with filters and pagination parameters.
For agent use-cases, we should know exact history per resource on list API request. 
It is better we can know a resource deleted explicit manner, otherwise an agent should maintain all caches of server side DB, and this will a complicates implementation.

I believe implementing etcd like API can support both of this. We can create one Go routine per client, and it watches specific key on the etcd. and we can apply policy per resource.

marcin-ptaszynski:
> For the UI use-case, A single notification which tells client there is some update in list request, then UI can simply reissue list API request with filters and pagination parameters.

This may result in high throughput if we flatten the key-space per resource type by default. Instead of doing db refresh and comparing it in Gohan, we'd reissue refresh via network regardless of change.

> For agent use-cases, we should know exact history per resource on list API request.
It is better we can know a resource deleted explicit manner, otherwise an agent should maintain all caches of server side DB, and this will a complicates implementation.

@m-kostrzewa was working for UI usecase and IMO, it makes sense to have similar API to normal List operation and not require UI app to maintain sync with Gohan. Otherwise, we're talking not about a small feature, but major refactor. In our system we use etcd directly for Agents and if this is any help, etcd API since v3.1 will support dynamic RBAC Auth module.

nati:
You are right. 
I think it is a good idea to apply filters in server side. 

In both cases, We would end up making tons of requests for DB for each update.
Once again, I believe supporting to watch capability with pagination isn't feasible for large resources. We should only support this with limiting a total number of resources. 

I'm not sure how we can apply our policy and extension if we let client directly connect agent.

marcin-ptaszynski:
Hi @nati ,

How do you want to proceed further regarding this issue? Can we separate UI usecase and Agent usecase and mark current API as unstable, after we implement List handling described above? Then, when etcd like Agent API comes in UI can implement sync with Gohan using new API.

Regarding Agent API, I think I'm not entirely briefed in usecase, but I assume, it's worker like process with limited trust, so we need policing. IMO, it's tough, but it should be possible to provide tenant isolation by filtering out rpc messages.  Definitely, not pagination, etc.

m-kostrzewa:
@nati how should we proceed regarding this issue? Should this PR be entirely for the UI use-case? If so, what changes need to be made for it to be accepted?

nati:
It won't make sense to have separate API for UI and the tenant side agent. so the API should suit both of that.

Also, Once we accept it as an API, we end up to support long term. so we can't accept it.

nati:
close this PR due to inactivity. plz reopen when you needed

PR #308, opened 2016-10-11
add support for path whitelists

przemyslaw-dobrowolski-cl:
a test is provided in examples/path_whitelist


nati:
This is CI side failure. I'm fixing it.

nati:
I would rather add extend policy framework side.
how about to create nobody user like unix?

examples/path_whitelist/README.md
```diff
@@ -0,0 +1,34 @@
...
+In this example, we show how to define a list of paths that never
+require an authorization. A single path is defined as a regular
+expression (regex).
+
+In order to define a whitelisted path, add the following entry
+in the configuration file:
+
+``` yaml
+# path whitelist configuration
+path_whitelist:
```
marcin-ptaszynski: As I stated in previous post, `path_whitelist`, IMO isn't descriptive enough. Additionally, I don't believe this config should be in config root. Have you considered putting it in `route` section.

Actually, since we already have builtin Policy resource, it would make most sense from app developer perspective to simply put this information there. I understand, it may require additional refactor, so perhaps we should skip it for now. @nati , what do you think about it?

examples/path_whitelist/README.md
```diff
@@ -0,0 +1,34 @@
+Gohan path whitelist example
```
marcin-ptaszynski: please, put into documentation found in `./doc`

examples/path_whitelist/README.md
przemyslaw-dobrowolski-cl: How about creating a route section (which covers all settings related to paths) and add a property called: noauth_paths ?

examples/path_whitelist/README.md
przemyslaw-dobrowolski-cl: done

przemyslaw-dobrowolski-cl:
Uploaded a new version with most of the comments applied.

examples/path_whitelist/README.md
marcin-ptaszynski: I think this should be fine.

przemyslaw-dobrowolski-cl:
@nati: Creating a new user is an alternative implementation but my concern is that having a nobody user will still require the client to login but without a password (equivalently, any password). As far as I know the idea was to abandon the login process at all. Is it right?

nati:
@przemyslaw-dobrowolski-cl  We may skip authentication for the path where nobody role allowed in the policy, then we can make a unauthenticated user as nobody.

PR #314, opened 2016-10-12
Make example Go extension work again

p-kozlowski:
The Go extension example was not working since Motto was introduced. The call to require in Otto has been replaced by require in Motto. To support both NPM and Go extensions, we should do the lookup in both registers.

I've decided that Motto (NPM) lookup should take precedence over Otto (Go extensions) because it's more widely known.


vozhyk-:
Please merge in a consistent way - either always making merge commits, or always cherry-picking the commit. The former is preferable because then I can see the PR number for a commit in the git log (and not just on the GitHub commit page).

PR #325, opened 2016-11-04
OpenAPI generator:

Zimnx:
OpenApi API generated with example schemas (for example keystone) wasn't valid on http://editor.swagger.io/

Current implementation remove extended properties only on first and second depth of property tree. Proposed version iterate recursively over whole tree and removes them on every level.
It also fixes default value of enum type fields and remove empty 'required' fields.


nati:
thanks!

PR #328, opened 2016-11-08
[WIP] Check sync before running cron.

m-kostrzewa:
This change introduces a panic if sync backend is not configured but we want to run cron. This fixes ESI-9506.

WIP, because I'm not sure how to write tests for this change. Should I create additional server config .yamls and create a new test suite file? @morrisson


nati:
i'm ok with this change without UT. 
Could you rebase?

PR #331, opened 2016-11-16
Ability to specify in schema if column should be indexed in DB

Zimnx:


nati:
Thanks I like this patch but could you also update the doc?

Zimnx:
I've already mentioned about this field in schema.md. Should I provide any additional information anywhere in docs? I think this field is pretty straightforward.

Zimnx:
Hi, I've added another commit. In new one user can specify which column relation references. Until now relation always referenced "id" column, which not always fits user needs. Now it can be parametrized. I've also updated docs with info about this field.

nati:
ok it looks like I missed doc change in the first commit. LGTM

PR #334, opened 2016-11-29
Add etcd v3 support

yudai:
Support etcd v3 API:

* A new config property `sync` is added
   * Provide `etcdv3` to this property to use the etcd v3 API
   * Provide `etcd` to this property to use the etcd v2 API, which the current implementation uses
   * When missing, the default value is `etcd` for backward compatibility

* Fixed some minor bugs:
   * Removed double slashes like `/config//v1.0/....`

* Use the port 2379 in documents and examples
   * 4001 is outdated


yudai:
I'll add some tests.

yudai:
Added tests. Ready to review

sync/etcdv3/etcd.go
```diff
@@ -0,0 +1,324 @@
...
+	sync.processID = hostname + uuid.NewV4().String()
+	return sync, nil
+}
+
+//Update sync update sync
+//When jsonString is empty, this method do nothing because
+//etcd v3 doesn't support directories.
+func (s *Sync) Update(key, jsonString string) error {
+	var err error
+	if jsonString == "" {
```
nati: i prefer returning here. 
if jsonString == "" {
return nil
}

yudai: Done

yudai:
Squashed!

PR #335, opened 2016-12-06
Add sync_plain and sync_property

yudai:
New metadata for schemas:

`sync_plain (boolean)`

Write plain data, which is not JSON marshaled, to the sync backend if true.

`sync_property (string)`

Write only the value of the specified property to the sync backend.

These options are useful when a listener on etcd, which comes from a different projects, doesn't support the format of Gohan sync payload.

Also removing format validation on sync_key_template, because requiring always `{{id}}` in templates decreases flexibility. (Gohan should support composite unique btw)


yudai:
will add some tests later.

nati:
This looks good except following

- Could you add example how to use this in the doc?
- Unit test as you mentioned

yudai:
Added tests and a bit more detailed example to the doc.

PR #337, opened 2016-12-12
Fixed logging when module is not loaded in otto or motto

wsikorska-codilime:
We need more logs to debug issues with loading modules in otto or motto. This PR just adds more logs.

### DIFF
--- a/extension/otto/gohan_core.go
+++ b/extension/otto/gohan_core.go
@@ -25,6 +25,7 @@ import (
 	"github.com/ddliu/motto"
 	"io/ioutil"
 	"strings"
+	"fmt"
 )
 
 var log = logging.MustGetLogger(l.GetModuleName())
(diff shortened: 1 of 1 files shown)


extension/otto/gohan_core.go
```diff
@@ -193,12 +195,15 @@ func requireFromOtto(moduleName string, vm *otto.Otto) (otto.Value, error) {
 }
 
 func requireFromMotto(moduleName string, vm *motto.Motto) (otto.Value, error) {
+	log.Info(fmt.Sprintf("Loading module %s from motto", moduleName))
```
vozhyk-: This can produce too many messages. What about `log.Debug`?

nati: I agree

nati:
@wsikorska-codilime How about chaning level to debug?

nati:
Could you push rebased version? CI should work now.

PR #343, opened 2017-01-04
Fixed loading npm modules with entrypoint without js extension

Zimnx:
Some NPMs (for example: big-integer) have entrypoint specified without ".js" suffix.  
Correct entrypoint for for this case was node_modules/big-integer/BigInteger.js.  
Gohan tried to load node_modules/big-integer/BigInteger/index.js which doesn't exists

### DIFF
--- a/extension/otto/gohan_core.go
+++ b/extension/otto/gohan_core.go
@@ -19,13 +19,13 @@ import (
 	"github.com/cloudwan/gohan/schema"
 	"github.com/op/go-logging"
 	"github.com/robertkrimen/otto"
-
 	l "github.com/cloudwan/gohan/log"
 	"github.com/cloudwan/gohan/util"
 	"github.com/ddliu/motto"
 	"io/ioutil"
 	"strings"
 	"fmt"
+	"os"
 )
 
 var log = logging.MustGetLogger(l.GetModuleName())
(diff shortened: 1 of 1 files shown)


nati:
LGTM but we need CI get working

nati:
CI get fixed, so could you push this branch again if you need to rebase or recreate pull request?

Zimnx:
Done

extension/otto/gohan_core.go
```diff
@@ -217,7 +222,13 @@ func loadNPMModules() {
 				break
 			}
 			if !strings.HasSuffix(module, ".js") {
-				module = module + "/index.js"
+				if moduleFile, err := os.Stat(module); err == nil {
```
marcin-ptaszynski: not sure if this is correct computation, although it will work most of the time. IMO, we should follow npm spec and lookup package.json/main. See https://docs.npmjs.com/files/package.json#main , source https://github.com/nodejs/node/blob/master/lib/module.js

Zimnx:
I've changed way we choose entry point for module according to [official code](https://github.com/nodejs/node/blob/master/lib/module.js#L65).

Zimnx:
@nati kind remainder :slightly_smiling_face:

vozhyk-:
A related change to motto (handles the same case, but when `require` is called from a module): https://github.com/vozhyk-/motto/commit/38e3168541f8dfd5516231cf313f1fe24803bec3.

nati:
@Zimnx could you resolve conflict?

Zimnx:
Conflict resolved

Issue #340, opened 2016-12-16
File handle leak when running JS extension tests


p-kozlowski:
When running extension tests, I've come across a sqlite file handles link. Before each test, a new sqlite DB is created and after each test the file is deleted, but the 2 DB connection (and file handle) remains open. One connection is main one that the tests use, the second one is used only to load schemas.

When we reach ~500 test cases, an error message `12:30:33.112 extest ERROR      FAIL    test_XYZ.js: Failed to read file 'test_XYZ.js': open test_XYZ.js: too many open files` is reported. As a result, all TCs after ~500th fail with "too many open files" error message. One possible workaround is to play with ulimit, but I haven't tried that.

One could watch the increasing number of open handles by running ```gohan test_ex``` in one terminal and watch ```lsof -p `pidof gohan` | wc -l``` in another one. Individual handles could be shown by running ```lsof -p `pidof gohan` | awk '{ print $9 }' | sort | uniq -c | sort -n```

PR #351, opened 2017-01-10
Close DB connections on JS tests teardown (#340), go-sqlite3 updated

p-kozlowski:


db/file/file.go
```diff
@@ -54,6 +54,10 @@ func (db *DB) Connect(format, conn string, maxOpenConn int) error {
 	return nil
 }
 
+func (db *DB) Close() {
+	// nothing to do
```
przemyslaw-dobrowolski-cl: is flush not needed for file?

p-kozlowski: No, because there are no open file handles in file db.

For instance, have a look at any of mutating methods (like func (tx *Transaction) Update(resource *schema.Resource) error). First the file is read to memory (using db.load()), then the mutation takes place in-memory, and finally the file is written back (using db.write()).

przemyslaw-dobrowolski-cl: ok, I see

p-kozlowski:
PR rebased and tests fixed.

PR #360, opened 2017-01-24
Register routes for custom actions with parent prefix

yudai:
When a schema with a parent resource type has custom actions,
add a route for the actions with the parent URL prefix.


yudai:
will add some tests.

yudai:
added tests.

Issue #361, opened 2017-01-26
Allow setting maximum execution time for builtins


przemyslaw-dobrowolski-cl:
If an extension is called that uses builtins like gohan_sync_watch, the time spent in the builtin is included in the total time spent in the extension. Since, the default time limit for an extension execution is 30 secs, it effectively disallows long-polling from extensions with an arbitrary timeout. In large scale solutions, 30 secs long-polling is not enough.

A proposed change is to add a configuration option 'builtin-timelimit' and to not include time spent in a builtin to the total execution time of the executing extension. A time spent in a builtin should counted separately and be limited according to the configuration setting mentioned above.

nati:
I would rather prefer set timeout for each extension.

marcin-ptaszynski:
@przemyslaw-dobrowolski-cl , IMO overly complicated method without clear goal. If we want to support long poll built on gohan_sync_watch, what does it matter, how long we spend in js code and builtin code. I think developer is more interested in total value anyway.

@nati , do you suggest using per (path template, method) tuple override? Or do you want to extend extension schema to provide optional timeout property (which would effectively set different timeout per loaded file)? This may be more difficult, since all extensions are executed from js.

Another idea is to place override builtin in extension, although it may be difficult to manage later.

nati:
@marcin-ptaszynski  my suggestion is to extend extension schema to have own timeout.
I believe this is simple to implement since we can override default value by this new value when we run extension code.

przemyslaw-dobrowolski-cl:
I'm not sure if in the proposed implementation it is possible to determine the maximum running time for a particular extension in HandleEvent() / otto.go. Assuming that there is a declaration in schema/extensions section of the maximum running time for all extensions declared in this schema, one would need to track in Environment a mapping from an event type and a list of registered event receivers to the maximum running time. Am I missing something?

nati:
@marcin-ptaszynski  @przemyslaw-dobrowolski-cl 
I got what you meant after rechecking code.

This is the code we implement timeout.
https://github.com/cloudwan/gohan/blob/master/extension/otto/otto.go#L184

so I believe it is trivial change if we set time per path & event.

PR #369, opened 2017-02-03
Time limit overrides for extensions

przemyslaw-dobrowolski-cl:
This commit adds a configuration option to setup orverrides
for the default time limit for running an extension. A single
setting consists of a regex path, a regex event name and
a time duration.

https://github.com/cloudwan/gohan/issues/361

### DIFF
--- a/etc/gohan.yaml
+++ b/etc/gohan.yaml
@@ -68,3 +68,12 @@ extension:
   - gohanscript
  # - javascript
   - go
+
+  # default time limit in secs for running extension
+  timelimit: 10
+
+  # time limit overrides in secs for running particular extension in path
+  timelimits:
+  - path: /v1.0/store/orders
+    event: pre_*
+    timelimit: 1
--- a/extension/otto/otto.go
+++ b/extension/otto/otto.go
@@ -31,6 +31,8 @@ import (
 
 	"reflect"
 	//Import otto underscore lib
+	"regexp"
+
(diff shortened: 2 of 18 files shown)


extension/otto/otto.go
```diff
@@ -128,10 +129,15 @@ func (env *Environment) LoadExtensionsForPath(extensions []*schema.Extension, pa
 			if err != nil {
 				return err
 			}
-
 		}
 	}
-
+	// setup time limits for matching extensions
```
p-kozlowski: code duplicated with extension/gohanscript/env.go

etc/gohan.yaml
```diff
@@ -67,4 +67,13 @@ extension:
...
+  - go
+
+  # default time limit in secs for running extension
+  timelimit: 10
+
+  # time limit overrides in secs for running particular extension in path
+  timelimits:
+  - path: /v1.0/store/orders
+    event: pre_*
+    timelimit: 1
```
p-kozlowski: without looking through the source code it's hard to guess what is the unit here. timelimit_seconds?

przemyslaw-dobrowolski-cl: If it wouldn't be for the consistency with the existing field 'timelimit' I'd change it according to your suggestion. IMO, its better to copy the problematic naming in this case.

przemyslaw-dobrowolski-cl:
@nati  a kind remainder about this PR

Issue #359, opened 2017-01-24
Add support for 'float' type in schemas


przemyslaw-dobrowolski-cl:
According to documentation:
http://gohan.cloudwan.io/gohan/schema.html?highlight=number#type-integer-or-type-number
there is no support for floating-point types. This is problematic if one needs to convert schema
to an openapi file which supports floats. Also, floating-point types are simply useful in some scenarios.

PR #372, opened 2017-02-06
Number as float

mmatczuk:
This PR introduces distinction between types `number` and `integer` to conform to JSON spec where number is double-precision floating-point.

See https://github.com/cloudwan/gohan/issues/359.

### DIFF
--- a/db/sql/sql.go
+++ b/db/sql/sql.go
@@ -68,7 +68,7 @@ func NewDB() *DB {
 	//TODO(nati) dynamic configuration
 	handlers["string"] = &stringHandler{}
 	handlers["number"] = &numberHandler{}
-	handlers["integer"] = &numberHandler{}
+	handlers["integer"] = &integerHandler{}
 	handlers["object"] = &jsonHandler{}
 	handlers["array"] = &jsonHandler{}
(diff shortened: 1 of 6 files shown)


tools/mysql.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
przemyslaw-dobrowolski-cl: This script does not relate to the original issue. Can you split this PR and add another PR with a information what this script is for?

db/sql/sql.go
```diff
@@ -156,6 +156,35 @@ func (handler *numberHandler) encode(property *schema.Property, data interface{}
...
 func (handler *numberHandler) decode(property *schema.Property, data interface{}) (res interface{}, err error) {
+	if data == nil {
+		return nil, nil
+	}
+	switch t := data.(type) {
+	default:
+		return nil, fmt.Errorf("number: unknown type %T", t)
+
+	case []uint8: // mysql
+		res, _ = strconv.ParseFloat(string(t), 64)
```
przemyslaw-dobrowolski-cl: t is a type of data. Why are we parsing float from type name?

PR #373, opened 2017-02-06
Log buffering imporvements

mmatczuk:
This PR contains two changes, refactoring of logging to avoid direct use of 3rd party logging library and `buflog` rewrite. This fixes race condition on logging configuration.


log/log.go
```diff
@@ -133,3 +155,30 @@ func addLevelsToBackend(config *util.Config, prefix string, backend logging.Leve
...
+		panic("Invalid number of parameters")
+	}
+
+	backendFormatter := logging.NewBackendFormatter(
+		logging.NewLogBackend(w, "", 0),
+		logging.MustStringFormatter(format),
+	)
+	leveledBackendFormatter := logging.AddModuleLevel(backendFormatter)
+
+	for i := 0; i < len(modlevs); i += 2 {
```
przemyslaw-dobrowolski-cl: This looks strange for me. IMO a structure with a parameter and value would be better than assuming specific parameter types and specific order.

log/bufwriter.go
```diff
@@ -0,0 +1,42 @@
...
+// Reset truncates buffer to 0.
+func (w BufWritter) Reset() {
+	buffer().Reset()
+}
+
+var bufferMu sync.Mutex
+
+func buffer() *bytes.Buffer {
+	bufferMu.Lock()
+	defer bufferMu.Unlock()
```
przemyslaw-dobrowolski-cl: I don't get it - we are using a mutex and thread-local-storage in the same function. But gls won't be changed by two different threads so what is mutex used for?

PR #374, opened 2017-02-07
Store watch revisions and use them on start

yudai:
Not to get all entries on ETCD, store last seen revisions to ETCD and
start watching from them on boot.

### DIFF
--- a/server/sync_watch.go
+++ b/server/sync_watch.go
@@ -17,14 +17,15 @@ package server
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/cloudwan/gohan/extension"
 	"github.com/cloudwan/gohan/job"
 
-	gohan_sync "github.com/cloudwan/gohan/sync"
 	l "github.com/cloudwan/gohan/log"
+	gohan_sync "github.com/cloudwan/gohan/sync"
 	"github.com/cloudwan/gohan/util"
 )
 
@@ -33,6 +34,8 @@ const (
 	StateUpdateEventName = "state_update"
 	//MonitoringUpdateEventName used in etcd path
 	MonitoringUpdateEventName = "monitoring_update"
+
+	SyncWatchRevisionPrefix = "/gohan/watch/revision"
 )
 
 //Sync Watch Process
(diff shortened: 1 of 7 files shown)


yudai:
will add tests later.

yudai:
cc / @marcin-ptaszynski

server/sync_watch.go
```diff
@@ -51,13 +54,15 @@ func startSyncWatchProcess(server *Server) {
...
+	responseChans := make(map[string]chan *gohan_sync.Event)
 	stopChan := make(chan bool)
 	for _, path := range watch {
+		responseChans[path] = make(chan *gohan_sync.Event)
 		go func(path string) {
+			responseChan := responseChans[path]
 			defer util.LogFatalPanic(log)
 			for server.running {
-				lockKey := lockPath + "/watch"
+				lockKey := lockPath + "/watch" + path
```
marcin-ptaszynski: nit: [path.Join](https://golang.org/pkg/path/#Join)?

server/sync_watch.go
```diff
@@ -67,32 +72,50 @@ func startSyncWatchProcess(server *Server) {
 				defer func() {
 					server.sync.Unlock(lockKey)
 				}()
-				err = server.sync.Watch(path, responseChan, stopChan, gohan_sync.RevisionCurrent)
+
+				fromRevision := int64(gohan_sync.RevisionCurrent)
+				lastSeen, err := server.sync.Fetch(SyncWatchRevisionPrefix + path)
```
marcin-ptaszynski: ditto

server/sync_watch.go
```diff
@@ -67,32 +72,50 @@ func startSyncWatchProcess(server *Server) {
...
-							env := extensions[event]
-							runExtensionOnSync(server, response, env.Clone())
-							return
+	for _, path := range watch {
+		go func(path string) {
+			defer util.LogFatalPanic(log)
+			responseChan := responseChans[path]
+			for server.running {
+				response := <-responseChan
+				err := server.sync.Update(SyncWatchRevisionPrefix+path, strconv.FormatInt(response.Revision, 10))
```
marcin-ptaszynski: ditto

server/sync_watch.go
yudai: etcdv3's key names are simple strings and etcd doesn't handle key names as slash separated strings. So simple concatenating would be better I think.

yudai:
Hmm, it's pretty hard to test this in the current test suite, because it doesn't use `RunServer()`. Any good idea to run watch process?

nati:
We may make RunServer easy with https://golang.org/pkg/net/http/httptest/ . but I admit this is beyond scope of this one. so I'm ok for having separate patch for future testing.

yudai:
ok, rebased the commits anyaway.

PR #375, opened 2017-02-08
Mock for gohan_exec

morrisson:


morrisson:
@nati -san,

This is very simple patch. Could you please take a look at it?

nati:
@morrisson yes waiting CI get pass

PR #376, opened 2017-02-08
[ESI-10454] Handling nullable attributes in update and create queries

Zimnx:


db/sql/sql_test.go
```diff
@@ -173,6 +177,58 @@ var _ = Describe("Sql", func() {
...
+					err := tx.Update(resource)
+					Expect(err).ToNot(BeNil())
+				})
+
+				It("Setting null to nullable field", func() {
+					resource := extractSingleResource()
+					delete(resource.Data(), "description")
+
+					err := tx.Update(resource)
+					Expect(err).To(BeNil())
```
p-kozlowski: no error, but let's check if the resource was updated as expected

db/sql/sql_test.go
```diff
@@ -173,6 +177,58 @@ var _ = Describe("Sql", func() {
...
+					Expect(err).ToNot(BeNil())
+				})
+
+				It("Setting null to nullable field", func() {
+					resource := extractSingleResource()
+					resource.Data()["id"] = "id_1"
+					delete(resource.Data(), "description")
+
+					err := tx.Create(resource)
+					Expect(err).To(BeNil())
```
p-kozlowski: as above

nati:
LGTM waiting CI

Zimnx:
@nati CI already passed.

nati:
we need running CI again after update..

PR #377, opened 2017-02-09
fix extension builtin sync watch leaks

marcin-ptaszynski:
- fix bug where stopChan wasn't set, which caused goroutine leak
 - fix environment timeout not being respected when extension is
    waiting on gohan_sync_watch. It's general issue for builtins, but
    this commit provides a fix for gohan_sync_watch
 - trigger extension interrupt when server detects client connection
    closed. This prevents CLOSE_WAIT connections stacking up when
    handling Long Poll and memory leaks

### DIFF
--- a/extension/otto/otto.go
+++ b/extension/otto/otto.go
@@ -17,6 +17,7 @@ package otto
 
 import (
 	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/cloudwan/gohan/db"
@@ -171,20 +172,28 @@ func convertNilsToNulls(object interface{}) {
 //HandleEvent handles event
 func (env *Environment) HandleEvent(event string, context map[string]interface{}) (err error) {
 	vm := env.VM
+	var closeNotifier http.CloseNotifier
+	var closeNotify <-chan bool
+	if httpResponse, ok := context["http_response"]; ok {
+		if closeNotifier, ok = httpResponse.(http.CloseNotifier); ok {
+			closeNotify = closeNotifier.CloseNotify()
+		}
+	}
 	context["event_type"] = event
-	var halt = fmt.Errorf("exceed timeout for extension execution")
+	var timeout = fmt.Errorf("exceed timeout for extension execution")
+	var disconnected = fmt.Errorf("client disconnected")
 
 	defer func() {
 		if caught := recover(); caught != nil {
 			if caughtError, ok := caught.(error); ok {
-				if caughtError.Error() == halt.Error() {
-					log.Error(halt.Error())
-					err = halt
-					return
+				switch caughtError {
+				case timeout, disconnected:
+					err = caughtError
+					log.Warning(caughtError.Error())
+				default:
+					panic(caught) // Something else happened, repanic!
 				}
 			}
-
-			panic(caught) // Something else happened, repanic!
 		}
 	}()
 	// take time limit from first passing regex or default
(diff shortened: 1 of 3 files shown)


mmatczuk:
I confirm this fix works, upon disconnect both connection and go routines are closed.

extension/otto/otto.go
```diff
@@ -171,15 +172,24 @@ func convertNilsToNulls(object interface{}) {
...
+	var timeout = fmt.Errorf("exceed timeout for extension execution")
+	var disconnected = fmt.Errorf("Client disconnected")
 
 	defer func() {
 		if caught := recover(); caught != nil {
 			if caughtError, ok := caught.(error); ok {
-				if caughtError.Error() == halt.Error() {
-					log.Error(halt.Error())
-					err = halt
+				if caughtError.Error() == timeout.Error() ||
```
mmatczuk: switch would be more readable

server/middleware/middleware.go
```diff
@@ -45,6 +45,10 @@ func (rh *responseHijacker) Write(b []byte) (int, error) {
 	return rh.ResponseWriter.Write(b)
 }
 
+func (rh *responseHijacker) CloseNotify() <-chan bool {
+	return rh.ResponseWriter.(http.CloseNotifier).CloseNotify()
```
mmatczuk: It's not clear if gohan response writer must implement CloseNotifier interface or not. In https://github.com/cloudwan/gohan/pull/377/files#diff-457a66ed8b4911098a4e6bc1f0be58f7R178 you check if type assertion is safe, here the code would panic if `ResponseWriter` is not `CloseNotifier`. WDYT?

extension/otto/otto.go
```diff
@@ -171,15 +172,24 @@ func convertNilsToNulls(object interface{}) {
...
+	var closeNotify <-chan bool
+	if httpResponse, ok := context["http_response"]; ok {
+		if closeNotifier, ok = httpResponse.(http.CloseNotifier); ok {
+			closeNotify = closeNotifier.CloseNotify()
+		}
+	}
 	context["event_type"] = event
-	var halt = fmt.Errorf("exceed timeout for extension execution")
+	var timeout = fmt.Errorf("exceed timeout for extension execution")
+	var disconnected = fmt.Errorf("Client disconnected")
```
mmatczuk: `"Client` should be `"client`

extension/otto/otto.go
marcin-ptaszynski: I agree, refactor should be in place, but this fix requires minimal changes to enable proper Interrupts. In case of select, I'd be required to repeat myself.

server/middleware/middleware.go
marcin-ptaszynski: I don't agree, this middleware explicitly works with martini and [martini.ResponseWriter implements CloseNotifier](https://github.com/cloudwan/gohan/blob/635491f0ac41b34e387a4578689f67bb27a072b9/vendor/github.com/go-martini/martini/response_writer.go#L83-L83). Extensions part isn't coupled with martini and uses standard http.ResponseWriter and that's why we need to check it first.

Additionally, I don't see how we can implement `else` part here. If we decide to implement interface, we should carry it, otherwise otto part wouldn't be able to rely on that.

extension/otto/otto.go
mmatczuk: >  In case of select, I'd be required to repeat myself.

I was thinking about something like this https://play.golang.org/p/YTxLvxIvFZ

extension/otto/otto.go
```diff
@@ -16,6 +16,7 @@
 package otto
 
 import (
+	"net/http"
```
marcin-ptaszynski: thanks, done

extension/otto/otto.go
marcin-ptaszynski: done

extension/otto/otto.go
marcin-ptaszynski: Thanks for the tip. It's much cleaner now.

marcin-ptaszynski:
@nati , rebased

PR #381, opened 2017-02-14
Fixed memory leak in VM clone implementation

przemyslaw-dobrowolski-cl:
The leak was caused by a long-living reference to interrupt channel in otto VM.

### DIFF
--- a/extension/otto/otto.go
+++ b/extension/otto/otto.go
@@ -262,9 +262,8 @@ func (env *Environment) SetEventTimeLimit(eventRegex string, timeLimit time.Dura
 //Clone makes clone of the environment
 func (env *Environment) Clone() ext.Environment {
 	clone := NewEnvironment(env.Name, env.DataStore, env.Identity, env.Sync)
-	interrupt := clone.VM.Otto.Interrupt
 	clone.VM.Otto = env.VM.Copy()
-	clone.VM.Otto.Interrupt = interrupt
+	clone.VM.Otto.Interrupt = make(chan func(), 1)
 	clone.timeLimit = env.timeLimit
 	clone.timeLimits = env.timeLimits
 	return clone


przemyslaw-dobrowolski-cl:
@p-kozlowski @Zimnx please review

extension/otto/otto.go
```diff
@@ -261,10 +263,11 @@ func (env *Environment) SetEventTimeLimit(eventRegex string, timeLimit time.Dura
 
 //Clone makes clone of the environment
 func (env *Environment) Clone() ext.Environment {
+	cloneMtx.Lock()
```
marcin-ptaszynski: Are you sure this is required? https://github.com/cloudwan/gohan/blob/2e35ef3cc70937abe33604dd6e8c892b60c9897b/vendor/github.com/robertkrimen/otto/clone.go#L17

extension/otto/otto.go
```diff
@@ -261,10 +263,11 @@ func (env *Environment) SetEventTimeLimit(eventRegex string, timeLimit time.Dura
...
+	defer cloneMtx.Unlock()
 	clone := NewEnvironment(env.Name, env.DataStore, env.Identity, env.Sync)
-	interrupt := clone.VM.Otto.Interrupt
 	clone.VM.Otto = env.VM.Copy()
-	clone.VM.Otto.Interrupt = interrupt
+	clone.VM.Otto.Interrupt = make(chan func(), 1)
```
marcin-ptaszynski: Could you provide some background on this bug and how this fixes it?

extension/otto/otto.go
przemyslaw-dobrowolski-cl: indeed, it's not required - removed

extension/otto/otto.go
przemyslaw-dobrowolski-cl: here is a thread which describes the issue:
https://www.bountysource.com/issues/2289358-memleak-caused-by-otto-clones

PR #382, opened 2017-02-15
ESI-10965 cron fixes

marcin-ptaszynski:
- use environment Clone instead of reusing it between jobs
 - recover from panic with log.Error message instead of crashing
   whole server.


marcin-ptaszynski:
Closing in favor of https://github.com/cloudwan/gohan/pull/383 , which includes additional fixes.

PR #384, opened 2017-02-15
gofmt -s -w

mmatczuk:


vendor/github.com/flosch/pongo2/lexer.go
```diff
@@ -70,7 +70,7 @@ type lexer struct {
 func (t *Token) String() string {
 	val := t.Val
 	if len(val) > 1000 {
-		val = fmt.Sprintf("%s...%s", val[:10], val[len(val)-5:len(val)])
```
przemyslaw-dobrowolski-cl: Gofmt for gohan source is ok, but why do we enforce gofmt for vendor libraries which we do not own?

mmatczuk: @przemyslaw-dobrowolski-cl the short answer is for convenience, you can format code in fire and forget mode. Otherwise you have to always revert `vendor` or do some sort of `grep -v vendor`. The long answer is that I'd recommend you go and vote fort https://github.com/golang/go/issues/19090 then the issue will be solved.

nati:
I don't think we need change vendor code.
We can use govendor code to exclude it from fmt command.
also our make format command do same thing

mmatczuk:
We sure can use `govendor`. I fixed `make format` as it was doing `gofmt ./...` furthermore I have split bindata generation from format and adjusted ci script.

nati:
awesome. Thx

PR #385, opened 2017-02-15
added database/auto_migrate config flag

mmatczuk:
This PR adds `auto_migrate` flag (default true) that prevents from alter table on start, sample config snippet

    database:
        type: "mysql"
        auto_migrate: false
        connection: "gohan:gohan@/gohan_test"

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -136,7 +136,7 @@ func (server *Server) initDB() error {
 
 func (server *Server) connectDB() error {
 	config := util.GetConfig()
-	dbType, dbConnection, _, _ := server.getDatabaseConfig()
+	dbType, dbConnection, _, _, _ := server.getDatabaseConfig()
 	maxConn := config.GetInt("database/max_open_conn", db.DefaultMaxOpenConn)
 	dbConn, err := db.ConnectDB(dbType, dbConnection, maxConn)
 	if server.sync == nil {
@@ -147,7 +147,7 @@ func (server *Server) connectDB() error {
 	return err
 }
 
-func (server *Server) getDatabaseConfig() (string, string, bool, bool) {
+func (server *Server) getDatabaseConfig() (string, string, bool, bool, bool) {
 	config := util.GetConfig()
 	databaseType := config.GetString("database/type", "sqlite3")
 	if databaseType == "json" || databaseType == "yaml" {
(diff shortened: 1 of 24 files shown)


server/server.go
```diff
@@ -159,7 +159,8 @@ func (server *Server) getDatabaseConfig() (string, string, bool, bool) {
 	}
 	databaseDropOnCreate := config.GetBool("database/drop_on_create", false)
 	databaseCascade := config.GetBool("database/cascade_delete", false)
-	return databaseType, databaseConnection, databaseDropOnCreate, databaseCascade
+	databaseAutoMigrate := config.GetBool("database/auto_migrate", true)
```
marcin-ptaszynski: I think it would be also nice to have auto_create flag, which would disable table creation and initial data insertion. This way, we can be sure, we don't mixup external migration with automatically created db.

mmatczuk: Sounds good

mmatczuk: @marcin-ptaszynski actually there is `no_init` flag already WDYT?

przemyslaw-dobrowolski-cl: There were some incidents with corrupted databases after auto-migration. I suggest setting this flag to false by default.

marcin-ptaszynski: @mmatczuk , I wasn't aware of it. It's fine then. Thanks.

marcin-ptaszynski: @mmatczuk , It seems no_init flag also disables migration, which is exactly what we need. Perhaps this change is not required, then. @przemyslaw-dobrowolski-cl ?

przemyslaw-dobrowolski-cl: There is such a flag, indeed. So IMO, this commit should only switch its default value to 'true' which means that no migrations are performed by default

mmatczuk: 1. `no_init` it lacks a sanity check, you will notice that db is in wrong version by crashing requests
1. `auto_migrate` default `true` is set to preserve current behavior, @nati do you have an opinion on that front?

przemyslaw-dobrowolski-cl: There is a WIP including the mentioned sanity checks:
https://github.com/przemyslaw-dobrowolski-cl/gohan/tree/migration
It contains a part of migration code that checks the DB during startup and if there are any changes in DB, the startup is aborted. Maybe we can wait for it to be merged first and then we can continue with the change of the default value of no_init flag?

mmatczuk: Sure.

nati:
Could you update doc?

mmatczuk:
I have also marked yaml snippets as `yaml` so that they are properly rendered.

nati:
Could you rebase this branch?
also it is better to add description of this flag in this section. https://github.com/cloudwan/gohan/blob/master/docs/configuration.md#database

mmatczuk:
Sure, thanks for pointing this, I was following `no_init` flag, will document that too.

nati:
ya we should add doc for no_init too

PR #388, opened 2017-02-20
cli: backup command

mmatczuk:
This PR introduces backup capabilities to gohan. I'd thankful for advice how do we want this to be integrated with migrations. Do we want another flag like `dump_on_migrate`? IMHO dumping a database can be quite a surprising side effect of running a server. On the other hand the `migrate` command just generates SQL files.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -18,28 +18,30 @@ package cli
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


nati:
We should use a proper tool for each DB. (eg. mysqldump for mysql), so I believe we don't 
need to have this.

mmatczuk:
Sure I got your point, please close the PR then.

cc: @przemyslaw-dobrowolski-cl

nati:
got it. thank you for your understandings.

PR #389, opened 2017-02-20
ESI-10968 Allow executing SELECT ... FOR UPDATE type queries

p-kozlowski:


extension/otto/gohan_db_test.go
```diff
@@ -136,230 +138,250 @@ var _ = Describe("GohanDb", func() {
...
-					),
-				)
-			})
+					  });`, function),
+						"path": ".*",
+					})
+					Expect(err).ToNot(HaveOccurred())
+					env := newEnvironmentWithExtension(extension, testDB)
+
+					var pagenator *pagination.Paginator
```
Zimnx: typo

extension/otto/gohan_db_test.go
```diff
@@ -136,230 +138,250 @@ var _ = Describe("GohanDb", func() {
...
```
Zimnx: typo

extension/otto/gohan_db_test.go
```diff
@@ -136,230 +138,250 @@ var _ = Describe("GohanDb", func() {
...
-					),
-				)
-			})
+					  });`, function),
+						"path": ".*",
+					})
+					Expect(err).ToNot(HaveOccurred())
+					env := newEnvironmentWithExtension(extension, testDB)
+
+					pagenator := &pagination.Paginator{
```
Zimnx: typo

extension/otto/gohan_db_test.go
```diff
@@ -136,230 +138,250 @@ var _ = Describe("GohanDb", func() {
...
```
Zimnx: typo

extension/otto/gohan_db_test.go
```diff
@@ -136,230 +138,250 @@ var _ = Describe("GohanDb", func() {
...
```
Zimnx: typo

db/transaction/mocks/Transaction.go
```diff
@@ -107,6 +107,29 @@ func (_m *Transaction) Fetch(_a0 *schema.Schema, _a1 transaction.Filter) (*schem
 	return r0, r1
 }
 
+// LockFetch mock
+func (_m *Transaction) LockFetch(_a0 *schema.Schema, _a1 transaction.Filter) (*schema.Resource, error) {
```
Zimnx: Is it generated code? If not, variable names are really bad.

extension/otto/gohan_db.go
```diff
@@ -147,6 +200,39 @@ func init() {
...
+					ThrowOttoException(&call, err.Error())
+				}
+
+				resp, err := GohanDbLockFetch(transaction, schemaID, ID, tenantID)
+				if err != nil {
+					ThrowOttoException(&call, err.Error())
+				}
+
+				if resp == nil {
+					otto.NullValue()
```
Zimnx: resp = otto.NullValue
Add test for it

db/transaction/mocks/Transaction.go
p-kozlowski: sort of auto generated (have a look at whole file). BTW @mmatczuk would probably not agree about the naming ;)

extension/otto/gohan_db_test.go
p-kozlowski: fixed (same elsewhere)

extension/otto/gohan_db.go
p-kozlowski: Nice catch.
In fact it's common code with gohan_db_fetch, and the contract of underlaying transaction. Fetch guarantees that we'd either get a result or an error. So I'd rather add a panic()

db/sql/sql.go
```diff
@@ -705,6 +703,30 @@ func (tx *Transaction) List(s *schema.Schema, filter transaction.Filter, pg *pag
...
+	sql, args, err := buildSelect(s, filter, pg)
+	if err != nil {
+		return nil, 0, err
+	}
+
+	return executeSelect(s, filter, sql, args, tx)
+}
+
+//Lock resources in the db
+func (tx *Transaction) LockList(s *schema.Schema, filter transaction.Filter, pg *pagination.Paginator) (list []*schema.Resource, total uint64, err error) {
```
marcin-ptaszynski: I think we shouldn't build with related properties. By default, we shouldn't lock more than specified. Consider optional parameter to pass related resources to lock.

p-kozlowski: I agree with adding special handling of related resources.

On the other hand, it's worth to mention that when using pagination we may lock more than expected ([gap locks](http://stackoverflow.com/a/5704927)). Example:
session1> set session transaction isolation level serializable;
session1> start transaction;
session1> select id, name from some_table order by id limit 1 offset 1 for update;
session2> set session transaction isolation level serializable;
session2> start transaction;
session2> select id, name from some_table order by id limit 1 for update;
session2> ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction

What's more, Fetch() uses List() and [filters results itself](https://github.com/cloudwan/gohan/blob/master/db/sql/sql.go#L773), so that's another situation of locking more than needed.

As a result, I don't think we could completely eliminate excessive locking.

nati:
LGTM but could you rebase commits?

p-kozlowski:
Rebased.

PR #392, opened 2017-02-22
Add resource permission

mmatczuk:
This PR modifies `GetSchema` to extend resource with `permission` field listing resource effective permissions.

Sample output for UI

```json
{
  "schemas": [
    {
      ...
      "schema": {
        "permission": [
          "read"
        ],
        "properties": {
           ...
```

PR #394, opened 2017-02-22
Apply policy

mmatczuk:
This PR is a followup to https://github.com/cloudwan/gohan/pull/392. It contains a change in `GetSchema` that filters property `permission` so that the values reflect the imposed policy.


mmatczuk:
please review @marcin-ptaszynski

PR #395, opened 2017-02-23
ESI-10735 Property specifier on Gohan List API

mmatczuk:
This PR adds `fields` flag to gohan client, if set, table output would be trimmed to only selected fields. Server is adjusted to only query database for relevant columns.

cc: @przemyslaw-dobrowolski-cl


nati:
@mmatczuk could you rebase?

mmatczuk:
@nati it's rebased

PR #399, opened 2017-02-27
Added option for multiple schema files to init-db

michalzurawski:


nati:
Could you take config-file instead of list of schemas?
i think it is more useful.

michalzurawski:
Done in separete [PR](https://github.com/cloudwan/gohan/pull/400).
Though this still may be helpful as it is somehow more general. Please close this one if you think it is not needed.

nati:
Thanks. Yes. This PR is still useful.

PR #404, opened 2017-03-03
[Blocker] [ESI-11374] Revert "apply policy to schema property"

marcin-ptaszynski:
This reverts commit 86899c193c0eed9d3a63e7ce0421a9a73d6ac93d.

This commit introduced a bug when used in multi-tenant and multi-role environment. clearPermissions overrides shared data structure, which causes admin permissions to be removed.

cc: @mmatczuk , @Zimnx


mmatczuk:
@marcin-ptaszynski you are right, we should not write to `s.JSONSchema`, at this point to fix this I'd add a [deep copy](https://github.com/mohae/deepcopy/blob/master/deepcopy.go) of `schemaProperties` at https://github.com/cloudwan/gohan/blob/master/server/editor.go#L44 WDYT?

przemyslaw-dobrowolski-cl:
reupload with fixes applied:
https://github.com/cloudwan/gohan/pull/450

PR #411, opened 2017-03-09
fix deep equal in js test mocks

marcin-ptaszynski:
cc: @p-kozlowski , @Zimnx , @mmatczuk


marcin-ptaszynski:
Hi @nati , could you please have a look?

PR #413, opened 2017-03-14
Workaround for a goose bug

mmatczuk:
This is a workaround for a bug in goose, resulting in 

```2017/03/14 10:24:00 sql: Scan error on column index 0: unsupported Scan, storing driver.Value type []uint8 into type *time.Time```

on start.

### DIFF
--- a/db/migration/migration.go
+++ b/db/migration/migration.go
@@ -62,18 +62,23 @@ func Init() error {
 	dbType, dbConnection, migrationsPath := readGooseConfig()
 
 	if err := goose.SetDialect(dbType); err != nil {
-		fmt.Printf("error: failed to set goose dialect: %s\n", err.Error())
-		return err
+		return fmt.Errorf("migration: failed to set goose dialect: %s", err)
 	}
 
 	db, err := sql.Open(dbType, dbConnection)
 
 	if err != nil {
-		fmt.Printf("error: failed to open db: %s\n", err.Error())
-		return err
+		return fmt.Errorf("migration: failed to open db: %s", err)
 	}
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by przemyslaw-dobrowolski-cl** [comment]
@Zimnx @p-kozlowski @marcin-ptaszynski please review

File: db/migration/migration.go [hunk 1]
```diff
@@ -62,18 +62,28 @@ func Init() error {
...
 
-	return goose.Status(db, migrationsPath)
+	v, err := goose.EnsureDBVersion(db)
+	if err != nil {
+		return fmt.Errorf("migration: failed to ensure db version: %s", err)
+	}
+
+	logger.Info("migration path: %q, version: %d", migrationsPath, v)
+
+	// blocked by https://github.com/pressly/goose/pull/19
```
marcin-ptaszynski: Do we need to wait for this PR to be merged?

nati:
let's wait if they merge your PR code or not. 
We may manage our folk too. (this is anyway painful though..)

mmatczuk:
@nati the PR in goose is pending, please merge this one as it is, the difference is that while starting gohan only prints current version instead of a fancy table with the migrations and statuses.

mmatczuk:
It turns out to be a configuration issue to some extent, MySQL driver exposes https://github.com/go-sql-driver/mysql#parsetime flag that needs to be set to make it work.

mmatczuk:
I have dropped goose printing the table

* it's being printed directly to stdout outside of logger so if you use file logger the information is "missing"
* database version information is just enough for operations
* no global driver setting needed

PR #414, opened 2017-03-15
Added tags to openapi documentation

michalzurawski:
This will add tags based on resource_group field.

Have a question about this commit:
As you see now it's turn on by default - should I add some flag to enable it only when the flag is given?
And probably split-by-resource will now be redundant as this option allows to create categories in a single json file.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -83,6 +83,8 @@ GLOBAL OPTIONS:
     OPTIONS:
         --config-file "gohan.yaml"				Server config File
         --template, -t "embed://etc/templates/openapi.tmpl"	Template File
+        --split-by-resource-group ""                           Group by resource
+        --policy "admin"                                       Show only schema with chosen policy
 ```
 
 ## MarkDown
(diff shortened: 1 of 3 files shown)


nati:
resource_group tag should be documented, and you should check if a schema has this or not.
Current code may add a tag with "".

michalzurawski:
Thanks, @nati for that notes.
Have an additional question: how to "compile" this templates for Openapi? It seems that even if completely new Gohan is installed it still uses the old template file.

nati:
Have you done  this?

``` shell
make gen
```

We are using [go-bindata](https://github.com/jteeuwen/go-bindata) and converting template file for go binary. 
You can also manually specify template by command.

michalzurawski:
Thanks a lot `make gen` is exactly what I was looking for.
I knew about `--template` option but it was rather inconvenient.

michalzurawski:
@Zimnx @przemyslaw-dobrowolski-cl @p-kozlowski @marcin-ptaszynski @mmatczuk Please review.

przemyslaw-dobrowolski-cl:

thayashi:
@nati pls take a look this PR

nati:
got it

thayashi:
@nati thanks!

PR #415, opened 2017-03-16
Add builtin gohan_sync_delete for JS extension

morrisson:
Usecase: With configuration watch/keys list, gohan can run some events
  reported from southbound resources. However, this is not a sync_key of
  gohan resources, so gohan does not have a way to delete watching etcd
  pathes. These garbages will occur worse performance, and if gohan
  restarted for some reason, a lot of unnecessary events might happen
  due to remaining sync pathes. Of course, cleanup of sync pathes can be
  done from outside of gohan, but this new builtin lets gohan manage
  sync pathes efficiently based on gohan's internal status.

  This builtin can delete multiple pathes based on prefix if etcd v3 is
  used for sync. This should be also convenient for this kind of cleanup
  feature.

### DIFF
--- a/docs/js_extension.md
+++ b/docs/js_extension.md
@@ -295,6 +295,15 @@ Check if dir
 
 Fetch a given path from Sync   
 
+- gohan_sync_delete(path, prefix)
+
+Delete a given path from Sync. If prefix is given as ``true``, all paths which starts
+with ``path`` are deleted.
+path : sync path to be deleted
+prefix (optional) : boolean - whether to delete paths which starts with ``path``.
+This option is applied if gohan uses etcd v3 for sync. If gohan uses etcd v2, this option
+is translated as ``--recursive``
+
 - gohan_sync_watch(path, timeout, revision)
 
 Watch a given path in Sync starting from a given revision. This call is blocking no longer
--- a/extension/otto/gohan_sync.go
+++ b/extension/otto/gohan_sync.go
@@ -80,7 +80,7 @@ func init() {
 
 				select {
 				case interrupt := <-call.Otto.Interrupt:
-					log.Debug("Received otto interrupt in gohan_http")
+					log.Debug("Received otto interrupt in gohan_sync_fetch")
 					interrupt()
 				case <-done:
 				}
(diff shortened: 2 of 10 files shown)


sync/etcd/etcd.go
```diff
@@ -61,7 +61,7 @@ func (s *Sync) Update(key, jsonString string) error {
 }
 
 //Delete sync update sync
-func (s *Sync) Delete(key string) error {
+func (s *Sync) Delete(key string, prefix bool) error {
 	s.etcdClient.Delete(key, false)
```
p-kozlowski: prefix is ignored. I think it's safe to pass it to etcdClient.Delete(). if that's not the case, I'd add
```
if prefix {
    panic("Prefix not supported in ETCD v2")
}
```

morrisson: etcd v2 does not support 'prefix'. It supports 'recursive' because in Etcd v2 has a concept of 'directory'. 'recursive' is similar to 'prefix', but strictly speaking, it is different.

Therefore, actually I considered if this prefix argument should pass it to etcdClient.Delete() or not, then I decided not to pass. However, I do not have strong opinion to it.

Ok, I will pass this prefix argument to etcdClient.Delete(), and produce some info or warning log like 'Prefix option is translated as Recursive in ETCD v2'.

docs/js_extension.md
```diff
@@ -295,6 +295,15 @@ Check if dir
 
 Fetch a given path from Sync   
 
+- gohan_sync_delete(path, prefix)
+
+Delete a given path from Sync. If prefix is given as ``true``, all pathes which starts
+with ``path`` are deleted.
+path : sync path to be deleted
+prefix (optional) : boolean - whether to delete pathes which starts with ``path``.
```
marcin-ptaszynski: typo: pathes -> paths

extension/otto/otto_test.go
```diff
@@ -1859,6 +1859,61 @@ var _ = Describe("Otto extension manager", func() {
...
+
+	Describe("Using gohan_sync_delete builtin", func() {
+		It("Should delete sync", func() {
+			extension, err := schema.NewExtension(map[string]interface{}{
+				"id": "test_extension",
+				"code": `
+					gohan_register_handler(
+						"test_event",
+					 	function(context) {
+							gohan_sync_delete("/gohan_sync_delete_test");
```
yudai: no `, false` here?

morrisson: 2nd parameter is optional. Default is false. Here, I tested single argument case.

extension/otto/gohan_sync.go
```diff
@@ -96,6 +96,46 @@ func init() {
...
+				done := make(chan struct{})
+				go func() {
+					err = env.Sync.Delete(path, prefix)
+					close(done)
+				}()
+
+				select {
+				case interrupt := <-call.Otto.Interrupt:
+					log.Debug("Received otto interrupt in gohan_sync_delete")
+					interrupt()
```
yudai: what happens in this func here?

yudai: because, `if err != nil` below is not true?

marcin-ptaszynski: @yudai , [passed interrupts panic](https://github.com/cloudwan/gohan/blob/master/extension/otto/otto.go#L216)

extension/otto/otto_test.go
yudai: ok thanks

PR #419, opened 2017-03-20
Add support for GO database migrations via golang plugins system

przemyslaw-dobrowolski-cl:
This PR requires an upgrade to stable GO 1.8


przemyslaw-dobrowolski-cl:
@Zimnx @marcin-ptaszynski please review and test

db/migration/migration.go
```diff
@@ -24,10 +24,193 @@ import (
...
+			"-o", pluginName,
+		}, runtimeSources...)...).CombinedOutput()
+
+	if commandErr != nil {
+		logger.Error("migration: failed to build runtime migrations plugin: %s; error: %s",
+			commandErr.Error(), commandOutput)
+		return commandErr
+	}
+
+	_, openErr := plugin.Open(pluginName)
```
mmatczuk: @przemyslaw-dobrowolski-cl could we unload the plugins when migration is over? The flow could be:

* compile if not compiled
* load .so
* find all implementations of `Migrator` interface in the loaded code
* execute
* unload

WDYT?

przemyslaw-dobrowolski-cl: It is difficult to achieve since each go migration register its global methods in goose. There is no deregister handler, so goose would have to be patched. No need for such effort for so little memory savings.

przemyslaw-dobrowolski-cl:
rebased

przemyslaw-dobrowolski-cl:

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.7
+FROM golang:1.8
```
marcin-ptaszynski: @nati , are you ok with removing go1.7 compatibility?

@przemyslaw-dobrowolski-cl , please make sure, we upgrade our build and CI scripts before merging this change. AFAIR, we use go1.7 everywhere.

marcin-ptaszynski:
@przemyslaw-dobrowolski-cl , such feature definitely requires integration tests.

nati:
I don't understand why we need this in Gohan..
Why not just use goose?

przemyslaw-dobrowolski-cl:
@nati We'd like to have migrations in go as well. The problem is that in pressly fork of goose, goose migrations have to be compiled into the binary. Of course we cannot do that. One idea is to use a plugins to compile go migrations (go build) and attach them (dlopen) to a running process. Another solution is to mimic original goose approach and execute compilation and run externally (go run). Either way, we need a working go compiler in PATH. The reason that we chose pressly fork is that it is more up-to-date (fixes included), does not require configuration to be stored in files (original goose requires that) and it has no vendor deps.
More intrusive alternatives are:
- drop built-in migrations support and use external goose binary
- use other script to write advanced migrations (when sql is insufficient), maybe js?
- use an experimental golang interpreter for golang.

Which way do you think we should go?

nati:
I understand 

- a value for wrapping goose CLI because we can reuse same configuration file even if Gohan scope - let DB migration framework support go-plugin framework so that we don't require Go toolchain in running env.

However, we shouldn't do it in Gohan code tree.
We should propose this feature for Goose community ( or managing our Goose fork with this function ). and use it from Gohan framework

przemyslaw-dobrowolski-cl:
We redesigned the support for go migrations. The new version does not require a working compiler to be present. Instead of having migrations in raw source code (which has to be compiled), we will provide precompiled migrations is go. Each migration can have its own *.so library or there can be only one migrations library for all migrations. This way we decouple process of building migration libraries from loading and running them in runtime.
This versions requires some extensions to pressly/goose - there are two PRs:
https://github.com/pressly/goose/pull/22
https://github.com/pressly/goose/pull/23
For this version to compile, these changes were manually introduced to vendor version of pressly/goose. Once PRs are merged they can be reimported to vendor.
Note that this PR still requires golang 1.8 in order to load plugins.

przemyslaw-dobrowolski-cl:
@nati I'd be glad to know your thoughts about the redesigned approach

etc/gohan.yaml
```diff
@@ -14,6 +14,8 @@ database:
     max_open_conn: 100
     # please set no_init true in the production env, so that gohan don't initialize table
     # no_init: true
+    # migrations path contains SQL migrations (*.sql)
+    # and precompiled GO migrations (*.so)
```
marcin-ptaszynski: there should be an instruction on how to make those.

przemyslaw-dobrowolski-cl: done, migrations.md was added

przemyslaw-dobrowolski-cl:
rebased

przemyslaw-dobrowolski-cl:
@marcin-ptaszynski please review

przemyslaw-dobrowolski-cl:
rebased

db/migration/migration.go
```diff
@@ -19,29 +19,87 @@ import (
 	"fmt"
```
nati: As I mentioned in previous comment, this logic should be present in Goose side.

db/migration/template.go
```diff
@@ -0,0 +1,83 @@
...
+`-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+
+
+-- +goose Down
+-- SQL section 'Down' is executed when this migration is rolled back
+
+`))
+
+func createTemplate(w io.Writer, baseName string, migrationType string, version int64) error {
```
nati: tempalte part is OK. because we can only do this in gohan.

przemyslaw-dobrowolski-cl:
I managed to split this PR according to the suggestions. The related PR to goose is:
https://github.com/pressly/goose/pull/26
In the present version of PR for Gohan, the only change is a call LoadMigrationPlugins from goose.

nati:
OK thanks. how about manage our goose fork until this problem get merged?

pdobrowo:
I got an information that pressly/goose maintainer will not be available for the next two weeks. Initially, the PR looks ok for him but it needs some minor improvements. For gohan PR to work, I govendored my changes to goose so there is no need to maintain a fork. If you rather prefer to have a fork, my github repository may be used as a fork - it contains the needed changes. As soon as pressly/goose changes are merged I can update vendor to final changes from pressly/goose. Which option do you prefer?

przemyslaw-dobrowolski-cl:

przemyslaw-dobrowolski-cl:
I switched pressly/goose to cloudwan/goose (master).

nati:
ok thanks. 
I've also put some review comment

db/migration/migration.go
przemyslaw-dobrowolski-cl: whole logic has been moved to cloudwan/goose

przemyslaw-dobrowolski-cl:
@nati I believe all comments have been already addressed. Since there no updates to this PR I'd like to ask about the current status of this PR.

nati:
could you take a look migration.go?

PR #420, opened 2017-03-21
[ESI-10997] emit 'exit' event on test exit. gohan_load_hook and gohan_get_env hook...

Zimnx:
This PR contain few changes:

- Basic functionality of `crypto`, `fs`, `vm` modules from Node.js
- Update of ddliu/motto to current master
- emiting 'exit' event on extension test end - this can be used for gathering some runtime test results such as coverage, profiling
- gohan_load_hook builtin - ability to register hook for code loading. This can be for instrumenting code with coverage flags, profiler flags. It's implemented as "don't pay if you don't use it' manner. 
- gohan_get_env builtin - retrieving environmental variable from extension code
- removed "file://" prefix from paths passed to vm.Load()

We use these changes for code coverage together with `istanbul` library.

### DIFF
--- a/extension/otto/otto_test.go
+++ b/extension/otto/otto_test.go
@@ -28,6 +28,8 @@ import (
 	"github.com/onsi/gomega/ghttp"
 	ottopkg "github.com/xyproto/otto"
 
+	"os"
+
 	"github.com/cloudwan/gohan/db/transaction"
 	"github.com/cloudwan/gohan/extension"
 	"github.com/cloudwan/gohan/extension/otto"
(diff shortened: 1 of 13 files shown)


extension/otto/otto_test.go
```diff
@@ -1779,6 +1781,100 @@ var _ = Describe("Otto extension manager", func() {
...
+			})
+			Expect(err).ToNot(HaveOccurred())
+			extensions := []*schema.Extension{extension}
+			env := newEnvironment()
+			context := map[string]interface{}{}
+			Expect(env.LoadExtensionsForPath(extensions, timeLimit, timeLimits, "test_path")).To(Succeed())
+			Expect(env.HandleEvent("test_event", context)).To(Succeed())
+			Expect(context["resp"]).To(Equal("123"))
+
+			os.Unsetenv("GOHAN_TEST_GET_ENV")
```
p-kozlowski: it won't be called if the test fails, potentially harming other tests. move to Before/AfterEach

extension/otto/otto_test.go
```diff
@@ -1779,6 +1781,100 @@ var _ = Describe("Otto extension manager", func() {
...
+		})
+
+		It("Should return default value environmental variable is not set", func() {
+			extension, err := schema.NewExtension(map[string]interface{}{
```
p-kozlowski: if you want to be sure the env variable is not set (cf. comment above), call Unsetenv (in a fixture)

Zimnx:
rebased

PR #422, opened 2017-03-24
ESI-11829 Allow locking resorces in pre update/create

p-kozlowski:
In non-trivial cases of updating/deleting a resource one may want to lock the resource at the very beginning of handling. It's an opt-in mechanism, and it's backward compatible - by default, all resources are still fetched without DB locking.

### DIFF
--- a/tests/test_schema_metadata.yaml
+++ b/tests/test_schema_metadata.yaml
@@ -64,3 +64,22 @@ schemas:
     type: object
   singular: metadata_id
   title: MetadataId
+- description: MetadataPolicy
+  id: metadata_policy
+  plural: metadatas_policies
+  metadata:
+    locking_policy:
+      update: lock_related
+      delete: skip_related
+  schema:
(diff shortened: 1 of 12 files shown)


p-kozlowski:
@marcin-ptaszynski, @Zimnx, @przemyslaw-dobrowolski-cl please review

tests/test_schema_metadata.yaml
```diff
@@ -63,4 +63,23 @@ schemas:
...
     type: object
   singular: metadata_id
-  title: MetadataId
\ No newline at end of file
+  title: MetadataId
+- description: MetadataPolicy
+  id: metadata_policy
+  plural: metadatas_policies
+  metadata:
+    locking_policy:
```
marcin-ptaszynski: please update docs

p-kozlowski: done

PR #423, opened 2017-03-27
ESI-11891 Reject gohan startup when there is a migration pending

mmatczuk:


mmatczuk:
Still waiting for vendored code be merged.

mmatczuk:
@przemyslaw-dobrowolski-cl @Zimnx @marcin-ptaszynski please review.

marcin-ptaszynski:
@mmatczuk , this PR makes impossible for users to opt-out migration. Migrations should be opt-in feature, and not mandatory. Developers should be able to run dev server on sqlite3 backend enabled auto-creation of db schema.

mmatczuk:
Vendor code is merged now.

server/server.go
```diff
@@ -136,7 +136,7 @@ func (server *Server) initDB() error {
 }
 
 func (server *Server) connectDB() error {
-	if err := migration.Init(); err != nil {
+	if err := migration.EnsureVersion(); err != nil {
```
przemyslaw-dobrowolski-cl: it should check no_init flag and reject running only when no_init = true and there are migrations pending

mmatczuk:
Work moved to https://github.com/cloudwan/gohan/pull/429

PR #426, opened 2017-03-29
Made SkipRelatedResources the default locking policy

michalzurawski:
ESI-11495


Fenthick:
Approved - Daniel (please add me as a reviewer in future)

PR #427, opened 2017-03-29
Use sequential numbering for migrations

przemyslaw-dobrowolski-cl:
This commit switches migration file numbering from dates to sequential number so that developers will see a conflict when two migrations are made the same for the same master starting point.


nati:
Is this backward compatible change?
we should let user choose this option.

przemyslaw-dobrowolski-cl:
In this commit it is enforced to use sequential numbering for templates but I'll change it for backward comparability. Now it will be advised to use 'create-next' instead of 'create' to create migration templates

przemyslaw-dobrowolski-cl:

PR #429, opened 2017-03-30
ESI-11891 Reject gohan startup when there is a migration pending

przemyslaw-dobrowolski-cl:
This commit prevents from running gohan when there is a pending migration
(which means that gohan and its extensions are in a specific version)
and the database used in gohan is in a previous version.
This check only works when no_init is set to true (ie. in production
environments).
In that case, operator must manually execute migration to the new
version making safety db dumps and other safety checks.

### DIFF
--- a/db/migration/migration.go
+++ b/db/migration/migration.go
@@ -48,18 +48,19 @@ func LoadConfig(configFile string) (err error) {
 	return
 }
 
-func readGooseConfig() (dbType, dbConnection, migrationsPath string) {
+func readGooseConfig() (dbType, dbConnection, migrationsPath string, noInit bool) {
 	config := util.GetConfig()
 	dbType = config.GetString("database/type", "sqlite3")
 	dbConnection = config.GetString("database/connection", "")
 	migrationsPath = config.GetString("database/migrations", "etc/db/migrations")
+	noInit = config.GetBool("database/no_init", false)
 	return
 }
 
 func Init() error {
 	logger.Info("migration: init")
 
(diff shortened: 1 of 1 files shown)


przemyslaw-dobrowolski-cl:

db/migration/migration.go
```diff
@@ -78,30 +79,67 @@ func Init() error {
...
+		return fmt.Errorf("migration: failed to load migration plugins: %s", err)
+	}
+
+	if noInit {
+		// pending migrations are not allowed if no-init is enabled (no_init=true)
+		last, err := goose.LastMigration(migrationsPath)
+
+		if err != nil {
+			if err != goose.ErrNoNextVersion {
+				return fmt.Errorf("migration: %s", err)
```
p-kozlowski: correct me if I'm wrong, but what happens if there are no migrations and noInit is true? I guess we'll fail to start. I wouldn't mind having a UT for it...

przemyslaw-dobrowolski-cl: You're right - thanks. It crashed when there were no migrations at all. Fixed.
Regarding the tests - I've no idea for a simple test for that. Any suggestions?

p-kozlowski: 1. put some (or no) files in the migration directory
2. mock goose.GetDBVersion and goose.Status(db, migrationsPath)
3. run migration.Init and make asserts on the result

marcin-ptaszynski:
It needs rebasing.

przemyslaw-dobrowolski-cl:
rebased

PR #430, opened 2017-03-30
Post migration event

Zimnx:
An Opt-In mechanism of emiting post-migration event in extensions of modified schemas.

Go migrations can use `migration.markSchemaAsModified(schemaId)` callback within Gohan which marks schema as modified. 
After successfull migration, `post-migration` event is emitted in modified schemas to allow post migration processing.


Zimnx:
https://github.com/cloudwan/gohan/pull/419
need to be merged before this.

I'll rebase after merge

cli/cli.go
```diff
@@ -357,14 +364,175 @@ func getMigrateSubcommand(subcmd, usage string) cli.Command {
...
+
+			log.Debug("Modified schemas: %s", modifiedSchemas)
+
+			schemaFiles := config.GetStringList("schemas", nil)
+			if schemaFiles == nil {
+				log.Fatal("No schema specified in configuraion")
+			}
+
+			manager := schema.GetManager()
+			if err := manager.LoadSchemasFromFiles(schemaFiles...); err != nil {
```
p-kozlowski: is it possible that a post-migration event handler in one schema triggers an event (eg. update) on a different schema? if so, I think we have to load all schemas here.

cli/cli.go
```diff
@@ -357,14 +364,175 @@ func getMigrateSubcommand(subcmd, usage string) cli.Command {
...
+		}
+	default:
+		err = fmt.Errorf("invalid sync type: %s", syncType)
+		return
+	}
+	return
+}
+
+func createDbFromConfig(config *util.Config) (db.DB, error) {
+	dbType := config.GetString("database/type", "sqlite3")
```
p-kozlowski: code duplication with ConnectDB and getDatabaseConfig

cli/cli.go
```diff
@@ -357,14 +364,175 @@ func getMigrateSubcommand(subcmd, usage string) cli.Command {
 	}
 }
 
+func createSyncFromConfig(config *util.Config) (sync sync.Sync, err error) {
+	syncType := config.GetString("sync", "etcd")
+	switch syncType {
+	case "etcd":
```
p-kozlowski: you could reuse this function in server.NewServer

cli/cli.go
```diff
@@ -357,14 +364,175 @@ func getMigrateSubcommand(subcmd, usage string) cli.Command {
...
+	dbConn := sql.NewDB()
+	err := dbConn.Connect(dbType, dbConnection, db.DefaultMaxOpenConn)
+	if err != nil {
+		return nil, err
+	}
+	return dbConn, nil
+}
+
+func createIdentityFromConfig(config *util.Config) (middleware.IdentityService, error) {
+	if config.GetBool("keystone/use_keystone", false) {
```
p-kozlowski: again, this could be reused in server.NewServer

cli/cli.go
Zimnx: All schemas defined in config file are loaded. Editing other resources is possible.

p-kozlowski: I must be blind, I thought you're loading only modifiedSchemas. Sorry.

PR #432, opened 2017-04-04
ESI-10464 Gohan returns all resources related to the requested resource in show API

mmatczuk:
This PR introduces details flag used in list API to avoid joins.


db/pagination/pagination.go
```diff
@@ -35,15 +35,16 @@ const (
...
-	Key    string
-	Order  string
-	Limit  uint64
-	Offset uint64
-	Fields []string
+	Key     string
+	Order   string
+	Limit   uint64
+	Offset  uint64
+	Fields  []string
```
marcin-ptaszynski: I don't think this belongs here.

marcin-ptaszynski: This doesn't belong here. Paginator has very specific purpose and resource details ins't it.

mmatczuk:
@marcin-ptaszynski thanks for pointing this out, List API was extended with `*ListOptions` (in future all the `List` parameters can be moved to `ListOptions` like in AWS api for instance). This PR consist of

* Fields filtering was also moved out from Pagination 
* A bug in filtering was fixed
* Usage of testify's mock was dropped (gohan_db_test was using two mocking frameworks in one test)
* Some refactorings in transaction were performed

przemyslaw-dobrowolski-cl:

PR #433, opened 2017-04-05
Wait for test server to start

mmatczuk:
This PR makes `startTestServer` wait for server to start, otherwise on fast machines server test randomly crash (dial failed).


server/server_test.go
```diff
@@ -1322,12 +1324,28 @@ func startTestServer(config string) error {
...
 	}
+
 	go func() {
 		err := server.Start()
 		if err != nil {
 			return
 		}
 	}()
+
+	retry := 3
```
przemyslaw-dobrowolski-cl: This constant is in the middle of the code. It's better to move it to the const section at the top of the source code.

server/server_test.go
```diff
@@ -1322,12 +1324,28 @@ func startTestServer(config string) error {
...
+		conn, err := net.Dial("tcp", server.Address())
+		if err == nil {
+			conn.Close()
+			break
+		}
+		retry--
+		if retry == 0 {
+			return errors.New("server not started")
+		}
+		time.Sleep(50 * time.Millisecond)
```
przemyslaw-dobrowolski-cl: This constant should also be moved to the top

przemyslaw-dobrowolski-cl:
@mmatczuk  could you review the comments?

server/server_test.go
mmatczuk: Moving a well named constant used once over 1300 LOC up seems like micro overkill to me.

mmatczuk: We can have separate const section just before the function, but in this case it's not a big difference.

przemyslaw-dobrowolski-cl: You should look at this the other way: this constant is hidden somewhere in more than 1300 lines source file. A common and good practice is to expose configurable constants at the beginning of a source file - that's why golang has const section. If there are many common constants, it may also be a good idea to have a separate source file - but this is not the case here.

mmatczuk: Inlining things like n * time.Seconds and so on is a common practice, go check yourself https://github.com/golang/go/search?utf8=%E2%9C%93&q=sleep&type=

przemyslaw-dobrowolski-cl: I didn't notice it's a test suite (which is a composition of a separate tests) this constant may stay in this code. In the attached link, all similar usages are also only in tests.

PR #438, opened 2017-04-11
[ESI-11793] Distinguish DB fetch resource not found error

Fenthick:
* DB fetch NotFound case is distinguished from other errors (JS extensions can act based on that information)
* pass all error details in DB fetch error string

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -16,13 +16,12 @@
 package db
 
 import (
-	"fmt"
+	"errors"
 	"strings"
 
 	"github.com/cloudwan/gohan/db/file"
 	"github.com/cloudwan/gohan/db/sql"
 	"github.com/cloudwan/gohan/db/transaction"
-
 	"github.com/cloudwan/gohan/schema"
 	"github.com/cloudwan/gohan/util"
 )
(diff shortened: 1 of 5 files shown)


Fenthick:
@marcin-ptaszynski @p-kozlowski @michalzurawski @przemyslaw-dobrowolski-cl

marcin-ptaszynski:
@Fenthick  I think we should also integrate it to [js layer of errors](https://github.com/cloudwan/gohan/blob/master/extension/otto/gohan_core.go#L113), so it actually causes 404 response. From what I currently see, it doesn't behave like this, or am I missing something?

db/errors/errors.go
```diff
@@ -0,0 +1,11 @@
+package errors
+
+import stderrors "errors"
+
+var (
+	// NoSchemasInManager is returned when there is no schemas in schema manager
+	NoSchemasInManager = stderrors.New("No schemas in Manager. Did you remember to load them?")
+
+	// ResourceNotFound is returned when requested resource was not found
+	ResourceNotFound = stderrors.New("resource not found")
```
p-kozlowski: InternalError is returned when internal error happened.

Sorry, I could not resist. It's 2017, we've got modern IDEs, why do we still write comments?

Fenthick: Obvious comments removed.
Golint shows it as an error. Unfortunately it doesn't have an option to remove that check.

Fenthick:
@marcin-ptaszynski 
There is also PR in extension code: https://bitbucket.org/ntti3/esi-wan-gohan/pull-requests/670/wip-esi-11793-handle-db_fetch-notfound/diff

Fenthick:
Branch rebased.

Fenthick:
@nati Could you take a look at this PR? It has been already approved by other reviewers.

PR #440, opened 2017-04-14
Sync watch load balancing and auto rebalancing

morrisson:
Even if Gohan works as HA, state_watch and sync_watch are done in one
gohan process, and in many cases, both are done in the same process.
This can be a scale problem to handle southbound status update.

This patch enables gohan cluster to load-balance sync_watch keys. In
addition, on process increase/decrease, each process automatically
rebalances watch keys then re-assign keys based on calculated priority.
To enable this feature, the following logics are added.

1. New goroutine process_watch will watch under /gohan/cluster/process
   path. Process ID of running gohan is registered under this path, so
   each gohan process can be notified how many gohan processes are
   running in the same cluster.

2. In rebalance process, the priority for each watch path is calculated
   on each gohan process. For "low priority" path (actually higher
   priority has lower number for caluculation ease), it will take longer
   wait time for other gohan process to preferentially acquire a lock
   for the path.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -376,6 +376,40 @@ Any monitoring updates made when the state version does not yet equal
 the config version or the version in the JSON data doesn't match with
 the config version will be ignored.
 
+## Sync watch
+
+Gohan has another way to handle data reported from sync layer. Gohan
+will watch sync keys with prefix then if its key's creation/update/deletion
+happens, gohan will occur ``notification`` type event with action and data.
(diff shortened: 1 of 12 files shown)


morrisson:
I found a bug in watching sync path so will fix it soon.

nati:
could you add doc about this?

morrisson:
@nati Sure, I will update a PR with docs in a few days.

nati:
Could you also add Unit test using Mock?  I believe we need to have some given this complexity..

PR #443, opened 2017-04-18
ESI-12489 bring back default _details behavior

marcin-ptaszynski:


morrisson:
Let me merge this patch because this patch is clear what is done and is important for backward compatibility (especially for UI).

PR #445, opened 2017-04-20
ESI-12423 gohan client panic

mmatczuk:


mmatczuk:
@marcin-ptaszynski please review this patch

PR #456, opened 2017-05-04
[ESI-12716] `gohan migrate version` doesn't fail in case of error

przemyslaw-dobrowolski-cl:
If in the migration path there are invalid plugins present (compiled with an incorrect version of gohan) when 'migration version' fails, gohan should return a non-zero result value to shell.
This commit fixes the resulting value in case migration plugins are miscompiled.


przemyslaw-dobrowolski-cl:
@nati @marcin-ptaszynski @Zimnx @p-kozlowski please review

przemyslaw-dobrowolski-cl:
@p-kozlowski sure, good idea - fixed.

PR #457, opened 2017-05-04
[ESI 12764] Setting transaction isolation level before tx creation

Zimnx:
SetIsolationLevel method cannot work, because it's forbidden in mysql to change isolation level of current transaction - it must be specified before "begin"

In this commit I introduce ability to create transaction with TxOptions where Isolation Level is defined.
Current code, won't break, since default isolation level is preserved.


Zimnx:
I have tested it on my environment, and I communication with etcd was broken. Until I found out if this PR is a problem I'll close PR.

marcin-ptaszynski:
@Zimnx , could you share your problems with etcd and how it connects to this PR?

PR #424, opened 2017-03-27
pprof server support

morrisson:
For debugging long-term stability of gohan, pprof is useful profiling
tool. This patch enables gohan also starts server for pprof with
configuration in config file.

PR #460, opened 2017-05-11
pprof server support

morrisson:
For debugging long-term stability of gohan, pprof is useful profiling
tool. This patch enables gohan also starts server with profiling
feature.


morrisson:
Reopen PR #424 

I got a comment about performance impact from @nati . I read a source code and googled the same question, then I found that CPU sampling is activated only when requests come, so it should be almost no impact to performance.

nati:
Thank you for your investigation. will merge this

PR #461, opened 2017-05-15
[ESI-12764] getTransaction returns transaction with requested isolation level

Zimnx:


extension/framework/runner/environment.go
```diff
@@ -233,7 +233,7 @@ func (env *Environment) addTestingAPI() {
 func (env *Environment) getTransaction(isNew bool, options *transaction.TxOptions) transaction.Transaction {
 	if !isNew {
 		for _, tx := range env.dbTransactions {
-			if !tx.Closed() {
+			if !tx.Closed() && tx.GetIsolationLevel() == options.IsolationLevel {
```
marcin-ptaszynski: this may result in creation of nested transaction when isolation levels don't match. I think we should throw error instead creating "nested" one. Not truly nested second transaction can freeze while waiting for locks from the first one.

Zimnx:
@marcin-ptaszynski fixed

Zimnx:
I've also changed panic to OttoException. Panic caused all tests to stop runnning, exception fails only single test.

PR #463, opened 2017-05-22
[ESI-13047] Retryable transactions

Zimnx:
Ability to specify strategy to decide whether transaction returning error should be retried automatically.
It's configurable and by default retry is turned off.
The only implemented strategy is on deadlock.

### DIFF
--- a/db/transaction/retryable_transaction.go
+++ b/db/transaction/retryable_transaction.go
@@ -1,0 +1,171 @@
+package transaction
+
+import (
+	"time"
+
+	"strings"
+
+	"github.com/cloudwan/gohan/db/pagination"
+	"github.com/cloudwan/gohan/schema"
+	"github.com/cloudwan/gohan/util"
+	"github.com/jmoiron/sqlx"
+)
+
+const NUMBER_OF_RETRIES_CONFIG_KEY = "database/transaction_retries/attempts"
+const RETRY_STRATEGY_CONFIG_KEY = "database/transaction_retries/strategy"
+const INTERVAL_BETWEEN_ATTEMPTS_CONFIG_KEY = "database/transaction_retries/interval_between_attempts"
+const DEFAULT_NUMBER_OF_RETRIES = 1
+const DEFAULT_RETRY_STRATEGY = "deadlock"
+const DEFAULT_INTERVAL_BETWEEN_ATTEMPTS = "100ms"
+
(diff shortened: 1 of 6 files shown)


db/transaction/retryable_transaction.go
```diff
@@ -0,0 +1,171 @@
...
+
+func readConfig() RetryConfig {
+	if cacheInitialized {
+		return cache
+	}
+	config := util.GetConfig()
+	attempts := config.GetInt(NUMBER_OF_RETRIES_CONFIG_KEY, DEFAULT_NUMBER_OF_RETRIES)
+	configStrategy := config.GetString(RETRY_STRATEGY_CONFIG_KEY, DEFAULT_RETRY_STRATEGY)
+	retryStrategy, ok := strategies[configStrategy]
+	if !ok {
```
marcin-ptaszynski: this should err for good. Also, the default should be applied when strategy is missing (or is empty string, because it's easier).

db/transaction/retryable_transaction.go
```diff
@@ -0,0 +1,171 @@
...
+		SleepInterval: sleepInterval,
+	}
+	cacheInitialized = true
+	return cache
+}
+
+func (t *RetryableTransaction) retryOnError(f func() error) (err error) {
+	for attempt := 1; true; attempt++ {
+		err = f()
+		if err != nil && t.Config.Strategy(err) && (t.Config.Attempts < 0 || attempt < t.Config.Attempts) {
```
p-kozlowski: if err != nil && t.Config.shouldBeRetried(attemp, err)

db/transaction/transaction_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+	BeforeEach(func() {
+		givenRetryStrategy(transaction.IsDeadlock)
+		givenNumberOfAttempts(2)
+		givenSleepInterval(0 * time.Millisecond)
+	})
+
+	Context("Retry logic", func() {
+		var rawTxMock *mocks.MockTransaction
+		var sut transaction.RetryableTransaction
```
p-kozlowski: what does "sut" stand for?

Zimnx: https://en.wikipedia.org/wiki/System_under_test

Zimnx:
@p-kozlowski, that's correct. I'll prepare retrying on different level soon.

Issue #449, opened 2017-04-25
Runtime metrics


p-kozlowski:
I'd like to expose runtime metrics, like number of requests processed per second or min/max/avg request processing time. I was thinking of using https://github.com/rcrowley/go-metrics. This way Gohan could be easily integrated with monitoring frameworks like Graphite/Prometheus/ELK.

@nati, @astralblue, @marcin-ptaszynski, @przemyslaw-dobrowolski-cl what do you think about it?

marcin-ptaszynski:
Looks abstracted enough, that it can be used in generic layer and be fully configurable.

p-kozlowski:
Done in #464

PR #464, opened 2017-05-23
[ESI-12650] Runtime metrics (issue #449)

p-kozlowski:
Currently request processing time, extension execution time and sync/state watch processing time are measured and Graphite backend is available.

### DIFF
--- a/cli/client/commands.go
+++ b/cli/client/commands.go
@@ -361,7 +361,7 @@ func (gohanClientCLI *GohanClientCLI) getFieldsParam(prependQuestionMark bool) s
 			param = param + "&"
 		}
 
-		param  = param + "_fields=" + field
+		param = param + "_fields=" + field
 	}
 
 	return param
--- a/db/db_test.go
+++ b/db/db_test.go
@@ -270,7 +270,7 @@ var _ = Describe("Database operation test", func() {
 				It("Doesn't show related resources when fields is set and nothing is selected", func() {
 					list, num, err := tx.List(serverSchema, nil, &transaction.ListOptions{
 						Details: true,
-						Fields: []string{"id"},
+						Fields:  []string{"id"},
 					}, nil)
 					Expect(err).ToNot(HaveOccurred())
 					Expect(num).To(Equal(uint64(1)))
@@ -283,7 +283,7 @@ var _ = Describe("Database operation test", func() {
 				It("Show related resources when fields is set and something is selected", func() {
 					list, num, err := tx.List(serverSchema, nil, &transaction.ListOptions{
 						Details: true,
-						Fields: []string{"id", "network.name"},
+						Fields:  []string{"id", "network.name"},
 					}, nil)
 					Expect(err).ToNot(HaveOccurred())
 					Expect(num).To(Equal(uint64(1)))
--- a/db/migration/migration.go
+++ b/db/migration/migration.go
@@ -91,7 +91,7 @@ func Init() error {
 			if err != goose.ErrNoNextVersion {
 				return err
 			}
-			last = 0;
+			last = 0
 		} else {
 			last = m.Version
 		}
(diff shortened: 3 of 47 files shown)


p-kozlowski:
We need one more change: we may want to report to more than one Graphite node (think HA).

p-kozlowski:
@nati, branch rebased and ready to merge

nati:
can we disable this feature by configuration? I'm worried about the performance impact of this.

p-kozlowski:
@nati, I've done [measurements](https://github.com/p-kozlowski/go-metrics-benchmark), here's the output on my laptop:
```
~/code/src/github.com/p-kozlowski/go-metrics-benchmark$ go test -bench=. 
BenchmarkMetrics-4                 	 1000000	      1039 ns/op
BenchmarkMetricsLockContention-4   	  500000	      6482 ns/op
PASS
ok  	github.com/p-kozlowski/go-metrics-benchmark	4.381s
```

It means that single timer update takes 1039ns, and if there's [lock contention](https://github.com/rcrowley/go-metrics/blob/master/registry.go#L81), this time increases to 3241ns (6482/2).

In my opinion, this amount of time is negligible.

On the other hand, if you're concerned about the sending process, it already is [opt-in](https://github.com/p-kozlowski/gohan/blob/runtime-metrics/server/metrics.go#L9).

nati:
I don't think this micro benchmark proofs this is negligible. We should have end to end benchmark with high amount of requests.

p-kozlowski:
@nati, I've used [vegeta](https://github.com/tsenart/vegeta) to send 100reqs/s to Gohan server running [the default configuration](https://github.com/cloudwan/gohan/blob/master/etc/gohan.yaml), using latest master and this PR. The command I've run was
```
echo "GET http://192.168.0.4:9091/v1.0/store/orders" | ./vegeta attack -duration=30s -rate=100 | tee results.bin | ./vegeta report
```

**Results**

**master**
```
$ echo "GET http://192.168.0.4:9091/v1.0/store/orders" | ./vegeta attack -duration=30s -rate=100 | tee results.bin | ./vegeta report
Requests      [total, rate]            3000, 100.03
Duration      [total, attack, wait]    29.992928314s, 29.989999741s, 2.928573ms
Latencies     [mean, 50, 95, 99, max]  2.319187ms, 2.241628ms, 2.958864ms, 4.806448ms, 8.689095ms
Bytes In      [total, mean]            54000, 18.00
Bytes Out     [total, mean]            0, 0.00
Success       [ratio]                  100.00%
Status Codes  [code:count]             200:3000  
Error Set:
```
master

**this PR**
```
$ echo "GET http://192.168.0.4:9091/v1.0/store/orders" | ./vegeta attack -duration=30s -rate=100 | tee results.bin | ./vegeta report
Requests      [total, rate]            3000, 100.03
Duration      [total, attack, wait]    29.992424262s, 29.989999718s, 2.424544ms
Latencies     [mean, 50, 95, 99, max]  2.486378ms, 2.395113ms, 3.366532ms, 4.446446ms, 8.378278ms
Bytes In      [total, mean]            54000, 18.00
Bytes Out     [total, mean]            0, 0.00
Success       [ratio]                  100.00%
Status Codes  [code:count]             200:3000  
Error Set:
```
metrics

The the mean delay is 0,17ms, but in the 99th percentile, this PR seemed to be faster by 0,36ms. Therefor I think the overhead is so small that it could be confused with measurement noise. It's also important to note that I was not using etcd and the DB was sqlite. My strong guess is that if we try to run these measurements on a production-like environments (etcd, mysql, JS extensions), the overhead will remain the same, but the request processing time will increase significantly (due to network round-trips and JS processing time).

nati:
DB or test tool limit could be a reason you don't see the difference. 100 req/sec is a small number, so I don't think we put enough stress on API server itself.

p-kozlowski:
I've added an on/off switch, now calling UpdateTimer is no-op if metrics are disabled in the config file.

PR #468, opened 2017-05-26
Added builtin with null support for updates

michalzurawski:
Problem:
- Gohan does not allow setting field to null while updating the database

Solutions:
- add builtin which allows setting chosen fields to null (this commit)
- change otto vendor so it allows setting a null value to dictionary instead of removing its key

@nati this problem has two solutions. The second one, however, requires much more work. Could you please let me know which one suits better?

### DIFF
--- a/docs/js_extension.md
+++ b/docs/js_extension.md
@@ -154,6 +154,10 @@ create data in db
 
 update data in db
 
+- gohan_db_update_with_nulls(transaction, schema_id, object, fields)
+
+update data in db with setting all fields to null
+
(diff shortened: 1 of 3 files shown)


vozhyk-:
The problem is not related to the database, but is only about Otto.

So a built-in like `setNull(object, key)` that does `object[key] = otto.NullValue()` would be a simpler workaround. At least this used to work with dop251/otto; not sure about the Otto that's being used now.

michalzurawski:
Have been done in extensions

PR #470, opened 2017-05-30
Refactor sync watch and state watch

yudai:
# Issues Fixed

* Gohan cannot detect it has lost locks by restarting of/disconnect from the etcd cluster
* Gohan leaks goroutins 

# Resolution

* Add a channel to the return values of `sync.Lock()`, which will be closed when the acquired lock is lost
*  Refactor `server/sync.go` with creating a separate struct named `SyncWriter`.
   * The file is renamed to `sync_writer.go`
   * It stops syncing when it's the lock
   * API is context compatible
   * Some minor fixes like log handling

* Refactor `server/sync_watch.go` with creating a separate struct names `SyncWatcher`.
   * The file is renamed to `sync_watcher.go`
   * `process_watch.go` is merged into this struct as well.
   * It stops watching when it's lost the lock
   * Fixes that it cannot construct the list of Gohan processes properly
   * API is context compatible

* Refactor `server/state_watch.go` with creating a separate struct names `StateWatcher`.
   * The file is renamed to `state_watcher.go`
   * It stops watching when it's lost the lock
   * API is context compatible


server/sync_watcher.go
```diff
@@ -0,0 +1,332 @@
...
+			}
+
+			for {
+				err := watcher.processSyncWatch(ctx, path)
+				if err != nil && err != context.Canceled && err != lockFailedErr {
+					log.Error("SyncWatch on `%s` aborted, retrying...: %s", path, err)
+				}
+
+				select {
+				case <-time.After(time.Second * masterTTL):
```
morrisson: Here should be like this https://github.com/cloudwan/gohan/pull/476/files

yudai:
updated tests :)

marcin-ptaszynski:
@yudai , with those file moves, it's hard to see differences and tell what the point of refactor was. Can you summarize?

yudai:
@marcin-ptaszynski Added some comment to the description.

Basically, this PR was to fix the goroutine leak which happens gohan loses the connection to etcd.
I also discovered that gohan fails to handle locks when etcd is lost, which leads to silent stuck and/or duplicate processing of events.

PR #472, opened 2017-05-31
[ESI-12189] Indexes on multiple columns

Zimnx:
Extended schema with list of 'indexes'
Developer is able to specify on which columns index should be created

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -69,6 +69,12 @@ Developers can make a schema as abstract schema specifying type=abstract. The de
           title: Tenant
           type: string
           unique: false
+      indexes:
+        unique_name_per_tenant_id:
+          columns:
+          - tenant_id
+          - name
+          type: "unique"
       propertiesOrder:
       - id
       - name
@@ -174,7 +180,7 @@ Developers can make a schema as abstract schema specifying type=abstract. The de
   - `lock_related` - locks the resource and related resources
   - `skip_related` - locks the resource but leaves related resources unlocked
   - (empty): default, no locking
-    
+
 ## Properties
 
 We need to define properties of a resource using following parameters.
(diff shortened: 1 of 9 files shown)


schema/index.go
```diff
@@ -0,0 +1,84 @@
...
+		Name:    name,
+		Columns: columns,
+		Type:    indexType,
+	}
+}
+
+func mapIndexType(rawIndexType string) (IndexType, error) {
+	indexType := IndexType(strings.ToUpper(rawIndexType))
+	switch indexType {
+	case Unique:
```
michalzurawski: Maybe just
```go
case Unique, FullText, Spatial, None:
 		return indexType, nil
```

?

docs/schema.md
```diff
@@ -347,6 +353,27 @@ eg.
           unique: false
 ```
 
+## Indexes
+
+Developer can specify which on which columns index should be added.
```
michalzurawski: unnecessary 'which' before 'on'

docs/schema.md
```diff
@@ -347,6 +353,27 @@ eg.
...
+
+Each index has name and following options:
+
+- columns
+    List of column names to be indexed
+- type (optional)
+    Index type, available options are: "spatial", "fulltext", "unique"
```
p-kozlowski: not true for SQLite, which supports "unique" and "" (https://sqlite.org/lang_createindex.html)

docs/schema.md
```diff
@@ -69,6 +69,12 @@ Developers can make a schema as abstract schema specifying type=abstract. The de
           title: Tenant
           type: string
           unique: false
+      indexes:
```
marcin-ptaszynski: Have you tried it openapi generator? If not properly handled https://github.com/cloudwan/gohan/blob/master/cli/template.go it may output invalid schema.

Zimnx: Thanks, indeed swagger was broken.
Done

PR #439, opened 2017-04-12
ESI-12126 gohan migrate version sometimes fails when run on multiple nodes at once

mmatczuk:

PR #473, opened 2017-06-01
[ESI-12126] gohan migrate version sometimes fails

przemyslaw-dobrowolski-cl:
This PR replaces the previous one:
https://github.com/cloudwan/gohan/pull/439

### DIFF
--- a/db/migration/migration.go
+++ b/db/migration/migration.go
@@ -16,9 +16,8 @@
 package migration
 
 import (
-	"fmt"
-
 	"database/sql"
+	"fmt"
 	"os"
 	"path"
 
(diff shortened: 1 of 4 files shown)


db/migration/migration.go
```diff
@@ -126,25 +120,51 @@ func Run(subCmd string, args []string) error {
...
-		return err
+	return nil
+}
+
+// lock issues a lock on a shared mutex to avoid running migrations on the same
+// database in parallel, currently only supported on mysql.
+func lock(dbType string, db *sql.DB) error {
+	switch dbType {
+	case "mysql":
+		if _, err := db.Exec("LOCK TABLES `goose_db_version` WRITE"); err != nil {
```
p-kozlowski: this may not be the right thing to do, I think we have to use etcd instead:
> LOCK TABLES doesn't work when using Galera cluster. You may experience crashes or locks when used with Galera.

https://mariadb.com/kb/en/mariadb/lock-tables-and-unlock-tables/

przemyslaw-dobrowolski-cl: I'm closing this PR now since the presented solution is wrong in case of Galera cluster and from the ops-team perspective this can be easily handled.

PR #474, opened 2017-06-05
[ESI-12189] retry_create_on_duplicate_entry metadata flag

Zimnx:
Retrying create logic when DB returns 'Duplicate Entry' error.
Useful for logic where some field in resource is generated during pre_create event and unique index is applied to this field.

### DIFF
--- a/db/errors.go
+++ b/db/errors.go
@@ -1,0 +1,35 @@
+package db
+
+import (
+	"github.com/go-sql-driver/mysql"
+	"github.com/mattn/go-sqlite3"
+)
+
+func IsDeadlockError(err error) bool {
+	switch t := err.(type) {
+	case *mysql.MySQLError:
+		switch t.Number {
+		case 1213:
+			return true
+		}
+	}
+
+	return false
+}
+
+func IsDuplicateEntryError(err error) bool {
+	switch t := err.(type) {
+	case *mysql.MySQLError:
+		switch t.Number {
+		case 1062:
+			return true
+		}
+	case sqlite3.Error:
+		switch t.ExtendedCode {
+		case sqlite3.ErrConstraintUnique:
+			return true
+		}
+	}
+
+	return false
+}
--- a/db/sql/sql.go
+++ b/db/sql/sql.go
@@ -397,7 +397,7 @@ func (db *DB) genTableCols(s *schema.Schema, cascade bool, exclude []string) ([]
 
 		if property.Indexed {
 			prefix := ""
-			if sqlDataType == "text" {
+			if sqlDataType == "text" && db.sqlType != "sqlite3" {
 				prefix = "(255)"
 			}
 			indices = append(indices, fmt.Sprintf("CREATE INDEX %s_%s_idx ON `%s`(`%s`%s);", s.Plural, property.ID,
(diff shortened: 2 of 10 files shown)


db/sql/sql.go
```diff
@@ -469,8 +486,9 @@ func (db *DB) RegisterTable(s *schema.Schema, cascade, migrate bool) error {
 		return nil
 	}
 	_, err = db.DB.Exec(tableDef)
-	if err != nil && indices != nil {
+	if err == nil && indices != nil {
 		for _, indexSql := range indices {
+			fmt.Printf("Creating index %s\n", indexSql)
```
p-kozlowski: log.Debug?

db/errors.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func IsDuplicateEntryError(err error) bool {
+	switch t := err.(type) {
+	case *mysql.MySQLError:
+		switch t.Number {
+		case 1062:
+			return true
+		}
+	case sqlite3.Error:
```
p-kozlowski: *mysql.MySQLError vs. sqlite3.Error: why do sometimes use pointer, and sometimes use a value?

db/sql/sql.go
```diff
@@ -469,8 +486,9 @@ func (db *DB) RegisterTable(s *schema.Schema, cascade, migrate bool) error {
 		return nil
 	}
 	_, err = db.DB.Exec(tableDef)
-	if err != nil && indices != nil {
+	if err == nil && indices != nil {
```
p-kozlowski: nice catch. so until now, we've never had indices?

server/resources/resource_management_test.go
```diff
@@ -1133,6 +1133,33 @@ var _ = Describe("Resource manager", func() {
 				Expect(ok).To(BeTrue())
 			})
 		})
+
+		Describe("Retry during creation", func() {
```
p-kozlowski: Add a test for context deep-copy

docs/schema.md
```diff
@@ -174,7 +180,13 @@ Developers can make a schema as abstract schema specifying type=abstract. The de
   - `lock_related` - locks the resource and related resources
   - `skip_related` - locks the resource but leaves related resources unlocked
   - (empty): default, no locking
-    
+
+- retry_create_on_duplicate_entry (boolean)
+
+  Whether create logic should be retried in case of "Duplicate Entry" on DB. Default to false.
+  Useful when pre_create logic is setting some generated data on column where unique index was applied.
```
p-kozlowski: smells like we need the same mechanism for update and (maybe) delete

db/sql/sql.go
Zimnx: Yes

db/sql/sql.go
Zimnx: Thanks, forgot about it

db/errors.go
Zimnx: depends on library, mysql returns pointers, sqlite returns object.

docs/schema.md
Zimnx: @nati what do you think about it?

server/resources/resource_management_test.go
Zimnx: Done

docs/schema.md
marcin-ptaszynski: @p-kozlowski , agreed. It would be better to allow extensible error handler with with some flow control capabilities (retry, return, etc). With single flag `retry_create_on_duplicate_entry`, we can expect many more.

Zimnx:
@nati kindly reminder about this PR. Could you share your thoughts?

Zimnx:
Closed since problem can be resolved using proper transaction isolation level which is implemented in #479

PR #475, opened 2017-06-05
[ESI-12271] Informative error messages on failed fetch

p-kozlowski:


server/resources/resource_management.go
```diff
@@ -361,8 +361,14 @@ func GetSingleResourceInTransaction(context middleware.Context, resourceSchema *
...
 	object, err := mainTransaction.Fetch(resourceSchema, filter)
 
-	if err != nil || object == nil {
-		return ResourceError{err, "", NotFound}
+	if object == nil {
+		switch err {
+		case transaction.ErrResourceNotFound:
+			return ResourceError{err, "Resource not found", NotFound}
+		default:
+			log.Error("Fetch failed: %v", err)
```
marcin-ptaszynski: should we log error regardless of response?

p-kozlowski: trying to fetch or delete a non-existing resource is not an error condition, so let's do it on INFO level

PR #478, opened 2017-06-09
ESI-11623 update mysql driver vendor package

marcin-ptaszynski:
This enables proper `BeginTx` which properly configures isolation levels.


marcin-ptaszynski:
should've read source before posting this: https://github.com/go-sql-driver/mysql/blob/master/connection_go18.go#L44 Actually, only context is supported. Isolation levels aren't.

PR #481, opened 2017-06-23
Fixed color output in Makefile

przemyslaw-dobrowolski-cl:
echo does not emit colors by default (at least on arch-linux) so it must be called with parameter -e to see color output


przemyslaw-dobrowolski-cl:
@marcin-ptaszynski @Zimnx could you review this PR?

przemyslaw-dobrowolski-cl:
@nati rebased

Issue #454, opened 2017-05-01
Gohan doesn't create db table name from schema.plural


j-maxi:
I found that gohan creates db table name from schema.id + 's' instead of using schema.plural.
https://github.com/cloudwan/gohan/blob/master/schema/schema.go#L315

This doesn't affect user, but confuses developer/operator.

PR #483, opened 2017-07-04
ESI-12742 Gohan schema sql tables dont use correct plurals

jkmar:
Some tables might have different names if used on existing database. Data will not be lost but new empty tables with correct names will be created.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -33,6 +33,10 @@ Gohan server configuration uses YAML format.
       initial_data:
           - type: "yaml"
             connection: "./etc/examples/heat_template.yaml"
+      # if legacy is true table names are created from schema id,
+      # otherwise table names are based on schema plural,
+      # default true
+      legacy: false
   # schema path
   schemas:
     - "./etc/schema/gohan.json"
@@ -43,7 +47,7 @@ Gohan server configuration uses YAML format.
     cert_file: "./etc/cert.pem"
     key_file: "./etc/key.pem"
   # document root of gohan API server
-  # Note: only static and schema directoriy will be served
+  # Note: only static and schema directory will be served
   document_root: "./etc"
   # list of etcd backend servers
   etcd:
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -323,7 +323,11 @@ func (schema *Schema) GetPluralURLWithParents() string {
 
 // GetDbTableName returns a name of DB table used for storing schema instances
 func (schema *Schema) GetDbTableName() string {
-	return schema.ID + "s"
+	config := util.GetConfig()
+	if config.GetBool("database/legacy", true) {
+		return schema.ID + "s"
+	}
+	return schema.Plural
 }
 
 // GetParentURL returns Parent URL
--- a/schema/schema_test.go
+++ b/schema/schema_test.go
@@ -18,11 +18,44 @@ package schema
 import (
 	"fmt"
 
+	"github.com/cloudwan/gohan/util"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
 var _ = Describe("Schema", func() {
+	Describe("Schema plural", func() {
+		var (
+			manager *Manager
+			s       *Schema
+			ok      bool
+		)
+
+		BeforeEach(func() {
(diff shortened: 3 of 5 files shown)


vozhyk-:
This is related to #454.

schema/schema_test.go
```diff
@@ -23,6 +23,26 @@ import (
...
+		BeforeEach(func() {
+			manager = GetManager()
+			Expect(manager.LoadSchemasFromFiles(
+				"../tests/test_schema_plural.yaml")).To(Succeed())
+		})
+
+		It("should use plural as table name", func() {
+			s, ok := manager.Schema("box")
+			Expect(ok).To(BeTrue())
+			Expect(s.GetDbTableName()).To(Equal("boxes"))
```
Zimnx: use s.Plural as expected value instead of hardcoded "boxes"

morrisson:
The idea should be ok, but this apparently breaks backward compatibility. Updating gohan which includes this patch will not work with our current system. It's really dangerous. This patch must provide a logic to support "older" version.

schema/schema.go
```diff
@@ -323,7 +323,7 @@ func (schema *Schema) GetPluralURLWithParents() string {
 
 // GetDbTableName returns a name of DB table used for storing schema instances
 func (schema *Schema) GetDbTableName() string {
```
morrisson: Need older version support.

marcin-ptaszynski: I think some "legacy_table_names" option in config db section should do. What do you think?

vozhyk-:
Having a way to generate a DB migration from old to new names would be useful.

schema/schema.go
morrisson: Either seamless way or config way is fine to me.

przemyslaw-dobrowolski-cl:
@jkmar please apply the comments and rebase

jkmar:
@morrisson @nati 
I've applied changes requested in comments. Please review again.

docs/configuration.md
```diff
@@ -33,6 +33,10 @@ Gohan server configuration uses YAML format.
       initial_data:
           - type: "yaml"
             connection: "./etc/examples/heat_template.yaml"
+      # if legacy is true table names are based on schema plural,
+      # otherwise table names are created from schema id,
+      # default false
+      legacy: true
```
morrisson: I approve this change. However, we need to take care when upgrading gohan. Upgrade might break existing environment.

schema/schema.go
```diff
@@ -323,6 +323,10 @@ func (schema *Schema) GetPluralURLWithParents() string {
...
+	config := util.GetConfig()
+	if config.GetBool("database/legacy", false) {
+		return schema.Plural
```
morrisson: Here, we should use schema.ID + "s"

schema/schema.go
```diff
@@ -323,6 +323,10 @@ func (schema *Schema) GetPluralURLWithParents() string {
...
+	}
 	return schema.ID + "s"
```
morrisson: we should use schema.Plural

schema/schema_test.go
```diff
@@ -18,11 +18,44 @@ package schema
...
+		BeforeEach(func() {
+			manager = GetManager()
+			Expect(manager.LoadSchemasFromFiles(
+				"../tests/test_schema_plural.yaml")).To(Succeed())
+
+			s, ok = manager.Schema("box")
+			Expect(ok).To(BeTrue())
+		})
+
+		It("should use schema.ID + \"s\" as table name when legacy is false", func() {
```
morrisson: when legacy is TRUE

docs/configuration.md
```diff
@@ -33,6 +33,10 @@ Gohan server configuration uses YAML format.
       initial_data:
           - type: "yaml"
             connection: "./etc/examples/heat_template.yaml"
+      # if legacy is true table names are based on schema plural,
```
morrisson: I believe legacy == true should mean schema.ID + "s".

jkmar:
@morrisson 
I've changed the meaning of legacy, now legacy == true is default and means schema.ID + "s".

morrisson:
@jkmar thanks. merged.

PR #486, opened 2017-07-06
ESI-13579 Fix isolation overrides

marcin-ptaszynski:


db/transaction/transaction_test.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+var _ = Describe("Transaction", func() {
+	var netSchema *schema.Schema
+
+	Describe("GetTransactionIsolationLevel", func() {
+		BeforeEach(func() {
+			var exists bool
+			manager := schema.GetManager()
+			basePath := "../..//tests/test_abstract_schema.yaml"
```
morrisson: nit: double slashed path

marcin-ptaszynski: done

PR #487, opened 2017-07-07
added basic bash completion script

andriikonstantinov:


tools/bash_completion
```diff
@@ -0,0 +1,43 @@
+__gohan()
+{
+
+
+    declare -A possibleWord=( ["1"]="gohan"
+    ["2"]="--debug --help --version -d -h -v"
+    ["3"]="client validate v init-db idb convert conv server srv test_extensions test_ex migrate mig template run test test openapi markdown dot glace-server gsrv help h")
```
p-kozlowski: completion for `gohan client` will be much appreciated

przemyslaw-dobrowolski-cl:
there must be unit tests for this code

FYI: https://stackoverflow.com/questions/9137245/unit-test-for-bash-completion-script

tools/bash_completion
andriikonstantinov: This task has been done in this commit.
https://github.com/cloudwan/gohan/pull/487/commits/e775c76883c38c22d93467955055deafea0da2a3

andriikonstantinov:
@przemyslaw-dobrowolski-cl please review again.

tools/gohan_client_bash_completion.sh
```diff
@@ -0,0 +1,137 @@
...
+        fi
+done
+
+}
+
+getAllSchemas()
+{
+OIFS=$IFS
+IFS=$'\n'
+array=`gohan client`
```
vozhyk-: This doesn't work when the server can't be reached or when a required environment variable (like `GOHAN_SERVICE_NAME`) is not set.

Checking `gohan client`'s return value would fix it.

tools/bash_completion_instalation
```diff
@@ -0,0 +1 @@
+To install bash-completion - fire ". gohan_path/tools/bash_completion"
```
vozhyk-: Please update it to describe `gohan_client_bash_completion.sh`.

tools/bash_completion
vozhyk-: I couldn't get both `bash_completion` and `gohan_client_bash_completion.sh` to work at the same time. Sourcing one made the other have no effect, and vice-versa.

tools/gohan_client_bash_completion_tests.py
```diff
@@ -0,0 +1,74 @@
...
+    def test_7(self):
+        self.run_complete("client pet show --fields ","id name tenant_id description status")
+    def test_8(self):
+        self.run_complete("client namespace show --fields ","id name description prefix parent version metadata")
+    def test_9(self):
+        self.run_complete("client namespace show --verbosity ","0 1 2")
+    def test_10(self):
+        self.run_complete("client policy show --fields ","id principal resource action effect condition")
+    def test_11(self):
+        self.run_complete("client policy something",""  )
```
vozhyk-: `gohan client schema list <TAB>` produces
```
event      extension  namespace  pet        policy     schema
```
Is this the intended behavior?

vozhyk-: This doesn't seem to work when:
1. `gohan server --config-file examples/db_migration/gohan.yaml`
2. `gohan client network create --name test`

In this case, the created network's ID is inserted when requesting completions for any action (e.g. `gohan client network list <TAB>` or `gohan client network whatever <TAB>`).

tools/gohan_client_bash_completion_tests.py
```diff
@@ -0,0 +1,74 @@
...
+    def test_2(self):
+        self.run_complete("client ", "namespace event extension policy schema pet order")
+    def test_3(self):
+        self.run_complete("client pet","pet")
+    def test_4(self):
+        self.run_complete("client pet ","list show create set delete")
+    def test_5(self):
+        self.run_complete("client pet s","show set")
+    def test_6(self):
+        self.run_complete("client pet show -","--output-format --verbosity --fields --id --name --tenant_id --description --status")
```
andriikonstantinov: According to the documentation, there are 2 types of arguments: named and unnamed.Named arguments start with "--".That's why I show unnamed arguments when there is no '-'.

tools/bash_completion_tests.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/python
```
andriikonstantinov: @przemyslaw-dobrowolski-cl attached

andriikonstantinov:
@vozhyk- please review

przemyslaw-dobrowolski-cl:
@vozhyk- please review again

tools/bash_completion
```diff
@@ -0,0 +1,40 @@
+__gohan()
+{
+
```
przemyslaw-dobrowolski-cl: are this new lines required?

tools/gohan_client_bash_completion.sh
```diff
@@ -0,0 +1,151 @@
...
+prev="${COMP_WORDS[COMP_CWORD-1]}"
+c=" "
+let "c=(${COMP_CWORD}+$gohanCommandPos) % 2"
+if [[ "${COMP_CWORD}" -lt $((gohanCommandPos+1)) ]] ; then
+    return 0
+fi
+if [[ "${COMP_WORDS[0]}" != "gohan" ]] &&  [[ "${COMP_WORDS[1]}" != "client" ]] ; then
+    return 0
+fi
+if [[ `${COMP_WORDS[0]} client` == "Environment variable GOHAN_SERVICE_NAME needs to be set" ]] ; then
```
vozhyk-: This will break when a different error happens. It's better to check the exit code. [here](https://superuser.com/a/363454/583324)'s how it can be done for ``.

tools/gohan_client_bash_completion.sh
```diff
@@ -0,0 +1,151 @@
...
+commands="list show create set delete"
+verbosity="0 1 2"
+outputFormat="json table"
+gohanCommandPos=1
+unnamedCommandWords="show set delete"
+getUnnamedProperties()
+{
+schema_id=$1
+OIFS=$IFS
+IFS=$'\n' arr=`${COMP_WORDS[0]} client $schema_id list`
```
vozhyk-: I'd extract `${COMP_WORDS[0]}` to a new `GOHAN` variable.

tools/gohan_client_bash_completion.sh
```diff
@@ -0,0 +1,151 @@
...
+        if [[ $counter -gt 2 ]];then
+        args=$(echo $a | tr "|" "\n")
+        sub=0
+        for i in $args
+        do
+
+            let sub=sub+1
+            if [ $sub -eq 1 ] && [ ! -z "${i// }" ] ; then
+        		i="$(echo -e "${i}" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')"
+                unnamed="$i $unnamed"
```
vozhyk-: Most of this function can be replaced by just
```bash
unnamed="$(gohan client "$schema_id" list | grep -v -e '+-\++' -e '^[[:space:]]*$' | tail -n +2 | cut -d'|' -f2 | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')"
```
(The last 2 commands in the pipe can be replaced by one `set` command, but it would make it somewhat less readable.)

tools/gohan_client_bash_completion.sh
```diff
@@ -0,0 +1,151 @@
...
+        continue
+        fi
+        if [[ $counter -gt 2 ]];then
+        args=$(echo $a | tr "|" "\n")
+        sub=0
+        for i in $args
+        do
+
+            let sub=sub+1
+            if [ $sub -eq 1 ] && [ ! -z "${i// }" ] ; then
```
vozhyk-: From the little testing I did, `gohan client some_schema list` always has the ID as the first column. It would be better if the completion script didn't rely on this, though.

vozhyk-: This works and doesn't depend on the order of columns:
```bash
unnamed="$(${COMP_WORDS[0]} client "$schema_id" list | 
    grep -v -e '+-\++' -e '^[[:space:]]*$' | 
    awk -v col='ID' -F '|' 'NR==1 { for (i=1; i<=NF; i++) if ($i ~ "[[:space:]]*" col "[[:space:]]*") { c=i; break } } NR>1 { print $c }')"
```
(the awk is adapted from https://stackoverflow.com/a/32612999/795068).

tools/gohan_client_bash_completion.sh
```diff
@@ -0,0 +1,151 @@
+#!/bin/bash
+
+
+commands="list show create set delete"
+verbosity="0 1 2"
+outputFormat="json table"
+gohanCommandPos=1
+unnamedCommandWords="show set delete"
```
vozhyk-: These variables are leaked to the environment. Ideally, they shouldn't be visible outside the completion script.

Maybe there's a way to make them not visible. Or at least it's possible to add some underscores in the beginning of names.

vendor/github.com/codegangsta/cli/README.md
```diff
@@ -295,7 +295,7 @@ setting the `PROG` variable to the name of your program:
 
 Copy and modify `autocomplete/bash_autocomplete` to use your program name
 rather than `$PROG` and have the user copy the file into
-`/etc/bash_completion.d/` (or automatically install it there if you are
+`/etc/bash_completion.sh.d/` (or automatically install it there if you are
```
vozhyk-: Revert this

tools/gohan_client_bash_completion.sh
vozhyk-: It's fixed above now, so you can remove this.

tools/bash_completion_instalation
```diff
@@ -1 +1 @@
-To install bash-completion - fire ". gohan_path/tools/bash_completion" and ". gohan_path/tools/gohan_client_bash_completion.sh"
\ No newline at end of file
+To install bash-completion - fire ". gohan_path/tools/bash_completion.sh" and ". gohan_path/tools/gohan_client_bash_completion.sh"
```
vozhyk-: Don't mention the second script.

tools/bash_completion.sh
```diff
@@ -0,0 +1,176 @@
+#!/bin/bash
```
vozhyk-: Make this file executable.

p-kozlowski:
LGTM, just apply @vozhyk- comments

tools/gohan_client_bash_completion_tests.py
vozhyk-: OK.

tools/gohan_client_bash_completion_tests.py
przemyslaw-dobrowolski-cl: ok

PR #488, opened 2017-07-07
ESI-11996 [Gohan] protect against migrations race

jkmar:
Migration subcomands now use etcd lock before runing commands.

Additionally util/etcd.go CreateFromConfig now returns error - "no etcd in config file"
instead of nil sync when etcd info cannot be found in config file.

### DIFF
--- a/cli/cli_test.go
+++ b/cli/cli_test.go
@@ -1,0 +1,70 @@
+// Copyright (C) 2017 NTT Innovation Institute, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+import (
+	"flag"
+	"fmt"
+	"github.com/cloudwan/gohan/sync/etcdv3"
+	"github.com/codegangsta/cli"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"sync"
+	"time"
+)
+
+func getContextWithConfig(configFile string) *cli.Context {
+	configFlag := cli.StringFlag{Name: "config-file", Value: configFile}
+	set := flag.NewFlagSet("", flag.ContinueOnError)
+	configFlag.Apply(set)
+	return cli.NewContext(nil, set, &cli.Context{})
+}
+
+var _ = Describe("CLI", func() {
+	Describe("Post migration subcommand wrapper tests", func() {
+		It("Should lock - migrationsSubCommand wrapper", func() {
+			const (
+				configPath = "../tests/test_etcd.yaml"
+				etcdServer = "http://127.0.0.1:2379"
+			)
+
+			waitForLock := sync.WaitGroup{}
+			waitForFail := sync.WaitGroup{}
+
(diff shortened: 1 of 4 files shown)


cli/cli_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+var _ = Describe("CLI", func() {
+	Describe("Post migration subcommand wrapper tests", func() {
+		It("Should lock - migrationsSubCommand wrapper", func() {
+			const (
+				configPath = "../tests/test_etcd.yaml"
+				etcdServer = "http://127.0.0.1:2379"
+			)
+
+			x := 0
```
p-kozlowski: an atomic would be enough, no need for a mutex (cf. https://golang.org/pkg/sync/atomic/)

cli/cli_test.go
```diff
@@ -0,0 +1,83 @@
...
+				x++
+			}
+
+			read := func() int {
+				mutex.RLock()
+				defer mutex.RUnlock()
+				return x
+			}
+
+			s, err := etcdv3.NewSync([]string{etcdServer}, time.Second)
```
p-kozlowski: I know that some Golang conventions advice to use one-letter variable names, but we don't adhere to those

cli/cli_test.go
```diff
@@ -0,0 +1,83 @@
...
+				defer s.Unlock(syncMigrationsPath)
+				c <- 0
+				time.Sleep(masterTTL / 2 * time.Second)
+			}
+
+			wrapped := migrationSubcommandWrapper(change)
+			context := getContextWithConfig(configPath)
+			Expect(context.String("config-file")).To(Equal(configPath))
+
+			c := make(chan int)
```
p-kozlowski: what you need here is a WaitGroup (cf. https://golang.org/pkg/sync/#WaitGroup), not a channel

cli/cli_test.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+
+			s, err := etcdv3.NewSync([]string{etcdServer}, time.Second)
+			Expect(err).ToNot(HaveOccurred())
+
+			block := func(c chan int) {
+				s.Lock(syncMigrationsPath, true)
+				defer s.Unlock(syncMigrationsPath)
+				c <- 0
+				time.Sleep(masterTTL / 2 * time.Second)
```
p-kozlowski: why masterTTL?

cli/cli_test.go
```diff
@@ -0,0 +1,83 @@
...
+			wrapped := migrationSubcommandWrapper(change)
+			context := getContextWithConfig(configPath)
+			Expect(context.String("config-file")).To(Equal(configPath))
+
+			c := make(chan int)
+			go block(c)
+			<-c
+			go wrapped(context)
+
+			time.Sleep(masterTTL / 2 * time.Second)
```
p-kozlowski: tests that rely on timing tend to be flaky. is it possible to rewrite this test is a way that it does not relate on timing? this way we will make the test run much faster (no need for a multisecond sleep) and we'll avoid flakiness on slow machines.

sync/util/etcd.go
```diff
@@ -15,22 +15,21 @@ var log = l.NewLogger()
 
 func CreateFromConfig(config *util.Config) (s sync.Sync, err error) {
 	syncType := config.GetString("sync", "etcd")
+	etcdServers := config.GetStringList("etcd", nil)
```
p-kozlowski: I'm aware this is Not-Your-Code, and it's great you've refactored it a bit. Please remember to always add UTs for the changes you're making.

jkmar:
@p-kozlowski @przemyslaw-dobrowolski-cl
I've applied changes requested in comments. Please review again.

cli/cli_test.go
p-kozlowski: same problem exists in  sync/util/etcd.go

jkmar:
@nati @marcin-ptaszynski @morrisson 
Please review.

PR #489, opened 2017-07-11
ESI-13735 Re-sync resources CLI extension

marcin-ptaszynski:
Resync command takes server config and based on it, resyncs all
syncable resources to sync backend.

Especially useful for recreating environment on fresh etcd cluster

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -21,32 +21,30 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"runtime"
 	"strings"
 	"time"
 
-	"github.com/cloudwan/gohan/schema"
-	"github.com/cloudwan/gohan/util"
+	"github.com/codegangsta/cli"
 	"github.com/lestrrat/go-server-starter"
 
 	"github.com/cloudwan/gohan/cli/client"
 	"github.com/cloudwan/gohan/db"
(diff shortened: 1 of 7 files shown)


cli/cli.go
```diff
@@ -309,6 +308,52 @@ Useful for development purposes.`,
...
+			}
+			if err := os.Chdir(path.Dir(configFile)); err != nil {
+				panic(fmt.Errorf("Chdir error: %s", err))
+			}
+
+			dbConn, err := db.CreateFromConfig(config)
+			sync, err := sync_util.CreateFromConfig(config)
+			if err != nil {
+				log.Error("Failed to create sync, err: %s", err)
+
```
morrisson: Should gohan exit if sync is not available?

server/sync_writer.go
```diff
@@ -113,18 +113,19 @@ func (writer *SyncWriter) Run(ctx context.Context) error {
 // Sync runs a synchronization iteration, which
 // executes requests in the event table.
 // todo: make private once tests are fixed
```
morrisson: nit: this todo line should be deleted now Sync is public.

cli/cli.go
marcin-ptaszynski: done, using log.Fatal(f) now, which exits.

server/sync_writer.go
marcin-ptaszynski: done

PR #490, opened 2017-07-11
Make sync_watch events sequential for each watch path

morrisson:
Previously, sync_watcher hooked notification events as soon as it detects
sync states for watching path. That makes multiple events for the same
path can happen in very short time slot, so this makes it difficult to
avoid race problem.

This feature was introduced by gohan's internal task queue which allows
parallelism. By simply avoiding to use this queue to execute task, gohan
can assure sequential event notifications for each sync path.

Additionally, this patch introduces JS builtin gohan_sync_update. This
builtin is used in sync_watcher_test, but that should be general feature
like gohan_sync_fetch or gohan_sync_delete.

### DIFF
--- a/extension/otto/gohan_sync.go
+++ b/extension/otto/gohan_sync.go
@@ -17,9 +17,10 @@ package otto
 
 import (
 	"fmt"
+	"time"
+
 	"github.com/cloudwan/gohan/sync"
 	"github.com/xyproto/otto"
-	"time"
 )
 
 func convertSyncEvent(event *sync.Event) map[string]interface{} {
@@ -73,26 +74,26 @@ func init() {
 					return otto.NullValue()
 				}
 
-				done := make(chan struct{})
+				errCh := make(chan error)
+				defer close(errCh)
 				go func() {
 					node, err = env.Sync.Fetch(path)
-					close(done)
+					errCh <- err
 				}()
 
 				select {
 				case interrupt := <-call.Otto.Interrupt:
 					log.Debug("Received otto interrupt in gohan_sync_fetch")
 					interrupt()
-				case <-done:
-				}
+				case err = <-errCh:
+					if err != nil {
+						ThrowOttoException(&call, "Failed to fetch sync: "+err.Error())
+						return otto.NullValue()
+					}
 
-				if err != nil {
-					ThrowOttoException(&call, "Failed to fetch sync: "+err.Error())
-					return otto.NullValue()
-				}
-
-				if value, err = vm.ToValue(convertSyncNode(node)); err == nil {
-					return value
+					if value, err = vm.ToValue(convertSyncNode(node)); err == nil {
+						return value
+					}
 				}
 
 				return otto.NullValue()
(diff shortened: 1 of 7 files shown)


extension/otto/gohan_sync.go
```diff
@@ -136,6 +137,42 @@ func init() {
 
 				return otto.NullValue()
 			},
+			"gohan_sync_update": func(call otto.FunctionCall) otto.Value {
```
marcin-ptaszynski: is there a reason to include unrelated feature (this builtin) here?

server/sync_watcher.go
```diff
@@ -272,13 +272,8 @@ func (watcher *SyncWatcher) processSyncWatch(ctx context.Context, path string) e
 				if response.Err != nil {
 					return response.Err
 				}
+				watcher.watchExtensionHandler(response)
 
-				watcher.queue.Add(job.NewJob(
```
marcin-ptaszynski: I was hoping this could be achieved by configuration change (parallelism settings per watcher or sth). Maybe we shouldn't remove the whole feature.

extension/otto/gohan_sync.go
morrisson: I described why gohan_sync_update is included in this patch. Honestly, I wondered if I should create separate patch, but this code is basically copy-and-paste from gohan_sync_delete so I think it's not so important enough to create a separate PR, IMO. Of course, if other reviewers also requested, I can create a separate patch.

server/sync_watcher.go
morrisson: I think it is really rare to control this kind of finely granular logic therefore I do not optionize it. What do you think? @yudai

yudai: I'm not sure if there is use cases that requires non-sequential processing?

marcin-ptaszynski: Anyway, this PR removes use of `watcher.queue`, but doesn't remove queue itself.

morrisson: Thanks. Done.

przemyslaw-dobrowolski-cl:
@morrisson Any update on this PR? It looks like comments have been resolved but no more updates.

morrisson:
I think I do not need more updates to this patch so far because I showed my opinions to all the comments, so I just wait reviews. If I hear more disagreements, I will consider the change (or separating patches?).

extension/otto/gohan_sync.go
```diff
@@ -137,6 +138,42 @@ func init() {
...
+					return otto.NullValue()
+				}
+				if value, err = GetString(call.Argument(1)); err != nil {
+					ThrowOttoException(&call, "Invalid type of second argument: expected a string")
+					return otto.NullValue()
+				}
+
+				done := make(chan struct{})
+				go func() {
+					err = env.Sync.Update(path, value)
```
yudai: should sent via a channel, maybe to `done`

morrisson: done. applied to other related lines.

morrisson:
@marcin-ptaszynski Are you still requesting changes? Or it's ok so far. Could you review this again?

PR #496, opened 2017-07-18
bindata.go is a generated file

przemyslaw-dobrowolski-cl:
It should not be included in repo and should be ignored in .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 *.a
 *.so
 
+# Generated objects
+bindata.go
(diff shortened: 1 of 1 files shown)


przemyslaw-dobrowolski-cl:
Unfortunately, this change will cause 'go get' to fail since there are no post-go-get hooks that can recreate this file. I'm closing this PR.

vozhyk-:
I've created issue #498 as a follow-up.

PR #497, opened 2017-07-19
[ESI-13782] Watch() should not skip values when asked for a specific revision

p-kozlowski:
This PR fixes a situation when, for a specific key, there are revision r1, r2 and r3. When Watch() was called with r1, it will not report r2 in the results, but skip directly to r3 (and future revisions).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,9 +9,9 @@ machine:
 
 dependencies:
   override:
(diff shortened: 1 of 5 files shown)


p-kozlowski:
@yudai @morrisson @zimnx @marcin-ptaszynski 
Reviewers, gentle reminder about this PR

PR #504, opened 2017-07-21
ESI-13857 Clear lint/vet warnings in gohan

jkmar:
cleared go lint/vet warning that didn't require to change logic

Go vet detected some errors fixing which would require to change logic, here is a list of them:
server/state_watcher.go:100 : the watchCancel function is not used on all paths (possible context leak)
server/state_watcher.go:103: this return statement may be reached without using the watchCancel var defined on line 100
server/sync_watcher.go:103: the watchCancel function is not used on all paths (possible context leak)
server/sync_watcher.go:106: this return statement may be reached without using the watchCancel var defined on line 103
server/sync_watcher.go:200: the previousCancel function is not used on all paths (possible context leak)
server/sync_watcher.go:212: this return statement may be reached without using the previousCancel var defined on line 200
server/sync_watcher.go:262: the watchCancel function is not used on all paths (possible context leak)
server/sync_watcher.go:266: this return statement may be reached without using the watchCancel var defined on line 262
util/util.go:181: using resp before checking for errors
sync/etcdv3/etcd.go:50: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak

### DIFF
--- a/cli/client/commands.go
+++ b/cli/client/commands.go
@@ -110,7 +110,7 @@ func (gohanClientCLI *GohanClientCLI) getGetCommand(s *schema.Schema) gohanComma
 					buffer.WriteString(childSchema.Title)
 					buffer.WriteString("\n")
 					parentSchemaPropertyID := childSchema.ParentSchemaPropertyID()
-					url := fmt.Sprintf("%s%s?%s=%s", gohanClientCLI.opts.gohanEndpointURL, childSchema.URL, parentSchemaPropertyID, id, gohanClientCLI.getFieldsParam(false))
+					url := fmt.Sprintf("%s%s?%s=%s%s", gohanClientCLI.opts.gohanEndpointURL, childSchema.URL, parentSchemaPropertyID, id, gohanClientCLI.getFieldsParam(false))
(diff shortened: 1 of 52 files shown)


cli/client/commands.go
```diff
@@ -110,7 +110,7 @@ func (gohanClientCLI *GohanClientCLI) getGetCommand(s *schema.Schema) gohanComma
 					buffer.WriteString(childSchema.Title)
 					buffer.WriteString("\n")
 					parentSchemaPropertyID := childSchema.ParentSchemaPropertyID()
-					url := fmt.Sprintf("%s%s?%s=%s", gohanClientCLI.opts.gohanEndpointURL, childSchema.URL, parentSchemaPropertyID, id, gohanClientCLI.getFieldsParam(false))
+					url := fmt.Sprintf("%s%s?%s=%s%s", gohanClientCLI.opts.gohanEndpointURL, childSchema.URL, parentSchemaPropertyID, id, gohanClientCLI.getFieldsParam(false))
```
p-kozlowski: does this mean the URL was wrong?

PR #509, opened 2017-08-07
[ESI-13437] Global store must be synchronized

przemyslaw-dobrowolski-cl:
Lack of sychronization of global store caused panics (concurrent map writes)
when parallel JS tests were running. This change adds a missing mutex.

### DIFF
--- a/extension/otto/global_store.go
+++ b/extension/otto/global_store.go
@@ -15,9 +15,12 @@
 
 package otto
 
+import "sync"
+
 // GlobalStore for values living longer that a single extension environment
 type GlobalStore struct {
 	objects map[string]map[string]interface{}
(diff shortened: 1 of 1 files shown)


extension/otto/global_store.go
```diff
@@ -15,9 +15,12 @@
 
 package otto
 
+import "sync"
+
 // GlobalStore for values living longer that a single extension environment
 type GlobalStore struct {
 	objects map[string]map[string]interface{}
+	mtx sync.Mutex
```
p-kozlowski: go fmt

przemyslaw-dobrowolski-cl: fixed

PR #511, opened 2017-08-08
ESI-13857 cleared golint and go vet errors

jkmar:
Cleared most of the golint and go vet errors.
Remaining errors:
server/state_watcher.go: 100: the watchCancel function is not used on all paths (possible context leak)
server/state_watcher.go:103: this return statement may be reached without using the watchCancel var defined on line 100
server/sync_watcher.go:196: the previousCancel function is not used on all paths (possible context leak)
server/sync_watcher.go:208: this return statement may be reached without using the previousCancel var defined on line 196
sync/etcdv3/etcd.go:50: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -331,6 +331,10 @@ func getResyncCommand() cli.Command {
 			}
 
 			dbConn, err := db.CreateFromConfig(config)
+			if err != nil {
+				log.Fatalf("Failed to create db conn, err: %s", err)
+			}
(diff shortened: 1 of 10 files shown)


jkmar:
@nati @morrisson @marcin-ptaszynski @przemyslaw-dobrowolski-cl 
Please review.

PR #512, opened 2017-08-08
Improve gohan generate command

nati:
- improve template command for supporting split-by-resource
- support template configuraion file, so that gohan generate command may generate multiple files

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -464,11 +464,13 @@ USAGE:
    command generate [command options] [arguments...]
 
 OPTIONS:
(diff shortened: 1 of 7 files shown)


docs/cli.md
```diff
@@ -458,17 +458,20 @@ You can also change generated code providing custom template.
 
 ``` shell
 NAME:
-   generate - Generate ServerSide Code
+   generate - Genera
```
michalzurawski: new line in the middle of a word

nati: done

PR #521, opened 2017-09-06
CI for crosscompile fix

morrisson:
crosscompile.sh execution failed due to syntax error for windows with
following error message.

extension/goext/schemas.go:97: syntax error: unexpected = in type
declaration

Curiously, compile error did not happen for go1.9 on linux. However,
apparently this declaration is weird, so that should be better to fix
this line.

### DIFF
--- a/extension/goext/schemas.go
+++ b/extension/goext/schemas.go
@@ -94,7 +94,7 @@ func (ctx Context) Clone() Context {
 }
 
 // Priority represents handler priority; can be negative
-type Priority = int
+type Priority int
 
 // PriorityDefault is a default handler priority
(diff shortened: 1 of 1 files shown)


morrisson:
Oh, maybe it's not proper fix.

marcin-ptaszynski:
@morrisson ,
It's new syntax extension https://groups.google.com/forum/#!topic/golang-dev/LvvXAgJvFV0

morrisson:
The problem itself looks like xgo now does not support go 1.9.

morrisson:
Then, so far I think it's ok to wait until xgo supports 1.9.

PR #524, opened 2017-09-07
generate templates to support passing in package, language

skizhak:
- package and language uses 'gen' and 'go' by default. you can override them in the template yaml passed in.
- removed using template option under generate command. use gohan template option directly

### DIFF
--- a/cli/generate.go
+++ b/cli/generate.go
@@ -18,8 +18,10 @@ import (
 //TemplateDef template def defines file
 type TemplateDef struct {
 	Type         string `yaml:"type"`          // All, ResourceGroup, Resource
+	Package      string `yaml:"package"`       // Code generation directory. Default "gen".
 	OutputPath   string `yaml:"output_path"`   // This could be template string. "dir/{{Resource.ID}}".go
 	TemplatePath string `yaml:"template_path"` // Path to template
+	Language     string `yaml:"language"`      // Type of programming language. Eg: go(Default), typescript
 }
 
 func getGenerateCommand() cli.Command {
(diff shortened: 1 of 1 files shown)


cli/generate.go
```diff
@@ -28,9 +30,9 @@ func getGenerateCommand() cli.Command {
 		ShortName: "gen",
 		Usage:     "Generate ServerSide Code",
 		Flags: []cli.Flag{
-			cli.StringFlag{Name: "template, t", Value: "", Usage: "Application template path"},
 			cli.StringFlag{Name: "templates", Value: "", Usage: "Template Configuraion"},
```
nati: could you remove template option?

cli/generate.go
```diff
@@ -178,17 +185,11 @@ func gohanGenerate(c *cli.Context) {
...
-				"goimports -w %s/*.go", codeDir))
-	}
-	if template != "" {
-		execCommand(
-			fmt.Sprintf(
-				"gohan template --config-file %s --template %s | grep -v '^\\s*$' > %s/base_controller.go", configFile, template, packageName))
-		execCommand(
-			fmt.Sprintf(
-				"goimports -w %s/base_controller.go", codeDir))
+		if language == "go" {
```
nati: this should be in the loop.

skizhak: yes, updated.

cli/generate.go
skizhak: template is already removed..

nati: oh ok thx

PR #526, opened 2017-09-15
Environment of goplugin for each path properly

morrisson:
Current goplugin extension could not properly handle path. Therefore,
extensions which should not be loaded for a certain path were loaded.
Especially, this bug occurs critical issues for "notification" events
such as cron and sync watch events. Even if notification event handlers
are registered for a different path, all handlers were called on one
notification event.


przemyslaw-dobrowolski-cl:
@morrisson this PR is generally correct but it has already been integrated and extensed in PR: https://github.com/cloudwan/gohan/pull/527 counld you review this PR?

morrisson:
#527 is also taking care of this issue.

PR #527, opened 2017-09-15
Golang framework improvements

przemyslaw-dobrowolski-cl:
Golang framework improvements:
 - added LockFetch LockList
 - added multi-suite aggregating ginkgo reporter
 - added Http to golang exts
 - added Config to golang tests
 - added load binaries from schemas automatically during go tests
 - added NewError constructor
 - added informational message about missing raw types
 - added db delete and create
 - added test for handlers not invoked
 - added IAuth
 - added automatic transaction retry in goext/Within
 - fixed Within implementation
 - fixed error code handling
 - fixed test isolation in golang tests
 - fixed a bug in environment clone: it should correctly clone bound implementations
 - fixed a bug with missing Start of Environment
 - fixed "resource not found" to be strongly typed
 - fixed invalid clone of runtime types in clone (deepcopy does not work for reflect.Type)
 - remove global variables from golang env
 - remove IEnvironmentRef since is breaks environment mocking

### DIFF
--- a/extension/framework/goplugin/runner/reporter.go
+++ b/extension/framework/goplugin/runner/reporter.go
@@ -1,0 +1,228 @@
+// Copyright (C) 2017 NTT Innovation Institute, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package runner
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/mohae/deepcopy"
+	"github.com/onsi/ginkgo/config"
+	"github.com/onsi/ginkgo/reporters/stenographer"
+	"github.com/onsi/ginkgo/types"
+)
+
+const defaultStyle = "\x1b[0m"
+const boldStyle = "\x1b[1m"
+const redColor = "\x1b[91m"
+const greenColor = "\x1b[32m"
+const yellowColor = "\x1b[33m"
+const cyanColor = "\x1b[36m"
+const grayColor = "\x1b[90m"
+const lightGrayColor = "\x1b[37m"
+
+// Reporter is a custom test reporter which handles results from multiple test suites
+type Reporter struct {
+	suites []types.SuiteSummary
+	specs  []types.SpecSummary
+}
+
+const (
+	configSuccinct          = false
+	configFullTrace         = true
+	configNoisyPendings     = false
+	configSlowSpecThreshold = float64(0.5) // sec
+)
+
+// SpecSuiteWillBegin informs that suite will begin
+func (reporter *Reporter) SpecSuiteWillBegin(config config.GinkgoConfigType, summary *types.SuiteSummary) {
+}
+
+// BeforeSuiteDidRun informs that before suite did run
+func (reporter *Reporter) BeforeSuiteDidRun(setupSummary *types.SetupSummary) {
+}
+
+// SpecWillRun informs that spec will run
+func (reporter *Reporter) SpecWillRun(specSummary *types.SpecSummary) {
+}
+
+// SpecDidComplete informs that spec did complete
+func (reporter *Reporter) SpecDidComplete(specSummary *types.SpecSummary) {
+	reporter.specs = append(reporter.specs, deepcopy.Copy(*specSummary).(types.SpecSummary))
+}
+
+// AfterSuiteDidRun informs that after suite did run
+func (reporter *Reporter) AfterSuiteDidRun(setupSummary *types.SetupSummary) {
+}
+
+// SpecSuiteDidEnd informs that spec suite sis end
+func (reporter *Reporter) SpecSuiteDidEnd(summary *types.SuiteSummary) {
+	for i := range reporter.suites {
+		if reporter.suites[i].SuiteDescription == summary.SuiteDescription {
+			reporter.suites[i] = deepcopy.Copy(*summary).(types.SuiteSummary)
+			break
+		}
+	}
+}
+
+// Prepare prepares a next test suite to run and zeroes its results
+func (reporter *Reporter) Prepare(description string) {
+	reporter.suites = append(reporter.suites, types.SuiteSummary{
+		SuiteDescription: description,
+		SuiteSucceeded:   true,
+		SuiteID:          "undefined",
+		NumberOfSpecsBeforeParallelization: 0,
+		NumberOfTotalSpecs:                 0,
+		NumberOfSpecsThatWillBeRun:         0,
+		NumberOfPendingSpecs:               0,
+		NumberOfSkippedSpecs:               0,
+		NumberOfPassedSpecs:                0,
+		NumberOfFailedSpecs:                0,
+		RunTime:                            time.Duration(0),
+	})
+}
+
+// Report prints the final report
+func (reporter *Reporter) Report() {
+	fmt.Println("--------------------------------------------------------------------------------")
+
+	fmt.Println("Failures:")
+	fmt.Println()
+
+	steno := stenographer.New(true)
+
+	for _, spec := range reporter.specs {
+		steno.AnnounceCapturedOutput(spec.CapturedOutput)
+
+		switch spec.State {
+		case types.SpecStatePassed:
+			if spec.IsMeasurement {
+				steno.AnnounceSuccesfulMeasurement(&spec, configSuccinct)
+			} else if spec.RunTime.Seconds() >= configSlowSpecThreshold {
+				steno.AnnounceSuccesfulSlowSpec(&spec, configSuccinct)
+			} else {
+				steno.AnnounceSuccesfulSpec(&spec)
+			}
+
+		case types.SpecStatePending:
+			steno.AnnouncePendingSpec(&spec, configNoisyPendings && !configSuccinct)
+		case types.SpecStateSkipped:
+			steno.AnnounceSkippedSpec(&spec, configSuccinct, configFullTrace)
+		case types.SpecStateTimedOut:
+			steno.AnnounceSpecTimedOut(&spec, configSuccinct, configFullTrace)
+		case types.SpecStatePanicked:
+			steno.AnnounceSpecPanicked(&spec, configSuccinct, configFullTrace)
+		case types.SpecStateFailed:
+			steno.AnnounceSpecFailed(&spec, configSuccinct, configFullTrace)
+		}
+	}
+
+	fmt.Println()
+	fmt.Println("Report:")
+	fmt.Println()
+
+	// total
+	totalNumberOfTotalSpecs := 0
+	totalNumberOfPassedSpecs := 0
+	totalNumberOfFailedSpecs := 0
+	totalNumberOfSkippedSpecs := 0
+	totalNumberOfPendingSpecs := 0
+	totalRunTime := time.Duration(0) * time.Nanosecond
+
+	// suites
+	for index, suite := range reporter.suites {
+		fmt.Printf("[%4d] ", index+1)
+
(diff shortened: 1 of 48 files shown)


extension/framework/goplugin/runner/reporter.go
```diff
@@ -0,0 +1,195 @@
...
+	totalNumberOfFailedSpecs := 0
+	totalNumberOfSkippedSpecs := 0
+	totalNumberOfPendingSpecs := 0
+	totalRunTime := time.Duration(0) * time.Nanosecond
+
+	// suites
+	for index, suite := range reporter.suites {
+		fmt.Printf("[%4d] ", index+1)
+		if suite.NumberOfFailedSpecs == 0 {
+			fmt.Printf(greenColor+"%-80s"+defaultStyle, suite.SuiteDescription)
```
p-kozlowski: width of fields is too generous: even though I use widescreen terminal, I'm getting wrapped lines. 60-70 chars for test binary + 5-6 chars for number of test cases failures/successes should be fine.

extension/framework/goplugin/runner/runner.go
```diff
@@ -19,208 +19,237 @@ import (
...
 }
 
-// GoTestSuite is a test suite state for a go (plugin) extension runner
-type GoTestSuite struct {
-	plugin  *plugin.Plugin
-	db      db.DB
-	env     *goplugin.Environment
-	path    string
-	manager *schema.Manager
+func readBinaries(p *plugin.Plugin) ([]string, error) {
```
p-kozlowski: currently the test suites are bloated with a long list of schemas followed by a long list of binaries. is it possible to load the binaries specified in schemas instead providing a separate list?

extension/goplugin/environment.go
```diff
@@ -548,122 +589,125 @@ func (thisEnvironment *Environment) resourceFromContext(sch Schema, context map[
...
-		schemasFnRaw: thisEnvironment.schemasFnRaw,
-		schemasFn:    thisEnvironment.schemasFn,
-		schemas:      thisEnvironment.schemas,
-
-		initFnRaw: thisEnvironment.initFnRaw,
-		initFns:   thisEnvironment.initFns,
+		name:  env.name,
+		db:    env.db,
+		ident: env.ident,
+		sync:  env.sync,
```
morrisson: You must clone handlers and schemaHandlers.

przemyslaw-dobrowolski-cl: right, fixed

extension/goplugin/environment.go
```diff
@@ -65,234 +64,233 @@ type SchemaPrioritizedSchemaHandlers map[string]PrioritizedSchemaHandlers
...
-		// internals
-		thisEnvironment.name = existingEnvironment.name
-		thisEnvironment.db = existingEnvironment.db
-		thisEnvironment.ident = existingEnvironment.ident
-		thisEnvironment.sync = existingEnvironment.sync
-
-		// plugin related
-		thisEnvironment.manager = existingEnvironment.manager
-		thisEnvironment.plugin = existingEnvironment.plugin
+func (env *Environment) Load(binary string) (bool, error) {
```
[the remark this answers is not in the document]
przemyslaw-dobrowolski-cl: ok, changed

extension/goplugin/environment.go
```diff
@@ -65,234 +65,233 @@ type SchemaPrioritizedSchemaHandlers map[string]PrioritizedSchemaHandlers
...
+		return nil
 	}
 
-	log.Debug("Starting golang environment: %s", thisEnvironment.source)
+	log.Debug("Starting go environment: %s", env.name)
+
+	// bind
+	env.bindCore()
+	env.bindLogger()
+	env.bindSchemas()
```
morrisson: It looks there is no need to bind here because a constructor did the same thing.

server/resources/resource_management.go
```diff
@@ -518,9 +518,17 @@ func CreateResource(
 	}
 
 	//Validation
-	err = resourceSchema.ValidateOnCreate(dataMap)
-	if err != nil {
-		return ResourceError{err, fmt.Sprintf("Validation error: %s", err), WrongData}
+
+	if _, ok := context["go_validation"]; ok {
```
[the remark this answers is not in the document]
morrisson: @przemyslaw-dobrowolski-cl Could you answer this question?

morrisson:
Different PR is ok, but there are no `DbCreateRaw` and `DbDeleteRaw` functions in `extension/goplugin/schemas.go`.

extension/goplugin/environment.go
```diff
@@ -548,122 +590,134 @@ func (thisEnvironment *Environment) resourceFromContext(sch Schema, context map[
...
-		traceID: uuid.NewV4().String(),
+		rawTypes: env.rawTypes,
+		types:    env.types,
 	}
-	env.bind()
-
-	return env
+	clone.bindCore()
+	clone.bindLogger()
+	clone.bindSchemas()
```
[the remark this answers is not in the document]
przemyslaw-dobrowolski-cl: These are intentionally omitted. They must be set from outside. Database and Sync are bound during Start() in beforeStartHook() for each test-case in ginkgo to clear the environment (AfterEach -> env.Reset()). This ensures complete isolation between tests:
https://github.com/cloudwan/gohan/pull/527/files#diff-66a79b94ba7bb52f904be9e5d856f341R188
and
https://github.com/cloudwan/gohan/pull/527/files#diff-66a79b94ba7bb52f904be9e5d856f341R196

server/resources/resource_management.go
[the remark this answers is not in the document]
przemyslaw-dobrowolski-cl: Validation of struct-based resources in golang must be different from map-based resources, as in JS. This comes from the fact, that in a structure in go you are not able to omit some fields so that they get the default value. In JS this can be done since default fields are simply omitted. I consider this change as not-a-perfect solution but it does the job.
One may think of having structures will all fields being optional but that makes the code messy and currently no better solution is known fot this problem.

extension/framework/goplugin/runner/runner.go
przemyslaw-dobrowolski-cl: It is not possible in all cases. There are schemas which contain more than one plugin and therefore the test would load all of the plugins from a specific schema. This is not the desired behavior. Unit-tests can focus on testing only a single plugin and for that case it shouldn't be needed to create a special schema just to load this single plugin.

extension/goplugin/environment.go
przemyslaw-dobrowolski-cl: Right, there is a redundancy. In fact, there shouldn't be any binding in constructor and in clone. I removed those calls. Core, logger and Schemas is only available once environment is started and not yet stopped.

przemyslaw-dobrowolski-cl:
@morrisson @p-kozlowski I fixed all the comments.

michalzurawski:
@morrisson `DbCreateRaw` and `DbDeleteRaw` have been added.

morrisson:
@michalzurawski Thanks. Looks good.

extension/goplugin/environment.go
```diff
@@ -65,258 +65,247 @@ type SchemaPrioritizedSchemaHandlers map[string]PrioritizedSchemaHandlers
...
-			}
-			if loaded {
-				if err = thisEnvironment.Start(); err != nil {
-					return err
-				}
+			if err := env.Load(url); err != nil {
+				return fmt.Errorf("failed to load binary: %s", err)
 			}
 		}
 	}
```
morrisson: following lines are missing
```
if err := env.Start(); err != nil {
    return err
}
```

Therefore, no goplugin extensions are called now.

przemyslaw-dobrowolski-cl: done

extension/goplugin/environment.go
```diff
@@ -548,122 +580,131 @@ func (thisEnvironment *Environment) resourceFromContext(sch Schema, context map[
...
 
-		initFnRaw: thisEnvironment.initFnRaw,
-		initFns:   thisEnvironment.initFns,
+		handlers:       deepcopy.Copy(env.handlers).(EventPrioritizedHandlers),
+		schemaHandlers: deepcopy.Copy(env.schemaHandlers).(EventSchemaPrioritizedSchemaHandlers),
 
-		traceID: uuid.NewV4().String(),
+		rawTypes: env.rawTypes,
+		types:    env.types,
 	}
```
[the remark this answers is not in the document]
przemyslaw-dobrowolski-cl: There was a problem with clone implementation. It incorrectly set some fields to nil whereas these fields should be cloned iff they have non-nil value.
Fixed.

przemyslaw-dobrowolski-cl:
@morrisson I quashed and rebased the PR. I cleared linter warnings as well.
@marcin-ptaszynski @yudai @nati Could you review this PR?

extension/goplugin/auth.go
```diff
@@ -0,0 +1,45 @@
...
+// Auth is an implementation of IAuth
+type Auth struct{}
+
+// IsAdmin returns if context has admin priviledges
+func (auth *Auth) IsAdmin(context goext.Context) bool {
+	if roleRaw, ok := context["role"]; ok {
+		if role, ok := roleRaw.(*schema.Role); ok {
+			if authRaw, ok := context["auth"]; ok {
+				if auth, ok := authRaw.(schema.Authorization); ok {
+					return role.Match("admin") && auth.TenantName() == "admin"
```
morrisson: As an gohan policy, admin privilege is provided if the role name is "admin". See: https://github.com/cloudwan/gohan/blob/master/etc/schema/gohan.json#L7

Any tenant name is ok.

extension/goplugin/environment.go
```diff
@@ -65,258 +65,269 @@ type SchemaPrioritizedSchemaHandlers map[string]PrioritizedSchemaHandlers
...
+	env := &Environment{
+		initFns:         map[string]func(goext.IEnvironment) error{},
+		beforeStartHook: beforeStartHook,
+		afterStopHook:   afterStopHook,
+
 		name:  name,
-		db:    dataStore,
-		ident: ident,
-		sync:  sync,
+		ident: nil,
```
morrisson: How will `ident` be used? It looks `ident` is not used anywhere.

extension/goplugin/environment.go
```diff
@@ -65,258 +65,269 @@ type SchemaPrioritizedSchemaHandlers map[string]PrioritizedSchemaHandlers
...
+
+// HTTP returns an implementation to IHTTP interface
+func (env *Environment) HTTP() goext.IHTTP {
+	return &HTTP{}
+}
+
+// Auth returns an implementation to IAuth interface
+func (env *Environment) Auth() goext.IAuth {
+	return &Auth{}
 }
```
morrisson: `HTTP` and `Auth` do not have any relations to env. I am not sure if env should have these functions here. (just a comment. There is no need to change if you think it's unnecessary).

extension/goplugin/auth.go
przemyslaw-dobrowolski-cl: I just copied this code from auth.js:
```
        return role && auth && role.Match("admin") && (auth.TenantName() == "admin")
```
so it means that both JS and Go implementations are incorrect?

extension/goplugin/environment.go
przemyslaw-dobrowolski-cl: Currently, HTTP does not require anything from Env but that may change at some moment in future - for example it may require some configuration from config file or authorization credentials or similar. Generally, the assumption is that whatever a plugin needs from gohan it should be available in Env. That gives a clear isolation by interfaces and in fact, guarantees that plugin compilation work correctly. I'd rather leave this as is.

extension/goplugin/auth.go
morrisson: `auth.js` is our private javascript extension. That means our private function isAdminContext was incorrect...

Actually, I wondered in the previous review whether this `isAdmin` function should be placed in public gohan tree or not. However, in my current opinion, this function itself is ok because public `gohan.json` has a special statement about admin role.

extension/goplugin/environment.go
morrisson: got it. it seems fine.

extension/goplugin/auth.go
p-kozlowski: I've changed the code to only expose generic functions. Custom logic will be moved to extensions.

morrisson:
Thank you for your patience @przemyslaw-dobrowolski-cl , @p-kozlowski

PR #528, opened 2017-09-19
Mark a test that was never run and fails as pending

p-kozlowski:
This test was never run on the CI (example: https://circleci.com/gh/cloudwan/gohan/1171?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link, look for ok  	github.com/cloudwan/gohan/extension/golang	1.023s	coverage: 0.0% of statements [no tests to run]" in "make test"). When I've enabled it, it always fails.

For now, let's run the suite, but mark the test case as "pending".

Reviewers, what's your opinion about it? Should we fix this test or is it safe to delete?


przemyslaw-dobrowolski-cl:
If this test may be trivially fixed, then it should be fixed in this PR. Otherwise it should be dropped because it is useless.

przemyslaw-dobrowolski-cl:
@p-kozlowski Lets add Pit for this test, similarly to 3 != 4 test.

p-kozlowski:
reviewers, gentle reminder about this PR. how do you feel about this change?

przemyslaw-dobrowolski-cl:
@morrisson @marcin-ptaszynski @zimnx please review

PR #529, opened 2017-09-21
Add resource to context in GetSingleResource

p-kozlowski:
Resource is already added in Create/Update/Delete but not in Get

### DIFF
--- a/docs/extension.md
+++ b/docs/extension.md
@@ -91,6 +91,7 @@ Comparison of different types of extensions:
   Same as post_show but executed in the DB transaction context.transaction contains transaction object for DB operation
 
   id : request id
+  context.resource contains the resource associated with the event
   context.response contains response data.
   You can also update response here
 
@@ -99,6 +100,7 @@ Comparison of different types of extensions:
   show event after DB operation
 
   id : request id
+  context.resource contains the resource associated with the event
   context.response contains response data.
   You can also update response here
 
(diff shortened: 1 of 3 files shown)


morrisson:
As far as I understand, `context["resource"]` implies a resource manipulated by requests. In this sense, it seems reasonable and consistent to me that a context in read operations ("show" and "list") does not have a `resource` key. What do you think? @nati

p-kozlowski:
@morrisson @nati I agree that show/list are not supposed to modify the underlying resource, but lack of "resource" field in the context makes writing generic extension code much harder. pre/post_create/update/delete_tx handlers are allowed to read the resource through context["resource"], but show/list handlers don't. As a result, functions operating on context must have to versions: one for all handlers and one for show/list. In my opinion it's just an API inconsistency.

morrisson:
@p-kozlowski if you believe this PR is necessary, please update this PR not only for post_show. And please update doc `docs/extension.md`.

You said that context["resource"] is required for "just an API inconsistency", but I think it is difficult to have this consistency with a reasonable way and reason. It looks good for post_show at some extent, but how about post_list? Having a resource slice as a context["resource"] value is really consistent with other interface? In addition, how about pre_show and pre_list? It really does not make sense to pre-fetch a resource just only for filling context["resource"].

p-kozlowski:
@morrisson The context will be passed to both post_show_in_tx and post_show. As you mentioned, it does not make sense to do another fetch for pre_show(_in_tx), so let's keep those as-is. 

In case of post_list(_in_tx) there are no failing use cases, so in my opinion it's safe to leave it for know. If any error cases appear, how about addressing them later?

Also, I'll update the docs.

p-kozlowski:
reviewers, gentle reminder about this PR. what's you opinion? are we good to merge it?

przemyslaw-dobrowolski-cl:
@morrisson A kind remainder - this PR is still waiting for your review.

PR #530, opened 2017-09-27
[ESI-14112] Timeouts for golang extensions

p-kozlowski:
- pass cancelable context to golang handlers
- cancel the context on extension execution timeout
- pass context to sync.Watch


extension/goplugin/sync.go
```diff
@@ -79,26 +81,25 @@ func (sync *Sync) Delete(path string, prefix bool) error {
 }
 
 // Watch watches a single path in sync
-func (sync *Sync) Watch(path string, timeout time.Duration, revision int64) ([]*goext.Event, error) {
-	eventChan := make(chan *gohan_sync.Event, 32)
+func (sync *Sync) Watch(ctx context.Context, path string, timeout time.Duration, revision int64) ([]*goext.Event, error) {
 	stopChan := make(chan bool, 1)
 	defer close(stopChan)
```
morrisson: stopChan is not used anywhere.

extension/goplugin/environment.go
```diff
@@ -387,6 +408,56 @@ func (env *Environment) HandleEvent(event string, context map[string]interface{}
...
+
+func buildCancel(env *Environment, event string) (ctx context.Context, cancel context.CancelFunc) {
+	selectedTimeLimit := env.timeLimit
+	for _, timeLimit := range env.timeLimits {
+		if timeLimit.Match(event) {
+			selectedTimeLimit = timeLimit.TimeDuration
+			break
+		}
+	}
+	if selectedTimeLimit != 0 {
```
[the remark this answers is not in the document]
p-kozlowski: ok, fixed

extension/goplugin/sync.go
p-kozlowski: same thing for errorChan, both removed

p-kozlowski:
@morrisson the idea is a little bit different here. When an event is handled, two things may happen: 

1. API client disconnects
2. Extension takes too much time to complete

This patch addresses these two situations using Golangs standard context.Context. The policy here is too allow "graceful shutdown" - ie. extensions should handle timeout/client disconnect themselves. Situation in Golang is different that in JS vm - we can't simply interrupt processing. If (for example) we'll start event handling in a new goroutine and return 503 after timeout, we will have to deal with "dangling processing". The running goroutine will likely lead to logic errors and unnecessary resource consumption. If we rely on "cooperative termination" reasoning about the system will be much easier and handling an interrupted request will happen after an arbitrary, but finite amount of time.

Currently a known long-running method is sync.Watch. I've extended the interface in goext.ISync, to take advantage of the [cancelable version](https://github.com/cloudwan/gohan/blob/master/sync/sync.go#L44). Another good candidates are DB methods - internally they already accept a context.Context (internally, all db interactions are done using [Transaction.Exec](https://github.com/cloudwan/gohan/blob/master/db/sql/sql.go#L561)/[DB.Exec](https://github.com/golang/go/blob/master/src/database/sql/sql.go#L1263). I'll extend the transaction interface to also accept context.Context.

The last point of consideration are errors in plugin logic (eg. infitite loop). In my opinion, Gohan should not try to prohibit someone shooting in their foot. I'm strongly for the cooperative model here.

morrisson:
@p-kozlowski If extension developer did not write timeout handling logic, what happens if time limit exceeds?

p-kozlowski:
@morrisson with this current patch, if the plugin is somehow "active" - it performs operations on the DB, ISync or IHttp it will receive an error ("context deadline exceeded"). The common-sense error handler (if err != nil { return err }) will be just enough to stop further processing.

My strong opinion is that we should not make the framework overcomplicated for the common case because some plugins somewhere may be poorly written. It's the developers responsibility to do their job properly. If a developer does not care for such simple things as error handling, he will likely don't care for timeouts as well.

morrisson:
@p-kozlowski Thanks.

Btw, it looks current patch is very different from what I saw first, right? Current patch shows really clear what it does.

+1, but please squash commits. And it looks there is a bit relevant to #538 discussion so we should wait merge until #538 is done.

PR #531, opened 2017-09-27
Iterface segregation, type-safe casting of context[transaction]

p-kozlowski:
This PR consists of two parts:
1. refactoring: extracting Contig and UUID from Core into separate interfaces (to facilitate mocking in unit tests)
2. proper casting of context[transaction]

It may happen that context[transaction] is a transactionCommitInformer, which does not fulfill goext.ITransaction.

It must be a part of public (goext) interface because Within() relies on this functionality. It may be interesting to implement Within() using dependency injection, but that complicates testing of plugins when mocks are used.

### DIFF
--- a/extension/goext/util.go
+++ b/extension/goext/util.go
@@ -1,5 +1,6 @@
 package goext
 
 type IUtil interface {
+	NewUUID() string
 	GetTransaction(context Context) (ITransaction, bool)
 }
--- a/extension/goext/util.go
+++ b/extension/goext/util.go
@@ -1,0 +1,5 @@
+package goext
+
+type IUtil interface {
+	GetTransaction(context Context) (ITransaction, bool)
+}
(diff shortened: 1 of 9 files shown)


extension/goext/util.go
```diff
@@ -0,0 +1,6 @@
+package goext
```
morrisson: license

extension/goplugin/util.go
```diff
@@ -0,0 +1,41 @@
+package goplugin
```
morrisson: license

extension/goext/util.go
p-kozlowski: added

extension/goplugin/util.go
p-kozlowski: added

p-kozlowski:
@morrisson importing parts of Gohan (except goext) in plugins will work for same cases (like here), but in general we'd like to use Dependency Insertion and Dependency Injection for a number of reasons:

1. resulting binary size increases
2. you may encounter import cycles
3. compilation of plugins may fail

Therefor, I'd like to keep consistency and follow the rule for all interfaces.

PR #532, opened 2017-09-27
Fixed trigger event

p-kozlowski:
Events should be triggered only on environments registered for given schema, not on all environments


morrisson:
It works fine in my local env.

PR #533, opened 2017-09-27
Exposed info about schema relations

zimnx:
goext.Schemas exposes information about schema relations.


morrisson:
@zimnx, @p-kozlowski What is the use case of this `Relations()` method in goplugin extension?

morrisson:
+1, but please rebase and squash commits.

zimnx:
Rebased and squashed.

PR #537, opened 2017-09-28
Each instance of logger has different TraceID

zimnx:
Because logger Clone literally cloned object, same environment was always passed to logger. This resulted that each logger had same traceID. Now Clone generates new traceID for each instance.

### DIFF
--- a/extension/goplugin/core.go
+++ b/extension/goplugin/core.go
@@ -64,11 +64,11 @@ func NewCore(env *Environment) *Core {
 }
 
 // Clone allocates a clone of Core; object may be nil
-func (core *Core) Clone() *Core {
+func (core *Core) Clone(env *Environment) *Core {
 	if core == nil {
 		return nil
 	}
 	return &Core{
-		env: core.env,
+		env: env,
 	}
 }
--- a/extension/goplugin/environment.go
+++ b/extension/goplugin/environment.go
@@ -64,10 +64,6 @@ type SchemaPrioritizedSchemaHandlers map[string]PrioritizedSchemaHandlers
 // EventSchemaPrioritizedSchemaHandlers is a per-event per-schema prioritized list of schema handlers
 type EventSchemaPrioritizedSchemaHandlers map[string]SchemaPrioritizedSchemaHandlers
 
-func newTraceID() string {
-	return uuid.NewV4().String()
-}
-
 // Environment golang based environment for gohan extensions
 type Environment struct {
 	initFns         map[string]func(goext.IEnvironment) error
@@ -250,7 +246,6 @@ func (env *Environment) Start() error {
 		log.Debug("Before start init is not set for go environment: %s", env.name)
 	}
 
-	// generate trace ID
 	env.traceID = newTraceID()
 
 	// init extensions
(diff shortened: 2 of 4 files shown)


morrisson:
Thank you for bringing up this issue. This patch can resolve this logger issue, but I think this issue is occurred by a `Environment.Clone()` logic problem. Here is a Clone method in `extension/goplugin/environment.go`.

```
// Clone makes a clone of the rawEnvironment
func (env *Environment) Clone() extension.Environment {
	clone := &Environment{
		initFns:         env.initFns,
		beforeStartHook: env.beforeStartHook,
		afterStopHook:   env.afterStopHook,

		coreImpl:     env.coreImpl.Clone(),
		loggerImpl:   env.loggerImpl.Clone(),
		schemasImpl:  env.schemasImpl.Clone(),
		syncImpl:     env.syncImpl.Clone(),
		databaseImpl: env.databaseImpl.Clone(),

		name: env.name,

		traceID: newTraceID(),

		handlers:       deepcopy.Copy(env.handlers).(EventPrioritizedHandlers),
		schemaHandlers: deepcopy.Copy(env.schemaHandlers).(EventSchemaPrioritizedSchemaHandlers),

		rawTypes: make(map[string]reflect.Type),
		types:    make(map[string]reflect.Type),
	}
	for k, v := range env.rawTypes {
		clone.rawTypes[k] = v
	}
	for k, v := range env.types {
		clone.types[k] = v
	}
	return clone
}
```

Each core, logger, schemas, sync and database type needs to have a relation with parent env. However, each type has a Clone logic like this

```
// Clone allocates a clone of Core; object may be nil
func (core *Core) Clone() *Core {
	if core == nil {
		return nil
	}
	return &Core{
		env: core.env,
	}
}
```

That is, if `e2` is cloned from `e1`,  `e2.coreImpl.env` is `e1`. This is the root problem behind this issue.

Therefore, proper fix should be changing the logic of `Environment.Clone` as follows:

```
func (env *Environment) Clone() extension.Environment {
	clone := &Environment{
		initFns:         env.initFns,
		beforeStartHook: env.beforeStartHook,
		afterStopHook:   env.afterStopHook,

		name: env.name,

		traceID: newTraceID(),

		handlers:       deepcopy.Copy(env.handlers).(EventPrioritizedHandlers),
		schemaHandlers: deepcopy.Copy(env.schemaHandlers).(EventSchemaPrioritizedSchemaHandlers),

		rawTypes: make(map[string]reflect.Type),
		types:    make(map[string]reflect.Type),
	}

	clone.SetCore(env.coreImpl.Clone(clone))
	...

	return clone
}
```

zimnx:
@morrisson Thanks for bringing this. I've added new commit. Objects which are coupled with environment now accept env as argument in Clone method.

morrisson:
Thanks for updating. However, this change is a bit different from what I suggests. I think, traceID should be in Environment then Logger should refer parent environment's traceID because traceID is a kind of identifier of one independent process, that is to say, environment. Logger should just provide a logging feature.

zimnx:
@morrisson I've applied your remarks and squashed commits.

PR #538, opened 2017-09-28
Fix triggering external events

p-kozlowski:
- use proper types for schema and transaction
- require schema_id to be present in context
- fix infitine recursion in RawTransaction

### DIFF
--- a/extension/goplugin/core.go
+++ b/extension/goplugin/core.go
@@ -16,6 +16,8 @@
 package goplugin
 
 import (
+	"fmt"
+
 	"github.com/cloudwan/gohan/extension"
 	"github.com/cloudwan/gohan/extension/goext"
 )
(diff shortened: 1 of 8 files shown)


extension/goplugin/core.go
```diff
@@ -55,6 +63,42 @@ func (core *Core) TriggerEvent(event string, context goext.Context) error {
 	return nil
 }
 
+func getSchemaId(context goext.Context) (string, error) {
+	rawSchemaID, ok := context["schema_id"]
```
morrisson: `context["schema"].GetID()` should work, right?

p-kozlowski: If the schema is an ISchema (and it is present in the map), then yes, ```context["schema"].(ISchema).GetID()``` will do. 

Side note: the issue with current goplugin API design is that we're overusing goext.Context (which is map[string]interface{}). We should not make any assumptions what's inside the map, and what are the types of fields (ie. is it goext.ISchema, or *schema.Schema, is it goext.ITransaction or transaction.Transaction or maybe transactionEventLogger). This big issue - make the goext.Context fully typed and perhaps provide different types for incoming, outgoing and internal contexts - is a high priority item on my TODO list. I just want to address it later, not in this commit.

PR #540, opened 2017-09-29
Added marshaling to gohan golang extensions

przemyslaw-dobrowolski-cl:


extension/goplugin/schemas_test.go
```diff
@@ -304,4 +304,62 @@ var _ = Describe("Schemas", func() {
...
+		It("should convert context to resource", func() {
+			context := map[string]interface{}{
+				"id":          "42",
+				"description": "test",
+				"subobject": map[string]interface{}{
+					"subproperty": "testproperty",
+				},
+			}
+			expected := &test.Test{
+				ID:          "42",
```
p-kozlowski: please add more test cases:
1. nil/null instead of subobject (both ways: convertTo/convertFrom)
2. negative cases (like int instead of string, null where it's not allowed)

extension/goplugin/schemas.go
```diff
@@ -123,8 +123,54 @@ func (schema *Schema) ID() string {
 	return schema.raw.ID
 }
 
-// StructToMap converts structure representation of the resource to mapped representation
-func (schema *Schema) StructToMap(resource interface{}) map[string]interface{} {
+// ResourceFromContext converts mapped representation to structure representation of the resource
+func (schema *Schema) ResourceFromContext(context map[string]interface{}) (res goext.Resource, err error) {
+	rawType, ok := schema.env.rawTypes[schema.ID()]
```
p-kozlowski: I don't like the idea of having this as a member of Schema - it could complicate testing (think mocking). Why not make it a free function taking env as an argument? env already exposes RawTypes().

p-kozlowski:
please align these changes with #539

extension/goext/util.go
```diff
@@ -18,4 +18,10 @@ package goext
 type IUtil interface {
 	NewUUID() string
 	GetTransaction(context Context) (ITransaction, bool)
+
+	// ResourceFromContextForType converts mapped representation to structure representation of the resource for given type
+	ResourceFromContextForType(context map[string]interface{}, rawResource interface{}) (Resource, error)
```
przemyslaw-dobrowolski-cl: I don't like context name here since this can be a part of context or even unrelated to any context. I'd prefer a name like ResourceFromMap which gets mappedRes map[string] and the type

extension/goext/util.go
```diff
@@ -18,4 +18,10 @@ package goext
...
+
+	// ResourceToContext converts structure representation of the resource to mapped representation
+	ResourceToContext(resource interface{}) (map[string]interface{}, error)
```
przemyslaw-dobrowolski-cl: I'd prefer name ResourceToMap

morrisson:
+1. Ok to merge after resolving conflicts and squashing commits.

zimnx:
Rebased and squashed

PR #541, opened 2017-10-02
Added DerivedSchemas, ColumnNames and Properties to ISchema

jkmar:


morrisson:
@jkmar What is the usecase for an extension to access (a kind of) super type of resource?

extension/goext/schemas.go
```diff
@@ -156,6 +156,9 @@ type ISchema interface {
 
 	// RegisterRawType registers a raw resource type, containing db annotations
 	RegisterRawType(rawResourceType interface{})
+
+	// Children returns list of schemas that extend schema with given id
+	Children() []ISchema
```
morrisson: In this sense, children is not proper term even in an object-oriented meaning. Should be `SuperSchemas()` or something

extension/goplugin/schemas.go
```diff
@@ -683,6 +683,20 @@ func (schema *Schema) RegisterType(typeValue interface{}) {
 	schema.env.RegisterType(schema.raw.ID, typeValue)
 }
 
+// Children returns list of schemas that extend schema with given id
+func (schema *Schema) Children() []goext.ISchema {
```
marcin-ptaszynski: confusing naming, since parent-child relationship in gohan is already reserved for special foreign key case. Inheritance (using extend) should follow different naming.

extension/goplugin/schemas.go
```diff
@@ -683,6 +683,20 @@ func (schema *Schema) RegisterType(typeValue interface{}) {
...
+	manager := gohan_schema.GetManager()
+	children := []goext.ISchema{}
+	for _, raw := range manager.OrderedSchemas() {
+		for _, parent := range raw.Extends {
+			if parent == schema.ID() {
+				children = append(children, NewSchema(schema.env, raw))
```
p-kozlowski: once you've appended a child to the result list, there's not need to further iterate raw.Extends. add a `brake`

jkmar:
@morrisson @marcin-ptaszynski @p-kozlowski I've changed the name of that function to DerivedSchemas (also added break).

Use case for accessing super schema:
Register a handler for all schemas of given type.

Use case for make columns:
Part of query for raw query (from ITransaction).

extension/goext/schemas.go
```diff
@@ -158,6 +158,22 @@ type ISchema interface {
 	RegisterRawType(rawResourceType interface{})
 
 	RawSchema() interface{}
+
+	// DerivedSchemas returns list of schemas that extend schema with given id
+	DerivedSchemas() []ISchema
+
+	// MakeColumns generates an array that has Gohan style colmun names
+	MakeColumns() []string
```
zimnx: I don't like name "MakeColumns" since this doesn't make any columns. It's a read operation so word 'make' seems wrong. It's not really tied to Schema, so maybe introduce free function, not as interface method.

jkmar: This function is called after the function from the "sql" package with the same name, which uses raw schema, so that's why it's a part of schema interface.

marcin-ptaszynski: How does it relate to Properties?

jkmar: Properties is another function that returns properties of given schema.

zimnx: Fact that function name in `sql` package is wrong, doesn't implies that new code should follow it. I meant that this free function could take schema as parameter or take list Properties to iterate over.

zimnx: Ok, we can't make it a free function. Just change name to something else. 'MakeColumns' seems not right.

jkmar:
@morrisson @marcin-ptaszynski @zimnx I've applied your changes, please review again.

morrisson:
Actually, I still wonder how these methods are used, but it's ok to be added.

przemyslaw-dobrowolski-cl:
@marcin-ptaszynski This PR is still blocked because of one reject. If you are satisfied with the changes, please accept it or elaborate why you are still rejecting this PR.

PR #543, opened 2017-10-04
Fixed missing wrappers for errors returned from event handlers

przemyslaw-dobrowolski-cl:


extension/goplugin/environment.go
```diff
@@ -469,8 +471,8 @@ func handleEventForEnv(env IEnvironment, event string, requestContext map[string
 		for _, priority := range sortHandlers(prioritizedEventHandlers) {
 			for index, eventHandler := range prioritizedEventHandlers[priority] {
 				if err := eventHandler(requestContext, env); err != nil {
-					return fmt.Errorf("failed to dispatch event '%s' at priority '%d' with index '%d': %s",
-						event, priority, index, err)
+					env.Logger().Warningf("failed to handle event '%s' at priority '%d' with index '%d': %s", event, priority, index, err)
+					return goext.NewError(goext.ErrorBadRequest, err)
```
p-kozlowski: err may already be a goext.Error. What happens in that case? I suggest modification of eventHandlers to always return goext.Error instead of error.

przemyslaw-dobrowolski-cl: fixed. Now, an extension can return either untyped error or typed goext.Error with some specific error code. Changing event handler declaration would be a massive backward compatibility break and would require many changes which IMO are not required.

extension/goplugin/environment.go
```diff
@@ -356,17 +356,22 @@ func dispatchSchemaEventForEnv(env IEnvironment, prioritizedSchemaHandlers Prior
...
+		for index, schemaEventHandler := range prioritizedSchemaHandlers[priority] {
 			context["go_validation"] = true
 			if err := schemaEventHandler(context, resource, env); err != nil {
-				return err
+				env.Logger().Warningf("failed to handle schema '%s' event '%s' at priority '%d' with index '%d': %s", sch.ID(), event, priority, index, err)
+				switch err.(type) {
+				case goext.Error:
+					return err
+				default:
+					return goext.NewError(goext.ErrorBadRequest, err)
```
morrisson: I think all unspecified error should be handled as `InternalServerError`

extension/goplugin/environment.go
```diff
@@ -469,8 +476,13 @@ func handleEventForEnv(env IEnvironment, event string, requestContext map[string
...
 			for index, eventHandler := range prioritizedEventHandlers[priority] {
 				if err := eventHandler(requestContext, env); err != nil {
-					return fmt.Errorf("failed to dispatch event '%s' at priority '%d' with index '%d': %s",
-						event, priority, index, err)
+					env.Logger().Warningf("failed to handle event '%s' at priority '%d' with index '%d': %s", event, priority, index, err)
+					switch err.(type) {
+					case goext.Error:
+						return err
+					default:
+						return goext.NewError(goext.ErrorBadRequest, err)
```
morrisson: ditto

extension/goplugin/environment.go
przemyslaw-dobrowolski-cl: After offline discussion we came to a conclusion that the best solution is to enforce schema handlers to return always goext.Error and always define proper error. I changed the code accordingly. What do you think about this idea?

extension/goplugin/environment.go
przemyslaw-dobrowolski-cl: Fixed

extension/goext/core.go
```diff
@@ -20,6 +20,6 @@ type ICore interface {
 	TriggerEvent(event string, context Context) error
 	HandleEvent(event string, context Context) error
 
-	RegisterEventHandler(eventName string, handler func(context Context, environment IEnvironment) error, priority int)
-	RegisterSchemaEventHandler(schemaID string, eventName string, handler func(context Context, resource Resource, environment IEnvironment) error, priority int)
+	RegisterEventHandler(event string, handler func(context Context, environment IEnvironment) *Error, priority int)
```
p-kozlowski: minor: make handler a first-class type - it will simplify the code when handlers are generated dynamically

extension/goext/error.go
```diff
@@ -15,43 +15,127 @@
...
+	Err    error
+	Status int
 }
 
 // NewError returns a new error
-func NewError(code ErrorCode, err error) Error {
-	return Error{
-		Code: code,
-		Err:  err,
+func NewError(status int, err error) *Error {
```
p-kozlowski: if err is an goext.Error I think we should extract the error message without the code. It would be helpful when extensions call 
```
if err := TriggerEvent(...); err != nil { 
    return goext.NewError(..., err) 
}
```

extension/goplugin/environment.go
morrisson: Generally, it's good idea. Actually, "notification" event does not require status_code so please make sure this change works with cron or sync_watch case.

extension/goext/error.go
przemyslaw-dobrowolski-cl: good idea - fixed. Please take a look at error_test.go how the stacking works now.

extension/goext/error_test.go
```diff
@@ -0,0 +1,71 @@
...
+		var (
+			errSingle *goext.Error
+			errDouble *goext.Error
+		)
+		BeforeEach(func() {
+			errSingle = goext.NewError(http.StatusInternalServerError, fmt.Errorf("it does not work"))
+			errSingle.Origin = "first.go:100"
+
+			errDouble = goext.NewError(http.StatusServiceUnavailable, goext.NewError(http.StatusUseProxy, fmt.Errorf("nothing here")))
+			errDouble.Origin = "jeden.go:1000"
```
p-kozlowski: english, please ;)

extension/goext/core.go
p-kozlowski: still not done

przemyslaw-dobrowolski-cl: fixed

extension/goext/error_test.go
przemyslaw-dobrowolski-cl: :)
fixed

przemyslaw-dobrowolski-cl:
@p-kozlowski i applied missing changes - please review again

przemyslaw-dobrowolski-cl:
@marcin-ptaszynski @zimnx a kind remainder about this PR

przemyslaw-dobrowolski-cl:
rebased

extension/goext/error.go
```diff
@@ -15,43 +15,192 @@
...
+	return &Error{
+		Err:    err,
+		Status: status,
+		Origin: captureOrigin(),
+	}
+}
+
+func captureOrigin() string {
+	const SkipStackFrames = 3
```
p-kozlowski: please add a UT for this magic number

extension/goext/error.go
```diff
@@ -15,43 +15,192 @@
...
+		return "<unknown>"
+	}
+	return fmt.Sprintf("%s:%d", file, line)
+}
+
+type ErrorMatcher struct {
+	Expected *Error
+}
+
+func (matcher *ErrorMatcher) Match(actual interface{}) (bool, error) {
```
p-kozlowski: how is the Match function supposed to be used? please add some UTs

extension/goext/error.go
morrisson: Is there any case where we would like to see the full stack other than HandleEvent? My idea is changing this func to `func captureOrigin(skip int) string`, line 36 to `Origin: captureOrigin(3)` then add another constructor `func NewErrorFullStack(status int, err error) *Error`. What do you think?

przemyslaw-dobrowolski-cl: @p-kozlowski added UT
@morrisson I found current implementation to be very helpful when locating invalid error codes in code. In many cases, the same error code is returned from different places. Unless you have a working debugger for go extensions, it takes a lot of time to locate the exact "return Error" line which caused a particular error. In which places you would like to capture stack traces and in which this should be avoided in your opinion?

extension/goext/error.go
przemyslaw-dobrowolski-cl: @p-kozlowski utest added, docs added

extension/goext/error.go
morrisson: Ok. This explanation makes sense to me.

extension/goext/error_test.go
```diff
@@ -0,0 +1,87 @@
...
+		It("Should return full error stack on request", func() {
+			Expect(fmt.Sprintf("%s", errSingle.ErrorStack())).To(Equal("HTTP 500 (Internal Server Error) at first.go:100: it does not work"))
+			Expect(fmt.Sprintf("%s", errDouble.ErrorStack())).To(Equal(
+				`HTTP 503 (Service Unavailable) at one.go:1000 from
+  <- HTTP 305 (Use Proxy) at two.go:2000: nothing here`))
+		})
+
+		It("Should capture correct stack", func() {
+			error := goext.NewErrorInternalServerError(fmt.Errorf("test error"))
+			Expect(error.Origin).To(ContainSubstring("github.com/cloudwan/gohan/extension/goext/error_test.go:"))
```
p-kozlowski: minor: this test may be inaccurate: I've changed `SkipStackFrames` to 14 and the test still passes. what I'm worried is that we may get false reports in the logs. one way to fix the test would be to do something like
```
_, _, fileLine, _ := runtime.Caller(0)
expected = fmt.Sprintf("github.com/cloudwan/gohan/extension/goext/error_test.go:%d", fileLine-1)
Expect(error.Origin).To(Equal(expected))
```

extension/goext/error.go
przemyslaw-dobrowolski-cl: Adding Equal function will cause import of gomega in goext package. I don't think it's a good idea. What do you think about this issue?

extension/goext/error_test.go
przemyslaw-dobrowolski-cl: ok, fixed

extension/goext/error.go
p-kozlowski: I agree that we shouldn't mix test and production code. Perhaps we should move all test-related stuff (like the `ErrorMatcher` type) to a separate package. If someone uses `ErrorMatcher`, it's more than likely they're using gomega as well, so it should not be an issue to import it from the test package. Or would it?

przemyslaw-dobrowolski-cl: Good point, added exported test package for goext

PR #545, opened 2017-10-06
Extended build time information

przemyslaw-dobrowolski-cl:
Extended build time information is embedded into binary and logged during server startup


morrisson:
This patch should be okay. @nati , could you check this?

przemyslaw-dobrowolski-cl:
@marcin-ptaszynski could you review this change?

PR #546, opened 2017-10-10
Do not unneccesarily expose goplugin internals

przemyslaw-dobrowolski-cl:
There is no need to expose raw schema to the outside world. It is used only internally.


extension/goplugin/core.go
```diff
@@ -48,7 +48,7 @@ func (core *Core) TriggerEvent(event string, context goext.Context) error {
 	// If a schema is already set, we should overwrite it with proper type and
 	// restore it once we're done with handling the event
 	defer restoreOriginalSchema(context)()
-	context["schema"] = core.env.Schemas().Find(schemaID).RawSchema()
+	context["schema"] = core.env.Schemas().Find(schemaID).(*Schema).raw
```
p-kozlowski: this change makes mocking ISchema impossible. please do not merge it.

PR #547, opened 2017-10-10
TriggerEvent returns an error on JS exception

p-kozlowski:


extension/goplugin/js_integration_test.go
```diff
@@ -0,0 +1,128 @@
...
+		initGlobalObjects()
+		initGoExtension()
+		initOttoExtension()
+
+		testSchema := env.Schemas().Find("test")
+		Expect(testSchema).NotTo(BeNil())
+	})
+
+	AfterEach(func() {
+		env.Stop()
```
przemyslaw-dobrowolski-cl: manager should be cleared here otherwise, schema overwrite errors may occur

extension/goplugin/js_integration_test.go
```diff
@@ -0,0 +1,128 @@
...
+			Expect(ctx).To(HaveKeyWithValue("js_invoked", true))
+		})
+
+		It("should clean context", func() {
+			Expect(ctx).NotTo(HaveKey("exception"))
+			Expect(ctx).NotTo(HaveKey("exception_message"))
+		})
+
+		It("should propagate error code", func() {
+			Expect(err.(goext.Error).Code).To(Equal(goext.ErrorCode(456)))
```
przemyslaw-dobrowolski-cl: the error sould be *goext.Error since it is a pointer
also, check if message has also been passed correctly

p-kozlowski: goext.NewError returns an goext.Error, not *goext.Error. I'm aware this will change in #543, but that PR is not yet merged.
Depending on which PR is merged sooner, this will require a change, but that change will be guarded by the test, so we won't forget about it.

extension/goplugin/js_integration_test.go
p-kozlowski: ok

extension/goplugin/js_integration_test.go
przemyslaw-dobrowolski-cl: ok

PR #548, opened 2017-10-11
Properly convert resource to map

morrisson:
If struct has array of struct pointers, it was not peroperly converted
to map.


morrisson:
@p-kozlowski Ok, I also think #549 is better. Cc: @zimnx

morrisson:
will be handled in #549

PR #549, opened 2017-10-11
Supporting undefined values in jsons

zimnx:
+ Marshalling tests and fixes to match desired behavior

Still testing if marshal behaves same like in JS extensions, but code is almost ready, ready to review to speedup process.


extension/goext/environment.go
```diff
@@ -193,3 +233,35 @@ func MakeNullFloat(value float64) NullFloat {
 		Valid: true,
 	}
 }
+
+// MakeUndefinedString allocates a new undefined string and sets its value
+func MakeUndefinedString(value string) *NullString {
```
p-kozlowski: naming: "MakeUndefinedX" suggests you're making a value with an undefined value - but the code says that the value is valid.
I suggest making MakeNullX return a pointer and remove the MakeUndefinedX family

extension/goplugin/util.go
```diff
@@ -158,8 +286,39 @@ func (util *Util) ResourceToMap(resource interface{}) map[string]interface{} {
...
+						slice[i] = v.Index(i).Interface()
+					}
+					fieldsMap[field] = slice
+				} else {
+					slice := make([]map[string]interface{}, v.Len())
+					for i := 0; i < v.Len(); i++ {
+						elem := v.Index(i)
+						if anyOfKinds(elem.Kind(), nullableKinds) {
+							if !elem.IsNil() {
+								slice[i] = util.ResourceToMap(elem.Interface())
```
p-kozlowski: same here - 8 levels of nesting

extension/goext/environment.go
```diff
@@ -162,6 +162,46 @@ func (nf *NullFloat) UnmarshalJSON(b []byte) error {
...
+}
+
+// IsDefined returns whether undefined value is defined
+func (nf *NullFloat) IsDefined() bool {
+	return nf != nil
+}
+
+// Equals returns whether two undefined values are equal
+func (nf *NullFloat) Equals(rhv *NullFloat) bool {
+	return nf.IsDefined() && rhv.IsDefined() && nf.Valid && rhv.Valid && nf.Value == rhv.Value
```
przemyslaw-dobrowolski-cl: I am not sure about this logic.
Two undefined values should be equal in my opinion. The same applied to two invalid values.

extension/goext/environment.go
```diff
@@ -162,6 +162,46 @@ func (nf *NullFloat) UnmarshalJSON(b []byte) error {
...
+}
+
+// Equals returns whether two undefined values are equal
+func (nf *NullBool) Equals(rhv *NullBool) bool {
+	return nf.IsDefined() && rhv.IsDefined() && nf.Valid && rhv.Valid && nf.Value == rhv.Value
+}
+
+// Equals returns whether two undefined values are equal
+func (nf *NullString) Equals(rhv *NullString) bool {
+	return nf.IsDefined() && rhv.IsDefined() && nf.Valid && rhv.Valid && nf.Value == rhv.Value
```
przemyslaw-dobrowolski-cl: ditto

extension/goext/environment.go
```diff
@@ -162,6 +162,46 @@ func (nf *NullFloat) UnmarshalJSON(b []byte) error {
...
+}
+
+// Equals returns whether two undefined values are equal
+func (nf *NullFloat) Equals(rhv *NullFloat) bool {
+	return nf.IsDefined() && rhv.IsDefined() && nf.Valid && rhv.Valid && nf.Value == rhv.Value
+}
+
+// Equals returns whether two undefined values are equal
+func (nf *NullBool) Equals(rhv *NullBool) bool {
+	return nf.IsDefined() && rhv.IsDefined() && nf.Valid && rhv.Valid && nf.Value == rhv.Value
```
przemyslaw-dobrowolski-cl: ditto

extension/goext/environment.go
```diff
@@ -162,6 +162,46 @@ func (nf *NullFloat) UnmarshalJSON(b []byte) error {
...
+}
+
+// Equals returns whether two undefined values are equal
+func (nf *NullString) Equals(rhv *NullString) bool {
+	return nf.IsDefined() && rhv.IsDefined() && nf.Valid && rhv.Valid && nf.Value == rhv.Value
+}
+
+// Equals returns whether two undefined values are equal
+func (nf *NullInt) Equals(rhv *NullInt) bool {
+	return nf.IsDefined() && rhv.IsDefined() && nf.Valid && rhv.Valid && nf.Value == rhv.Value
```
przemyslaw-dobrowolski-cl: ditto

extension/goext/environment.go
```diff
@@ -193,3 +233,35 @@ func MakeNullFloat(value float64) NullFloat {
...
+// MakeUndefinedString allocates a new undefined string and sets its value
+func MakeUndefinedString(value string) *NullString {
+	return &NullString{
+		Value: value,
+		Valid: true,
+	}
+}
+
+// MakeUndefinedBool allocates a new undefined bool and sets its value
+func MakeUndefinedBool(value bool) *NullBool {
```
przemyslaw-dobrowolski-cl: The approach for tri-state values using pointer is dangerous in my opinion. In golang, bool, int and string types are value based whereas the newly introduced generalizations are pointer based. It means that the semantics of copy operation changed fundamentally:
```
var a *NullBool = MakeUndefinedBool(false)
b := a   // <- intuition is that we copy value, but in fact we copy pointer
a.Value = true
fmt.Println("%b", b.Value)    // <- returns true
```

extension/goext/environment.go
```diff
@@ -193,3 +233,35 @@ func MakeNullFloat(value float64) NullFloat {
...
+	return &NullBool{
+		Value: value,
+		Valid: true,
+	}
+}
+
+// MakeUndefinedInt allocates a new undefined int and sets its value
+func MakeUndefinedInt(value int) *NullInt {
```
przemyslaw-dobrowolski-cl: I'd suggest another approach:
```
int STATE_UNDEFINED = 0   // <- this is OK when set to default by compiler
int STATE_NULL = 1
int STATE_NOT_NULL = 2
type MaybeString struct {
    int state;
    Value string
}
```

extension/goplugin/util.go
```diff
@@ -118,16 +199,54 @@ func resourceFromMap(context map[string]interface{}, rawType reflect.Type) (res
...
 			}
 		}
 	}
 
-	return resource.Interface(), nil
+	return nil
 }
 
+func anyOfKinds(kind reflect.Kind, kinds []reflect.Kind) bool {
+	for _, k := range kinds {
```
przemyslaw-dobrowolski-cl: Use multicase switch for this, iteration is bad

zimnx: done for primitive and nullable types, I've left this method for short slices

extension/goext/environment.go
zimnx: done

extension/goext/environment.go
zimnx: I agree. Value based is better approach. Please see Maybe structs

extension/goext/environment.go
zimnx: done

extension/goext/environment.go
zimnx: done

extension/goext/environment.go
zimnx: done

extension/goext/environment.go
zimnx: I've applied equality logic from JS. Please check comment above Equals functions to see logic table

extension/goext/environment.go
zimnx: done

extension/goext/environment.go
```diff
@@ -162,6 +162,46 @@ func (nf *NullFloat) UnmarshalJSON(b []byte) error {
...
+	return nf != nil
+}
+
+// Equals returns whether two undefined values are equal
+func (nf *NullFloat) Equals(rhv *NullFloat) bool {
+	return nf.IsDefined() && rhv.IsDefined() && nf.Valid && rhv.Valid && nf.Value == rhv.Value
+}
+
+// Equals returns whether two undefined values are equal
+func (nf *NullBool) Equals(rhv *NullBool) bool {
```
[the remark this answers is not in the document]
zimnx: fixed

extension/goext/environment.go
```diff
@@ -162,6 +162,46 @@ func (nf *NullFloat) UnmarshalJSON(b []byte) error {
...
+	return ni != nil
+}
+
+// IsDefined returns whether undefined value is defined
+func (nf *NullFloat) IsDefined() bool {
+	return nf != nil
+}
+
+// Equals returns whether two undefined values are equal
+func (nf *NullFloat) Equals(rhv *NullFloat) bool {
```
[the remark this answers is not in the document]
zimnx: done

extension/goext/environment.go
```diff
@@ -162,6 +162,46 @@ func (nf *NullFloat) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// IsDefined returns whether undefined value is defined
```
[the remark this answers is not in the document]
zimnx: Changed in Maybe implementation

extension/goplugin/util.go
zimnx: done

extension/goplugin/util.go
```diff
@@ -74,41 +74,122 @@ func Finish(testReporter gomock.TestReporter) {
...
+					nestedField := reflect.New(elemType).Elem()
+					if structures {
+						if err := resourceFromMap(v.([]map[string]interface{})[i], nestedField); err != nil {
+							return err
+						}
+					} else if interfaces {
+						nestedValue := v.([]interface{})[i]
+						if nestedMap, ok := nestedValue.(map[string]interface{}); ok {
+							if err := resourceFromMap(nestedMap, nestedField); err != nil {
+								return err
```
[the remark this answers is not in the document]
zimnx: done

zimnx:
Review remars applied

extension/goplugin/schemas_test.go
```diff
@@ -436,11 +436,11 @@ var _ = Describe("Schemas", func() {
 			Expect(context).To(HaveKey("subobject"))
 			Expect(context["subobject"]).To(BeNil())
 		})
-		It("should convert resource to context with null string", func() {
+		It("should convert resource to context with null string defined", func() {
 			resource := &test.Test{
 				ID:          "42",
 				Description: "test",
-				Name:        goext.MakeNullString("testName"),
+				Name:        goext.MakeMaybeString("testName"),
```
przemyslaw-dobrowolski-cl: MakeMaybeString and MakeNullString are inconcise for me. Once there is Maybe word and otherwise not.
I'd suggest naming:
MakeString / MakeNullString / MakeUndefinedString

extension/goplugin/types_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	BeforeEach(func() {
+		env = goplugin.NewEnvironment("test", nil, nil)
+	})
+
+	AfterEach(func() {
+		env.Stop()
+	})
+
+	Describe("JSON Marshalling", func() {
```
przemyslaw-dobrowolski-cl: Please add some tests for the specific equality operator to ensure no future changes break its definition.

extension/goplugin/util.go
```diff
@@ -74,58 +73,157 @@ func Finish(testReporter gomock.TestReporter) {
...
+			} else {
+				return fmt.Errorf("invalid type of '%s' field (%s, expecting %s)", fieldName, value.Kind(), field.Kind())
+			}
+		}
+	}
+	return nil
+}
+
+func isMaybeType(kind reflect.Kind, typeName string) bool {
+	return anyOfKinds(kind, []reflect.Kind{reflect.Ptr, reflect.Struct}) && strings.HasPrefix(typeName, "goext.Maybe")
```
przemyslaw-dobrowolski-cl: I don't like checking substrings because this is an operation that compiler cannot optimize. Maybe it is possible to add simple comma-separated switch here?

extension/goplugin/util.go
```diff
@@ -74,58 +73,157 @@ func Finish(testReporter gomock.TestReporter) {
...
+		field.FieldByName("State").SetInt(int64(goext.MaybeNotNull))
+		if err := assignMapValueToField(mapValue, fieldName, field.FieldByName("Value")); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func sliceToMap(context map[string]interface{}, fieldName string, field reflect.Value) error {
+	if v, ok := context[fieldName]; ok {
```
przemyslaw-dobrowolski-cl: It better to implement from specific cases to generic cases: first handle any specific cases and only after interpret the generic case. This way there will be less indentation.

extension/goplugin/util.go
```diff
@@ -136,34 +234,148 @@ func (util *Util) ResourceToMap(resource interface{}) map[string]interface{} {
...
+		}
+	}
+}
+
+func maybeToMap(fieldsMap map[string]interface{}, fieldName string, value reflect.Value) {
+	if value.Kind() == reflect.Ptr {
+		if !value.IsNil() {
+			value = value.Elem()
+		} else {
+			// nothing, undefined value shoudn't appear in map
```
przemyslaw-dobrowolski-cl: I'd prefer to add panic here instead silently ignoring the problem

extension/goplugin/util.go
zimnx: I thought there's not but actually it's possible to use switch. Thanks

zimnx:
Thanks for comments. Remarks are applied

extension/goext/types.go
```diff
@@ -0,0 +1,298 @@
...
+// Add it to your custom structure as a anonymous field to make structure 3-value
+// 3-value structures are marshalled into JSON value based on their state
+// Structures in:
+// - Undefined state won't appear in JSON at all
+// - Null state will appear in JSON as null
+// - NotNull state will appear in JSON as they would normally be marshalled
+//
+// Example:
+// Given structures:
+// type TestSuite struct {
```
p-kozlowski: in fact, this is a test - so it should be run automatically

extension/goext/types.go
```diff
@@ -0,0 +1,298 @@
...
+	return m.MaybeState == MaybeNull
+}
+
+// IsNotNull returns whether value is defined and not null
+func (m Maybe) IsNotNull() bool {
+	return m.MaybeState == MaybeNotNull
+}
+
+// MaybeString represents 3-valued string
+type MaybeString struct {
```
p-kozlowski: the API became hard to use correctly and easy to use incorrectly. consider:
```
str := MakeNullString()
...
// later somewhere in the code
str.Value = "some value"
// str.MaybeState is still MaybeNull. I'd expect MaybeNotNull, because a value has been assigned
```
I suggest adding Get/SetValue methods and making Value private. SetValue should set state to MaybeNotNull, GetValue should return the value + state or value + error

extension/goplugin/types_test.go
```diff
@@ -0,0 +1,185 @@
...
+	})
+
+	AfterEach(func() {
+		env.Stop()
+	})
+
+	Describe("Equity", func() {
+		DescribeTable("String",
+			func(lhv goext.MaybeString, rhv goext.MaybeString, expected bool) {
+				expected = lhv.Equals(rhv)
```
p-kozlowski: ```Expect(lhv.Equals(rhv)).Should(Equal(expected))```
I agree that poor tests are better than no tests at all, but please at least make sure that the test could detect anything. I see a lot of effort in preparing these test cases, but it's for nothing if the tests do not test anything. Please use TDD.

extension/goplugin/types_test.go
```diff
@@ -0,0 +1,185 @@
...
+			It("undefined value", func() {
+				buf, err := json.Marshal(TestResource{})
+				Expect(err).To(BeNil())
+				Expect(string(buf)).To(Equal("{\"value\":null}"))
+			})
+
+			It("null value", func() {
+				buf, err := json.Marshal(TestResource{Value: goext.MakeNullFloat()})
+				Expect(err).To(BeNil())
+				Expect(string(buf)).To(Equal("{\"value\":null}"))
```
p-kozlowski: minor: use could use backticks to avoid escapes:
``` `{"value":null}` ```

File: extension/goext/types.go (hunk 20 above)
Comment (przemyslaw-dobrowolski-cl) [reply extension/goext/types.go]: +1

Comment (zimnx) [reply extension/goext/types.go]: I see, that is a big problem for custom structures embedding goext.Maybe. We can't enforce user to make all fields as private, and rely that all setters will change state of object. Even if this is possible, such object would be very painful in use. I'll revert last change (embedding Maybe in custom structs), and make primitive types immutable.

File: extension/goplugin/types_test.go [hunk 23]
```diff
@@ -0,0 +1,185 @@
...
+			Entry("undefined vs defined", goext.MakeUndefinedString(), goext.MakeString("hello"), false),
+			Entry("null vs undefined", goext.MakeNullString(), goext.MakeUndefinedString(), true),
+			Entry("null vs defined", goext.MakeNullString(), goext.MakeString("hello"), false),
+			Entry("defined vs defined", goext.MakeString("hello"), goext.MakeString("hello"), true),
+			Entry("defined vs defined different value", goext.MakeString("hello"), goext.MakeString("world"), false),
+		)
+
+		DescribeTable("Float",
+			func(lhv goext.MaybeFloat, rhv goext.MaybeFloat, expected bool) {
+				expected = lhv.Equals(rhv)
```
p-kozlowski: still not good (here and elsewhere)

zimnx:
Rebased. All comments are applied. Please review again

extension/goplugin/types_test.go
zimnx: fixed

extension/goext/types.go
```diff
@@ -0,0 +1,329 @@
...
+//	 Name string  `json:"name"`
+//	 TestSuite TestSuite `json:"test_suite"`
+// }
+// func MakeNullTestSuite() Test {
+// 	 return TestSuite{
+//   Maybe: goext.Maybe{MaybeState: goext.MaybeNull}
+//	 }
+// }
+//
+// json.Marshal(env.Util().ResourceToMap(&Test{ID: "some-id", Name:"My resource"}))
```
przemyslaw-dobrowolski-cl: Marshall

extension/goext/types.go
przemyslaw-dobrowolski-cl: ok, got it

extension/goplugin/types_test.go
```diff
@@ -0,0 +1,185 @@
...
+			It("value defined", func() {
+				buf, err := json.Marshal(TestResource{Value: goext.MakeString("hello")})
+				Expect(err).To(BeNil())
+				Expect(string(buf)).To(Equal(`{"value":"hello"}`))
+			})
+
+			It("value undefined", func() {
+				buf, err := json.Marshal(TestResource{})
+				Expect(err).To(BeNil())
+				Expect(string(buf)).To(Equal(`{"value":null}`))
```
morrisson: Please correct me if I'm wrong. As far as I remember, undefined value is omitted with key in JS extension. That is, if there is `{"key1": null, "key2": undefined, "key3": "string"}`, the result JSON will be `{"key1": null, "key3": "string"}`.

extension/goext/types.go
```diff
@@ -0,0 +1,329 @@
...
+}
+
+// IsNull returns whether value is null
+func (m Maybe) IsNull() bool {
+	return m.MaybeState == MaybeNull
+}
+
+// IsNotNull returns whether value is defined and not null
+func (m Maybe) IsNotNull() bool {
+	return m.MaybeState == MaybeNotNull
```
morrisson: In the case of undefined, this return false, right? Is it really convenient to strictly distinguish null and undefined in Go side?

zimnx: Yes, it will return false.  It is convenient, since JSON has 3 state values. Developer should take care of it, because API of other application may has different interpretation of undefined and null. If case doesn't apply, developer have choice to not care about difference.

extension/goplugin/types_test.go
zimnx: Yes, it should be interpreted like that. Unfortunately json.Marshaller and json.Unmarshaller interface AFAIK doesn't support this case. That's why we should always use env.Util().Resource(To/From)Map and then marshal return map to JSON. These tests just cover how UnmarshalJSON and MarshalJSON functions are implemented on this types. This case but when using env.Util().Resource(To/From)Map are covered in utils_test.go.

extension/goext/types.go
morrisson: Ok, then I think a kind of `IsValid` feature for this MaybeX type. I do not want to write so many `!val.IsNll() && !val.IsUndefined()`.

extension/goplugin/types_test.go
morrisson: Got it. That makes sense.

morrisson:
By the way, you need to squash commits. There are a lot of "Review remarks" commits.

zimnx:
> I believe there are some needs to properly handle undefined values. However, I could not get completely what the point is from this PR.
Is there any example where golang extension needs to be aware of undefined value except for cases of marshaling/unmarshaling JSON?

Basically it's only for proper marshalling and unmarshalling of JSONs. Gohan is a proxy to DB, where resources are stored. DB columns have only 2 value logic. Null or something primitive (number, string etc.). 

Golang also supports only 2 value logic. If you create instance of some struct, all fields will initialized 'zero-value'. It's very inconvenient for primitive types because integers will be 0, string will be "" etc.

But in resource schema it's possible to define object. Objects will be translated into JSON string which will be stored in DB.  As we know JSON supports additional logic value which is undefined. 

So if we wouldn't support undefined values, some zero-values would be putted into these json fields. This may cause some unwanted behavior. Image case when some microservice configuration is pushed remotely. 
Configuration is represented as given struct:
```
type Config struct {
DbDSN string `json:"db_dsn"`
MaxDatabaseConn int `json:"max_db_conn"`
}
```

Developer will allocate object of that type, and fill all fields except MaxDatabaseConn one. 
config := Config{ DbDSN: "something" }
Marshalled JSON will look like this: 
`{"db_dsn": "something", "max_db_conn": 0}`

This will cause that microservice won't create any DB connections because configuration tells it that connection pool size must be 0. 
If microservice would have it's own default value for this pool size, it could use it in case when this value wasn't provided. But it is so it uses it. 

But main reason is that we want to support undefined values in Go extensions is backward compatibility. Because all projects where Gohan is used was built using JS extensions (which supported 3-value logic) we must ensure all API responses will be the same to not change application behavior.

zimnx:
> By the way, you need to squash commits. There are a lot of "Review remarks" commits.

done

zimnx:
Please don't merge it. I'll do that. I need to sync one internal PR with this one.

PR #550, opened 2017-10-11
Refactored transactions

michalzurawski:
This commit removes Pre/Post handlers. They should not be trigger in general case - it is possible that another function, already in transaction, will call i.e. `CreateRaw` thus triggering PreCreate which should not be called in transaction.
Additionally Gohan no longer creates transaction if missing - `goext.Within` should be called to create one. To avoid mistakes all db functions will panic if called without or with closed transaction.

### DIFF
--- a/extension/goplugin/schemas.go
+++ b/extension/goplugin/schemas.go
@@ -17,7 +17,6 @@ package goplugin
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"reflect"
 
@@ -139,26 +138,6 @@ func (schema *Schema) structToResource(resource interface{}) (*gohan_schema.Reso
 	return gohan_schema.NewResource(schema.raw, fieldsMap)
 }
 
-func (schema *Schema) assignField(name string, field reflect.Value, value interface{}) error {
-	if field.Kind() == reflect.Struct || field.Kind() == reflect.Slice || field.Kind() == reflect.Ptr {
-		mapJSON, err := json.Marshal(value)
-		if err != nil {
-			return err
-		}
-		newField := reflect.New(field.Type())
-		fieldJSON := string(mapJSON)
-		fieldInterface := newField.Interface()
-		err = json.Unmarshal([]byte(fieldJSON), &fieldInterface)
-		if err != nil {
-			return err
-		}
-		field.Set(newField.Elem())
-	} else {
-		setValue(field, reflect.ValueOf(value))
-	}
-	return nil
-}
-
 // ListRaw lists schema raw resources
 func (schema *Schema) ListRaw(filter goext.Filter, paginator *goext.Paginator, requestContext goext.Context) ([]interface{}, error) {
 	return schema.listImpl(requestContext, func(ctx context.Context, tx goext.ITransaction) ([]map[string]interface{}, uint64, error) {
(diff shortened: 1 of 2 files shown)


marcin-ptaszynski:
@michalzurawski It's important to note that this in only for goplugin internals, removing pre/post_create in general would be bad, because of already present dependencies.

extension/goplugin/schemas.go
```diff
@@ -390,103 +328,40 @@ func (schema *Schema) create(rawResource interface{}, requestContext goext.Conte
...
-	if triggerEvents {
-		if err := schema.env.HandleEvent(goext.PreCreate, requestContext); err != nil {
-			return err
-		}
-	}
-
-	tx, err := schema.env.Database().Begin()
-	if err != nil {
-		return err
+	if !hasOpenTransaction || tx.Closed() {
```
p-kozlowski: 1. naming: the transaction may be there, but it may be closed
2. code duplication:
```
func mustGetActiveTransaction(ctx goext.Context) ITransaction {
    tx, hasTx := contextGetTransaction(ctx)
    if !hasTx || tx.Closed() {
        panic(...)
    }
    return tx
}
```

morrisson:
I agree `CreateRaw` shouldn't call `PreCreate` and `PostCreate` because it's already in the transaction. However, I couldn't get your point why `PreCreate` and `PostCreate` should be removed completely.  Actually, this was a bug of goplugin implementation so just we need to fix them.

`PreCreate` is very useful to validate API input values before opening DB transaction.

michalzurawski:
@morrisson Sorry for not being precise but as @marcin-ptaszynski mentioned this commit removes Pre/Post only from `CreateRaw` etc. not entirely from Gohan - this handlers will still be called after API call when we can be sure they aren't called from transaction.

morrisson:
@michalzurawski Oh, got it. Then that should be okay. By the way, how about `UpdateRaw` and `DeleteRaw`?

michalzurawski:
@p-kozlowski applied changes + rebased
@morrisson following handlers have been removed:

- `PreCreate` and `PostCreate` from `CreateRaw`
- `PreUpdate` and `PostUpdate` from `UpdateRaw`
- `PreDelete` and `PostDelete` from `DeleteRaw`

Additionally all database function (`CreateRaw`, `UpdateRaw`, `DeleteRaw`, all fetches and lists) now require that context with open transaction has been passed to them

extension/goplugin/schemas.go
```diff
@@ -714,3 +515,18 @@ func contextSetTransaction(ctx goext.Context, tx goext.ITransaction) goext.Conte
...
 	return ctx
 }
+
+func contextGetMapResource(ctx goext.Context) map[string]interface{} {
+	return ctx["resource"].(map[string]interface{})
+}
+
+func mustGetOpenTransactionFromContext(context goext.Context) goext.ITransaction {
+	if context == nil {
+		panic("Database function called without open transaction")
```
morrisson: nit: it's better to change error message with following message.

p-kozlowski: do you mean this should be ```return fmt.Errorf("database...")```? In my opinion this case is incorrect API usage that couldn't be spotted at compile time (at least at this moment). Because of that, I prefer early error detection and panic() is perfectly suitable for that, because it gives immediate feedback that's hard to miss.

PR #551, opened 2017-10-11
Don't lock resources during delete

p-kozlowski:
There's a couple of problems with locking:
- deadlocks may (and will) appear
- locking related resources is too course-grained: it may lock all resources belonging to given tenant

### DIFF
--- a/extension/goplugin/schemas.go
+++ b/extension/goplugin/schemas.go
@@ -606,7 +606,7 @@ func (schema *Schema) delete(filter goext.Filter, requestContext goext.Context, 
 	contextTx := goext.MakeContext()
 	contextSetTransaction(contextTx, tx)
 
-	fetched, err := schema.LockListRaw(filter, nil, contextTx, goext.LockRelatedResources)
+	fetched, err := schema.ListRaw(filter, nil, contextTx)
 	if err != nil {
 		return err
 	}


marcin-ptaszynski:
@michalzurawski , There is a good reason to have lock in initial phase of delete operation. First of all, DELETE SQL STMT will lock matched rows regardless, so Deadlocks can still occur anyway. I agree, locking with related resources in general is not great, but with SKIP_RELATED it should work fine.

p-kozlowski:
@marcin-ptaszynski it's a design question if we want to mirror Gohan-core behaviour (that locks (or not) according to the schema/metadata/lock_policy) or the JS extensions (which do not lock at all).

I think we should go in the latter direction - a lock inside the DB operation isn't always a good choice - if more than one lock should be taken, than it's still the caller responsibility to take all the necessary locks in proper order. Example:
```
Lock(x)
Lock(y)
Lock(z)
Delete(y)
    Lock(y)
    HandlePreDeleteTx(y)
    SqlDelete(y)
    HandlePostDeleteTx(y)
```

This way we're taking the same lock twice, and it looks like a waste. What's more, if the user cares about locking, in the trivial case (just one lock needed) they should define that in the schema, in non-trivial cases, the locking has to be done in the caller side anyway.

To sum up, I think we should either use locking_policy or skip locking altogether. I'd rather skip locking, but I don't have a very strong opinion about it. What do you think?

marcin-ptaszynski:
@p-kozlowski ,
{quote}
This way we're taking the same lock twice, and it looks like a waste.
{quote}
What's wasteful about that? Taking same lock twice doesn't cost anything, because we're executing SELECT, regardless.

I actually don't have strong opinion either, since it can be done by developer anyway.

But I wanted to point out, deadlocks will still happen.

Issue #498, opened 2017-07-19
Ensure bindata.go is not shown as "unstaged changes" needlessly


vozhyk-:
This is a follow-up for #496.

`bindata.go` is regenerated on every `make` and is added to the repository. This makes it appear as "unstaged changes" after the source is built.

There are 2 ways the problem can be fixed:
- Removing the file from the repository and adding it to `.gitignore` (#496). This isn't desirable because it
  >  will cause 'go get' to fail since there are no post-go-get hooks that can recreate this file.
- Making the rule that regenerates the file separate from the `all` rule and making sure the file is regenerated whenever the data changes
  - with a commit hook?

PR #553, opened 2017-10-11
Adding order_properties_before to schema

skizhak:
- This is to support ordering Properties of model which extends other models.
- Properties of schema listed under Extends will be ordered after current model properties if they're listed under order_properties_before. so the final order of properties is
properties of extends (not listed in order_properties_before)
current properties
properties of rest of extends model

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -21,6 +21,7 @@ Schemas might also have any of the following optional properties.
 - metadata  -- application specific schema metadata (object)
 - type      -- can be an abstract or empty string (see more in schema inheritance)
 - extends   -- list of base schemas
+- order_properties_before -- to order properties before properties of extended schemas. 
 
 ## Schema Inheritance
 
@@ -357,6 +358,9 @@ eg.
           type: object
           unique: false
 ```
+- order_properties_before
+
+  when resource is extended using 'extends' by default all properties of extended schema gets ordered before the current schema properties. Use this field to order current resource properties before extended resource properties.
 
 ## Indexes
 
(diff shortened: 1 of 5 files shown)


morrisson:
At least, you should add brief doc to `schema.md` about this meta property. This feature is ok for me, but I think it's enough and clear for me to write down every properties in propertyOrder list.

skizhak:
yes ill update the doc and add some UT.

przemyslaw-dobrowolski-cl:
@skizhak please do not commit util/bindata.go which is a generated file

.gitignore
```diff
@@ -45,3 +45,5 @@ node_modules
 .idea
 etc/gohan.log
 etc/gohan.db
+
+util/bindata.go
```
przemyslaw-dobrowolski-cl: bindata.go cannot be ignored. See the comments of the following rejected PR:
https://github.com/cloudwan/gohan/pull/496
and notes
https://github.com/cloudwan/gohan/issues/498

skizhak: i got it. but i think this particular commit will require the bindata change to use the new schema field. i will remove the .gitignore but i think i should be committing bindata so people doing 'go get ' install can use this.

przemyslaw-dobrowolski-cl:
@skizhak You're right - in this particular case, bindata must be regenerated and it is correctly added to commit. Approved.

skizhak:
@przemyslaw-dobrowolski-cl thanks. @morrisson @nati gentle reminder on this PR.

PR #554, opened 2017-10-12
[ESI-15270] No parallel builds on CircleCI to prevent out-of-memory crashes

p-kozlowski:


p-kozlowski:
this PR does not work, fixed correctly in #555

PR #555, opened 2017-10-12
[ESI-15270] Don't bloat memory with binary data

p-kozlowski:


p-kozlowski:
Test results:
1. fail (4 != 3 in sync)

PR #557, opened 2017-10-13
Make ErrMissingType more informative

p-kozlowski:
It happens the the error message "resource type not registered" gets printed in the logs, but this message is not informative because it's hard to tell which resource type is missing.

I was thinking of a panic() instead of returning an error, but finally I've decided not to change the API. Reviewers, what's your opinion about it?

### DIFF
--- a/extension/goplugin/schemas.go
+++ b/extension/goplugin/schemas.go
@@ -30,10 +30,11 @@ import (
 var (
 	// ErrNotPointer indicates that a resource was not passed by a pointer
 	ErrNotPointer = fmt.Errorf("raw resource must be passed by a pointer")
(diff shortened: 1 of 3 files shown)


michalzurawski:
Please add same for schemas.go:252 `resourceType, _ := schema.env.getType(schema.ID())`

p-kozlowski:
@michalzurawski done, I've also fixed (Lock)Fetch(Raw)

p-kozlowski:
rebased & squashed. @morrisson @marcin-ptaszynski @zimnx @przemyslaw-dobrowolski-cl gentle reminder about this PR

PR #560, opened 2017-10-16
Fixed a crash when using logger in an environment with no plugins loaded

przemyslaw-dobrowolski-cl:


extension/goplugin/environment_test.go
```diff
@@ -50,6 +50,16 @@ var _ = Describe("Environment", func() {
 		env = goplugin.NewEnvironment("test", nil, nil)
 	})
 
+	Describe("Internal state consistency", func() {
```
p-kozlowski: still failes if Schemas() returns empty list:
```
13:43:22.906 extest CRITICAL  failed to load schemas extensions: failed to init db: No schemas in Manager. Did you remember to load them?
```

extension/goplugin/transaction_test.go
```diff
@@ -68,7 +68,12 @@ var _ = Describe("Transaction", func() {
 			testSchema = env.Schemas().Find("test")
 			Expect(testSchema).To(Not(BeNil()))
 
-			err = db.InitDBWithSchemas(DbType, DbFile, true, true, false)
+			err = db.InitDBWithSchemas(DbType, DbFile, db.InitDBParams{
```
p-kozlowski: there's some suspicious inconsistency here: in some tests it's true/false/false/true, in others false/false/false/true and here it's true/true/false/true. why is that?
could we add some universal safe defaults? eg.
```
func DefaultTestDBParams() InitDbParams
```

przemyslaw-dobrowolski-cl: ok, fixed

przemyslaw-dobrowolski-cl:
rebased

extension/goplugin/transaction_test.go
```diff
@@ -67,9 +67,7 @@ var _ = Describe("Transaction", func() {
 			Expect(env.Start()).To(Succeed())
 			testSchema = env.Schemas().Find("test")
 			Expect(testSchema).To(Not(BeNil()))
-
-			err = db.InitDBWithSchemas(DbType, DbFile, true, true, false)
-			Expect(err).To(BeNil())
+			Expect(db.InitDBWithSchemas(DbType, DbFile, db.DefaultTestInitDBParams())).To(BeNil())
```
p-kozlowski: minor: Expect(...).To(Succeed())

przemyslaw-dobrowolski-cl: fixed

przemyslaw-dobrowolski-cl:
rebased

extension/goplugin/environment_test.go
przemyslaw-dobrowolski-cl: it should be already fixed

PR #562, opened 2017-10-16
Return non-zero value when any test is failing

zimnx:


p-kozlowski:
looks good, let's merge this PR. 
at the same time, there's a greater issue here. there's no a single UT for goplugin test framework. we should take care of it, as incorrect results (think false positives) could do a lot of harm, especially when run on an CI.

PR #563, opened 2017-10-18
Added copy context for delete

michalzurawski:


extension/goplugin/schemas.go
```diff
@@ -433,11 +433,11 @@ func (schema *Schema) delete(filter goext.Filter, requestContext goext.Context,
 		resourceID := mapper.FieldByName(resource, "id").Interface()
 
 		mapFromResource := schema.env.Util().ResourceToMap(resource.Interface())
-		contextTx = contextTx.WithResource(mapFromResource).
+		contextCopy = contextCopy.WithResource(mapFromResource).
```
morrisson: There is a bit different way to copy context among create, update and delete so please do it in the same way.

michalzurawski:
@morrisson  Thanks for comment - now they all contain `schema_id`, `resource` and `id`

PR #564, opened 2017-10-20
Ensure all goext resources inherit IResourceBase

przemyslaw-dobrowolski-cl:
This change ensures (during compilation) that RegisterType receives an object of type ResourceBase

### DIFF
--- a/extension/goext/schemas.go
+++ b/extension/goext/schemas.go
@@ -168,8 +168,8 @@ type ISchema interface {
 	// RegisterEventHandler registers an event handler for a named event with given priority
 	RegisterEventHandler(event string, handler func(context Context, resource Resource, environment IEnvironment) error, priority int)
 
-	// RegisterType registers a resource type, derived from BaseResource
-	RegisterType(resourceType interface{})
+	// RegisterType registers a resource type, derived from IResourceBase
+	RegisterType(resourceType IResourceBase)
 
 	// RegisterRawType registers a raw resource type, containing db annotations
 	RegisterRawType(rawResourceType interface{})
--- a/extension/goext/environment.go
+++ b/extension/goext/environment.go
@@ -35,11 +35,38 @@ type IEnvironment interface {
 	Reset()
 }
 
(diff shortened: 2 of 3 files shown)


extension/goext/schemas.go
```diff
@@ -162,8 +162,8 @@ type ISchema interface {
 	// RegisterEventHandler registers an event handler for a named event with given priority
 	RegisterEventHandler(event string, handler func(context Context, resource Resource, environment IEnvironment) error, priority int)
 
-	// RegisterType registers a resource type, derived from BaseResource
-	RegisterType(resourceType interface{})
+	// RegisterType registers a resource type, derived from IResourceBase
+	RegisterType(resourceType IResourceBase)
 
 	// RegisterRawType registers a raw resource type, containing db annotations
 	RegisterRawType(rawResourceType interface{})
```
p-kozlowski: perhaps this is not the perfect moment, but we should keep in mind, that similar issues may happen in `RegisterRawType`, `(Db)[Create|Update](Raw)` and return types of `(Lock)[Fetch|List](Raw)`

converter/item/dbkind_test.go
```diff
@@ -41,7 +41,7 @@ var _ = Describe("db kind tests", func() {
 
 			result := dbKind.Type("", newItem)
 
-			expected := "goext." + util.ToGoName("null", typeOfItem)
+			expected := "goext." + util.ToGoName("maybe", typeOfItem)
```
morrisson: Duplicated #549 and should not be included in this PR.

morrisson:
Please modify commit message and PR title. This does not describe general issue in open source context.

extension/goext/schemas.go
przemyslaw-dobrowolski-cl: Good idea - I'll create a separate PR for that.

przemyslaw-dobrowolski-cl:
@morrisson I fixed the description and removed unrelated commit from PR.

PR #565, opened 2017-10-20
Add possible deadlock detection and warnings

przemyslaw-dobrowolski-cl:


zimnx:
Add config flag which enables this feature. Release packages shouldn't have it.

extension/goplugin/database.go
```diff
@@ -51,17 +52,29 @@ func (db *Database) Clone() *Database {
 	}
 }
 
+func isWithin() bool {
+	pc, _, _, ok := runtime.Caller(4)
```
p-kozlowski: magic number. add a test for this feature in case someone changes something in the future.
Also, according to [the docs ](https://golang.org/pkg/runtime/#FuncForPC) the result may be inaccurate
> If pc represents multiple functions because of inlining, it returns the *Func describing the outermost function.

extension/goplugin/database.go
```diff
@@ -51,17 +52,29 @@ func (db *Database) Clone() *Database {
...
+		panic("runtime.FuncForPC failed")
+	}
+	return f.Name() == "github.com/cloudwan/gohan/extension/goext.within"
+}
+
 // Begin starts a new transaction
 func (db *Database) Begin() (goext.ITransaction, error) {
 	t, _ := db.raw.Begin()
-	return &Transaction{t}, nil
+	return &Transaction{t, isWithin()}, nil
```
p-kozlowski: If Begin() shouldn't be called directly, it should be a private function (and we wouldn't need to add any runtime checks).
On the other hand, if some plugins really need transactions not started by Within(), we'll get a lot of spurious erros. If we want to keep the functionality, let's call it BeginUnsafe()

przemyslaw-dobrowolski-cl:
closed, this feature is not useful

PR #569, opened 2017-10-26
Run go fmt

p-kozlowski:


util/bindata.go
```diff
@@ -581,28 +581,28 @@ func AssetNames() []string {
...
 // _bindata is a table, holding each asset generator, mapped to its name.
 var _bindata = map[string]func() (*asset, error){
-	"etc/schema/core.json": etcSchemaCoreJson,
-	"etc/schema/gohan.json": etcSchemaGohanJson,
-	"etc/extensions/gohan_extension.yaml": etcExtensionsGohan_extensionYaml,
-	"etc/templates/dot.tmpl": etcTemplatesDotTmpl,
-	"etc/templates/markdown.tmpl": etcTemplatesMarkdownTmpl,
-	"etc/templates/openapi.tmpl": etcTemplatesOpenapiTmpl,
-	"etc/templates/server.tmpl": etcTemplatesServerTmpl,
+	"etc/schema/core.json":                              etcSchemaCoreJson,
```
przemyslaw-dobrowolski-cl: I'm not quite sure if this is a good idea to gofmt a generated file. The corresponding Makefile gen secrtion doesn't do that so regenerating the same file will result in a non-empty diff. Maybe Makefile should also contain go fmt?

p-kozlowski: done

PR #571, opened 2017-10-30
Random coefficient added to sleep interval during deadlock handling

zimnx:


db/db.go
```diff
@@ -97,8 +98,12 @@ func withinTxImpl(db DB, beginStrategy func(db DB) (transaction.Transaction, err
 			return err
 		}
 
-		log.Warning("scoped transaction deadlocked, retrying %d / %d", attempt, db.Options().RetryTxCount)
-		time.Sleep(db.Options().RetryTxInterval)
+		retryInterval := db.Options().RetryTxInterval
```
p-kozlowski: this function violates SRP. I suggest introducing `func getRetryInterrval(options.Options) time.Duration` or `func getRetryInterrval(DB) time.Duration`

przemyslaw-dobrowolski-cl:
@zimnx please update

zimnx:
> In general I like the idea. Have you considered alternative approach of using exponential backoff?
I thought about it, it would require much more code then this approach. First let's see if it works good on production, and try another approach if this isn't enough.

PR #572, opened 2017-10-31
[ESI-15686] Keystone authorization cache

zimnx:
During scale tests we found out that hitting keystone during authorization of each request was performance bottle-neck. This commit introduces memory cache, which stores tokens for configurable time duration.
It's OFF by default.


tools/dev_setup.sh
```diff
@@ -6,3 +6,4 @@ go get -u github.com/gordonklaus/ineffassign
 go get -u github.com/jteeuwen/go-bindata/go-bindata
 go get -u github.com/client9/misspell/cmd/misspell
 go get -u github.com/vektra/mockery/cmd/...
+go get -u github.com/patrickmn/go-cache
```
morrisson: We should use vendoring, right?

server/middleware/cached.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (c *CachedIdentityService) GetClient() *gophercloud.ServiceClient {
+	return c.inner.GetClient()
+}
+
+func NewCachedIdentityService(inner IdentityService, ttl time.Duration) IdentityService {
+	cleanupInterval := 4 * ttl
+	return &CachedIdentityService{
+		inner: inner,
+		cache: cache.New(ttl, cleanupInterval),
```
morrisson: Keystone returns expiration time. This cache ttl must not exceed this expiration time.

server/middleware/cached_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"time"
+
+	"github.com/cloudwan/gohan/schema"
+	"github.com/golang/mock/gomock"
+	"github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/rackspace/gophercloud"
+)
+
+var _ = ginkgo.Describe("Cached identity service", func() {
```
p-kozlowski: why not the . notation?

server/middleware/cached_test.go
```diff
@@ -0,0 +1,67 @@
...
+	ginkgo.It("Pass GetTenantID to inner service", func() {
+		mockedIdentityService.EXPECT().GetTenantID(tenantName).Return(tenantID, nil)
+		rv, err := cachedIdentityService.GetTenantID(tenantName)
+		Expect(rv).To(Equal(tenantID))
+		Expect(err).To(BeNil())
+	})
+
+	ginkgo.It("Pass GetTenantName to inner service", func() {
+		mockedIdentityService.EXPECT().GetTenantName(tenantID).Return(tenantName, nil)
+		rv, err := cachedIdentityService.GetTenantName(tenantID)
```
p-kozlowski: the inner service is actually a `keystoneV3Client`, it will send at least one HTTP request per each call. have you measured what's the performance overhead and if that's acceptable?
same issue applies to GetTenantId

server/middleware/cached_test.go
```diff
@@ -0,0 +1,67 @@
...
+		rv, err := cachedIdentityService.VerifyToken(token)
+		Expect(rv).To(Equal(auth))
+		Expect(err).To(BeNil())
+		rv, err = cachedIdentityService.VerifyToken(token)
+		Expect(rv).To(Equal(auth))
+		Expect(err).To(BeNil())
+	})
+
+	ginkgo.It("Pass GetTenantID to inner service", func() {
+		mockedIdentityService.EXPECT().GetTenantID(tenantName).Return(tenantID, nil)
```
p-kozlowski: false positive is possible here if the call is not forwarded to the inner service. if we really need to assert on the call to the inner service, there's a need for strict mock.

server/middleware/cached_test.go
zimnx: Because ginkgo declares struct with name "Context", and struct with same name is declared in middleware package. Compiler complains about double declaration.

tools/dev_setup.sh
zimnx: done

server/middleware/cached.go
zimnx: Added notes in documentation about it

server/middleware/cached_test.go
zimnx: Added caching for GetTenantName since it's used in each request. GetTenantID is not used anywhere.

server/middleware/cached_test.go
zimnx: Thanks, added strict mock.

docs/configuration.md
```diff
@@ -240,6 +240,17 @@ Gohan supports OpenStack Keystone authentication backend.
 
   v2.0 or v3 is supported
 
+- use_auth_cache
+
+  enable memory cache which stores keystone authorization responses for configurable time duration.    
+  Please note that any token may that will be revoked before TTL expiration.
```
michalzurawski: nit: _that will_ seems redundant

zimnx: done

morrisson:
@nati This is an improvement for gohan core part, so I would like you to take a glance to this PR. Thanks.

etc/gohan.yaml
```diff
@@ -53,6 +53,10 @@ keystone:
     user_name: "admin"
     tenant_name: "admin"
     password: "gohan"
+    # Whether should use cache for storing and fetching authorizations
+    use_auth_cache: true
```
morrisson: Default is set as false in the following code, so we should set false in this example config file.

zimnx: done

zimnx:
> Plus, we should add some lines in vendor.json

done

etc/gohan.yaml
morrisson: done? Maybe you forgot to update here.

zimnx: oops, lost this change during branch switching. Added again ;)

server/middleware/cached.go
```diff
@@ -0,0 +1,60 @@
...
+func (c *CachedIdentityService) GetTenantName(tenantID string) (string, error) {
+	i, ok := c.cache.Get(tenantID)
+	if ok {
+		return i.(string), nil
+	}
+	name, err := c.inner.GetTenantName(tenantID)
+	if err != nil {
+		return "", err
+	}
+	err = c.cache.Add(tenantID, name, cache.DefaultExpiration)
```
p-kozlowski: there's a race here. if the same token is received twice in a short time window, it'll be added twice to the cache. the first attempt will succeed, but the second will fail with an error. the error will be returned to the client as 401

p-kozlowski:
@nati, gentle reminder about this PR. Have you got any comments?

PR #573, opened 2017-11-02
Wait for response in HTTP client implementation

zimnx:
Current implementation contain race, and doesn't at all wait for http request result.

### DIFF
--- a/extension/goplugin/http.go
+++ b/extension/goplugin/http.go
@@ -49,14 +49,7 @@ func (http *HTTP) RequestRaw(ctx context.Context, method, rawURL string, headers
 		req.Header.Set(header, value)
 	}
 
-	var resp *net_http.Response
-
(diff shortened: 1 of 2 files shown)


extension/goplugin/http.go
```diff
@@ -57,6 +57,9 @@ func (http *HTTP) RequestRaw(ctx context.Context, method, rawURL string, headers
 		resp, err = net_http.DefaultTransport.RoundTrip(req)
 		close(done)
 	}()
+
+	<-done
```
marcin-ptaszynski: this whole approach with spawning goroutine is pretty pointless. Why not simply call:

```
resp, err = net_http.DefaultTransport.RoundTrip(req)
```
?

zimnx: Agree, I'll change that

PR #574, opened 2017-11-06
[ESI-15659] Fixed goroutine/memory leak when goplugin environment is stopped

przemyslaw-dobrowolski-cl:


extension/goplugin/sync.go
```diff
@@ -99,15 +103,27 @@ func (sync *Sync) Watch(ctx context.Context, path string, timeout time.Duration,
...
+	}
+}
+
+// Close closes Sync
+func (sync *Sync) Close() {
+	if atomic.AddUint32(&sync.ref.cnt, ^uint32(0)) == 0 {
+		sync.ref.raw.Close()
+		sync.ref.raw = nil
 	}
+	log.Debug("sync close: refs = %d", sync.ref.cnt)
```
p-kozlowski: I believe these logs were useful when debugging this specific problem, but as soon as this code goes to production, I'm worried it could be very noisy.
Let's add proper tests (I can't see any in this PR) instead of polluting the log.

p-kozlowski:
@przemyslaw-dobrowolski-cl could you elaborate a little bit what was the leak and how does this PR fix it?

marcin-ptaszynski:
@przemyslaw-dobrowolski-cl , in addition to @p-kozlowski question, why would you even want to Close DB object at all? AFAIK, it's only created once per Server and should be Closed only on gentle stop (or during tests). I don't see how this PR would fix anything.

przemyslaw-dobrowolski-cl:
@p-kozlowski @marcin-ptaszynski I uploaded the final PR for revision. Previous one was used by me for testing ans wasn't ready for review. I even didn't add any reviewers. Regarding the present version - it turned out that WatchContext implementation was leaking goroutines. This function is used by goext extensions in Sync and hence it caused massive goroutine leak. I included a test suite to ensure that none of Sync functions is leaking goroutines (previous implementation fails in this test suite).

sync/etcdv3/etcd.go
```diff
@@ -328,25 +329,32 @@ func (s *Sync) Watch(path string, responseChan chan *sync.Event, stopChan chan b
...
+				}
 			}
 		}
+		close(errCh)
+		close(stopCh)
+		close(eventCh)
 	}()
-
-	return responseChan, nil
+	return eventCh, nil
```
p-kozlowski: this function never returns an error, so I think we could remove the `error` from returned values

przemyslaw-dobrowolski-cl: ok, removed

sync/etcdv3/etcd_test.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	gohan_sync "github.com/cloudwan/gohan/sync"
 	etcd "github.com/coreos/etcd/clientv3"
+	"runtime"
```
marcin-ptaszynski: nit: import grouping

przemyslaw-dobrowolski-cl:
It seems that this fix works in production environments but it's difficult to write a UT test for that.
@p-kozlowski the package you mentioned gives even worse results that the one included in this PR. I will just remove UT for leaks for now.

PR #575, opened 2017-11-09
Fixed setting mock modules

michalzurawski:
Currently setting one mock module affects another i.e. after creating an object and setting sync to be mock module all information from DB is gone.

### DIFF
--- a/extension/goplugin/environment_mock.go
+++ b/extension/goplugin/environment_mock.go
@@ -41,62 +41,57 @@ type MockIEnvironment struct {
 	util     goext.IUtil
 }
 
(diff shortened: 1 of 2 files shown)


extension/goplugin/environment_mock.go
```diff
@@ -41,62 +41,57 @@ type MockIEnvironment struct {
 	util     goext.IUtil
 }
 
+func (mockEnv *MockIEnvironment) setModules() {
```
przemyslaw-dobrowolski-cl: A better name would be resetModules

PR #577, opened 2017-11-13
Add support for running gohan tests with mysql (MYSQL_TEST flag)

przemyslaw-dobrowolski-cl:


extension/framework/goplugin/runner/runner.go
```diff
@@ -180,9 +182,19 @@ func (testRunner *TestRunner) runSingle(t ginkgo.GinkgoTestingT, reporter *Repor
...
 		// db
-		dbFileName := dbBaseFileName + "_" + uuid.NewV4().String()
-		dbConnString := fmt.Sprintf("file:%s?mode=memory&cache=shared", dbFileName)
-		db, err := gohan_db.ConnectDB("sqlite3", dbConnString, gohan_db.DefaultMaxOpenConn, options.Default())
+		dbType := ""
+		dbConn := ""
+
+		if os.Getenv("MYSQL_TEST") == "true" {
+			dbType = "mysql"
+			dbConn = "root@/gohan_test"
```
zimnx: what about taking env variable MYSQL_DSN with this as default?

przemyslaw-dobrowolski-cl: This variable is never used in gohan. What kind of usage you have in mind?

zimnx: I mean that I have different credentials to my local mysql server. I would need to recompile gohan if I would like to run tests. Getting DSN string from env variable is much better solution.

przemyslaw-dobrowolski-cl: done, added optional DSN parameter

extension/framework/goplugin/runner/runner.go
```diff
@@ -180,9 +182,23 @@ func (testRunner *TestRunner) runSingle(t ginkgo.GinkgoTestingT, reporter *Repor
...
-		db, err := gohan_db.ConnectDB("sqlite3", dbConnString, gohan_db.DefaultMaxOpenConn, options.Default())
+		dbType := ""
+		dbConn := ""
+
+		if os.Getenv("MYSQL_TEST") == "true" {
+			dbType = "mysql"
+			if dsn, ok := os.LookupEnv("MYSQL_DSN"); ok {
+				dbConn = dsn
+			} else {
+				dbConn = "root@/gohan_test"
```
morrisson: Is hardcode ok?

extension/framework/goplugin/runner/runner.go
```diff
@@ -22,6 +22,8 @@ import (
 	"regexp"
 	"testing"
 
+	"os"
```
p-kozlowski: go fmt

extension/framework/goplugin/runner/runner.go
przemyslaw-dobrowolski-cl: It's not the best solution here. There were a dozen of similar places in code where connection was hardcoded. I added a commit which contains a common file for local database connection and clears Connect and InitSchema functions naming.

PR #578, opened 2017-11-13
Update vendor mattn/go-sqlite3

przemyslaw-dobrowolski-cl:


morrisson:
Updating vendor is basically ok, but could you describe why we need to update this?

przemyslaw-dobrowolski-cl:
@morrisson I found that sqlite3 library was recently updated a few times to fix some problems race condition problems. There exist an issue with interrupted transactions in gohan (goext test runner) which is related to a race condition in mattn/go-sqlite3, Therefore, I checked if the newer version of the library fixes that. It turned out that the number of tx interruptions decreased a few times but still occurs from time to time. I hope, that the library will be fixed completely but for now this PR reduces the number of interruptions and the other PR #579 is a hack to prevent interruptions at all.

PR #582, opened 2017-11-16
Exposing schema parent information

zimnx:


extension/goplugin/schemas.go
```diff
@@ -523,6 +523,11 @@ func (schema *Schema) Properties() []goext.Property {
 	return properties
 }
 
+// Parents returns schema_ids which given schema extends
+func (schema *Schema) Parents() []string {
+	return schema.raw.Extends
```
przemyslaw-dobrowolski-cl: In gohan, it it named 'extends' but in the interface it is parent. I think it's an inconsistency that should be fixed.

extension/goext/schemas.go
```diff
@@ -187,6 +187,9 @@ type ISchema interface {
 
 	// Properties returns properties of schema
 	Properties() []Property
+
+	// Parents return list of schema_ids which given schema extends
+	Parents() []string
```
przemyslaw-dobrowolski-cl: What is the rationale tor this new addition?

zimnx: Register event handlers only if schema extends some other schema. There is no way to check it right now

extension/goplugin/schemas.go
zimnx: I've changed that to keep consistency, but Extends doesn't fit to me at all.

extension/goext/schemas.go
jkmar: What about this: #541?

zimnx: #541 gives relation info but in other direction.

PR #584, opened 2017-11-17
Print stack trace on cron panic

zimnx:


server/cron.go
```diff
@@ -22,6 +22,8 @@ import (
 	"github.com/robfig/cron"
 
 	"errors"
+	"runtime/debug"
```
p-kozlowski: goimports

PR #586, opened 2017-11-21
[ESI-15953] Fixed panic when inner error of goext.Error was nil

zimnx:


extension/goext/error_test.go
```diff
@@ -45,6 +46,10 @@ var _ = Describe("Error", func() {
 			errDouble = goext.NewError(http.StatusServiceUnavailable, goext.NewError(http.StatusUseProxy, fmt.Errorf("nothing here")))
 			errDouble.Origin = "one.go:1000"
 			errDouble.Err.(*goext.Error).Origin = "two.go:2000"
+
+			var nilErr *goext.Error = nil
+			errDoubleNil = goext.NewError(http.StatusInternalServerError, nilErr)
```
p-kozlowski: this looks very strange. why do we allow wrapping a nil error with an error code? what's the use case?

zimnx: We don't, but we can't defend from default value of goext.Error. I've changed testcase.

PR #588, opened 2017-11-23
Inform about missing schema

michalzurawski:
Please note that without this PR there will be panic in line 723 after recurrence call related to nil pointer - this PR just inform about the reason for panic

### DIFF
--- a/db/sql/sql.go
+++ b/db/sql/sql.go
@@ -722,7 +722,10 @@ func MakeColumns(s *schema.Schema, tableName string, fields []string, join bool)
 	var cols []string
 	for _, property := range s.Properties {
 		if property.RelationProperty != "" && join {
-			relatedSchema, _ := manager.Schema(property.Relation)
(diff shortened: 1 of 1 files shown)


db/sql/sql.go
```diff
@@ -722,7 +722,10 @@ func MakeColumns(s *schema.Schema, tableName string, fields []string, join bool)
 	var cols []string
 	for _, property := range s.Properties {
 		if property.RelationProperty != "" && join {
-			relatedSchema, _ := manager.Schema(property.Relation)
+			relatedSchema, ok := manager.Schema(property.Relation)
+			if !ok {
+				panic(fmt.Sprintf("missing schema %s", property.Relation))
```
przemyslaw-dobrowolski-cl: i'd prefer message:
MakeColumns: missing schema %s
so that it informs exactly what is wrong

przemyslaw-dobrowolski-cl: ok, stack is also informative

PR #589, opened 2017-11-23
Pretty printing of Maybe* types

p-kozlowski:


extension/goext/types.go
```diff
@@ -90,6 +94,19 @@ func (m Maybe) HasValue() bool {
...
 }
 
+func (m Maybe) String() string {
+	switch m.MaybeState {
+	case MaybeNull:
+		return "<null>"
+	case MaybeUndefined:
+		return "<undefined>"
+	case MaybeValue:
+		return "<valid value>"
```
przemyslaw-dobrowolski-cl: why not just valid?

extension/goplugin/types_test.go
```diff
@@ -18,6 +18,8 @@ package goplugin_test
 import (
 	"encoding/json"
 
+	"fmt"
+
```
przemyslaw-dobrowolski-cl: unnecessary enter, gofmt

extension/goext/types.go
zimnx: Maybe(String|Int|Float|Bool) also implement String(). They will print 'just value' if they have it. Maybe type doesn't know anything about value, only state.

extension/goplugin/types_test.go
p-kozlowski: fixed

extension/goext/types.go
p-kozlowski: fixed

PR #593, opened 2017-12-06
[ESI-16064] print error stack traces in the logs

p-kozlowski:


extension/goext/error_test.go
```diff
@@ -71,21 +73,85 @@ var _ = Describe("Error", func() {
...
+				return goext.NewErrorBadRequest(middleFunc())
+			}
+
+			_, _, line, _ := runtime.Caller(0)
+			callLine := line + 2
+			err := outerFunc()
+			Expect(err.ErrorStack()).To(ContainSubstring(fmt.Sprintf("error_test.go:%d", innerLine)))
+			Expect(err.ErrorStack()).To(ContainSubstring(fmt.Sprintf("error_test.go:%d", middleLine)))
+			Expect(err.ErrorStack()).To(ContainSubstring(fmt.Sprintf("error_test.go:%d", outerLine)))
+			Expect(err.ErrorStack()).To(ContainSubstring(fmt.Sprintf("error_test.go:%d", callLine)))
```
morrisson: Also should we check BadGateway here?

p-kozlowski: good idea. also a similar assertion added to `Should capture full stack for built-in errors`

extension/goext/error_test.go
```diff
@@ -71,21 +73,90 @@ var _ = Describe("Error", func() {
...
+
+			Expect(err.ErrorStack()).To(ContainSubstring("Bad Gateway"))
+			Expect(err.ErrorStack()).To(ContainSubstring("Bad Request"))
```
morrisson: Maybe I should give a comment in some previous PR properly, but I think the logic which HTTP Error code can be wrapped with other HTTP Error is not proper way to handle internal errors. HTTP error code is used to explain the error reason to clients, so it's not for internal error management. Some operators should be confused in this error message.

I think we should prohibit to wrap goext.Error by goext.Error. Or, at least we should skip the procedure to wrap goext.Error by type check with warning message.

p-kozlowski: For the top level handlers this should be easy and it also sounds good to decorate the error with an HTTP code at the very last moment.

On the other hand, let's consider a scenario of nested handlers scenario. Handler1 calls TriggerEvent (directly or indirectly eg. through CRUD operation on SomeDependentResource). Another handler, Handler2 is invoked. In this case, Handler2 should return a plain error (because goext.Error can't be wrapped in another goext.Error).

But Handler2 can be invoked as a top level handler through CRUD API calls on SomeDependentResource. Now it should return a goext.Error (because it is top level). What's your idea how to solve this issue?

What's more, I think we should not prohibit wrapping. I can easily imagine a case when an error from a dependent handler should be translated to a different error in an upper layer. A recent example is translating 404 Not Found to 400 Bad Request.

As a middle ground solution we could:
1. Allow wrapping
2. Let the extensions decide if they want to wrap or not
3. Return just the top level goext.Error in the API
4. Print the error stack (+ the stack trace) in the debug logs.

@morrisson what's your opinion?

p-kozlowski: @morrisson any updates?

PR #594, opened 2017-12-08
[ESI-16119][ESI-16077] Goext: Log messages for disconnect and timeout

p-kozlowski:


extension/goplugin/environment.go
```diff
@@ -482,26 +479,73 @@ func handleEventForEnv(env IEnvironment, event string, requestContext map[string
...
+	event          string
+	requestContext map[string]interface{}
+	done           chan struct{}
+	ctx            context.Context
+	cancel         context.CancelFunc
+}
+
+func newInterrupt(env IEnvironment, event string, requestContext map[string]interface{}) *interrupt {
+	ctx, cancel := context.WithCancel(context.Background())
+	interrupt := &interrupt{env, event, requestContext, make(chan struct{}, 1), ctx, cancel}
```
zimnx: use named field initialization

p-kozlowski: as discussed offline: only the `done` channel field lacks a proper name, fixed

PR #600, opened 2018-01-02
Added counting function to Transaction

michalzurawski:


extension/goplugin/schemas_test.go
```diff
@@ -197,6 +197,16 @@ var _ = Describe("Schemas", func() {
 			Expect(&createdResource).To(Equal(returnedResource))
 		})
 
+		It("Should count resources", func() {
+			c, err := testSchema.Count(goext.Filter{}, context)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(c).To(Equal(uint64(0)))
+			Expect(testSchema.CreateRaw(&createdResource, context)).To(Succeed())
+			c, err = testSchema.Count(goext.Filter{}, context)
```
p-kozlowski: for completeness let's add a case when the filter matches only a part of resources (eg. 5 resources in DB, the filter matches 3 of those)

db/file/file.go
```diff
@@ -315,6 +315,11 @@ func (tx *Transaction) LockListContext(_ context.Context, s *schema.Schema, filt
 	return tx.LockList(s, filter, options, pg, policy)
 }
 
+func (tx *Transaction) CountContext(_ context.Context, s *schema.Schema, filter transaction.Filter) (uint64, error) {
+	_, total, err := tx.List(s, filter, nil, nil)
```
zimnx: I'm worried about performance of this method. In case of large tables, results of SELECT query may be big. Gohan will need to download all data provided by DB, load into memory and here dump it because it's not needed. I would prefer lightweight solution which will explicitly use COUNT(*), to acquire single row with single value.

zimnx: I'm worried about performance of this method. In case of large tables, results of SELECT query may be big. Gohan will need to download all data provided by DB, load into memory and here dump it because it's not needed. I would prefer lightweight solution which will explicitly use COUNT(*), to acquire single row with single value.

zimnx: Ok I see now that's only for File DB. SQL DB is using COUNT(*).

PR #601, opened 2018-01-04
[ESI-16108] Include module name (schema ID) in the logs

p-kozlowski:


extension/goext/environment.go
```diff
@@ -22,6 +22,7 @@ type IEnvironment interface {
 	Config() IConfig
 	Core() ICore
 	Logger() ILogger
+	LoggerForSchema(schema ISchema) ILogger
```
marcin-ptaszynski: I agree with @morrisson , it's not very clean approach.

p-kozlowski: Let's follow this discussion in Jira

p-kozlowski:
Will be done in a different manner in a separate PR

PR #603, opened 2018-01-08
log error cause of begin tx

marcin-ptaszynski:


db/db.go
```diff
@@ -78,7 +78,7 @@ func withinTxImpl(db DB, beginStrategy func(db DB) (transaction.Transaction, err
 		tx, err = beginStrategy(db)
 
 		if err != nil {
-			log.Warning("failed to begin scoped transaction")
+			log.Warning("failed to begin scoped transaction: %s", err)
```
p-kozlowski: There are also [other places](https://github.com/cloudwan/gohan/blob/master/extension/goplugin/database.go#L64) where transactions are started and the error may potentially not be logged.
I think we should either print this log message everywhere BeginTx (and Begin, too) is called, or print this log message directly in Begin()/BeginTx()

marcin-ptaszynski: thanks @p-kozlowski , I think goplugin tx handling and other require refactor due to code duplication and changed convention. I'd like to include this small change only to help with debugging.

PR #605, opened 2018-01-09
[ESI-13967] More logs and metrics

p-kozlowski:
@yudai can you have a look on the ETCD part and share your ideas? Can you think of any additional metrics we may want to add?

### DIFF
--- a/db/sql/sql.go
+++ b/db/sql/sql.go
@@ -39,6 +39,7 @@ import (
 	"github.com/cloudwan/gohan/db/pagination"
 	"github.com/cloudwan/gohan/db/transaction"
 	"github.com/cloudwan/gohan/extension/goext"
+	"github.com/cloudwan/gohan/metrics"
 	"github.com/cloudwan/gohan/schema"
 	"github.com/cloudwan/gohan/util"
 )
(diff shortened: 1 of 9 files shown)


db/sql/sql.go
```diff
@@ -563,12 +588,18 @@ func (tx *Transaction) logQuery(sql string, args ...interface{}) {
...
+func (tx *Transaction) measureTime(timeStarted time.Time, schemaId, action string) {
+	metrics.UpdateTimer(timeStarted, "tx.%s.%s", schemaId, action)
+}
+
 func (tx *Transaction) Exec(sql string, args ...interface{}) error {
 	return tx.ExecContext(context.Background(), sql, args...)
 }
 
 // Exec executes sql in transaction
 func (tx *Transaction) ExecContext(ctx context.Context, sql string, args ...interface{}) error {
```
p-kozlowski: `ExecContext` may be invoked from other interface methods, resulting in not reliable data. I'll fix it (later) with a small refactoring: public methods shouldn't be invoked from other public methods, so I'll extract the implementation of `ExecContext` into a private `exec` method. The metrics will be gathered only in on the interface level, not in the helper.

yudai:
@p-kozlowski I think it would be great if we can have counters for inflight watches and inflight locks. These can be long running and some resource will be kept for them. So to monitor resource consumption, watching inflight tasks is important.

p-kozlowski:
@yudai agreed. metrics added.

db/sql/sql.go
p-kozlowski: done

yudai:
Oops, one more thing.
I think that simple counters for number of requests are also good to have. We want to see sync/update/fetch rate per sec as well.

p-kozlowski:
@yudai update and fetch (and lock/unlock, too) are using timers, so this data is already available under `[prefix].sync.v3.fetch.count`. I've also updated the demo dashboard (second row)

yudai:
@p-kozlowski I see. that's great :)

PR #606, opened 2018-01-10
ESI-16372 ESI-16552 disable passing cancellable context to tx

marcin-ptaszynski:


extension/goplugin/transaction.go
```diff
@@ -67,7 +67,10 @@ func (t *Transaction) Create(ctx context.Context, s goext.ISchema, resource map[
 	if err != nil {
 		return err
 	}
-	return t.tx.CreateContext(ctx, res)
+	// use context.Background to avoid cancellation mid-query for all Queries/Exec
+	// ESI-16552 context cancellation tends to break next Begin
+	// It doesn't work in mysql driver anyway https://github.com/go-sql-driver/mysql/issues/731
+	return t.tx.CreateContext(context.Background(), res)
```
p-kozlowski: Even though this change is required, it will break the extension execution time out.
The only reasonable solution that comes to my mind is sth like:
```
func (t *Transaction) xxxContext(ctx context.Context, ...) error {
    if err := ctx.Err(); err != nil {
        return err
    }
    return t.tx.xxxContext(context.Background(), ...)
}
```
This way we'll still be able to interrupt long running requests, even though this is just a best-effort solution (ie. won't work in all cases). Reviewers, what do you think about it?

marcin-ptaszynski: Agreed, will do.

marcin-ptaszynski: done

PR #609, opened 2018-01-16
Remove resource from custom action handlers

jkmar:


extension/goext/schemas.go
```diff
@@ -168,8 +168,11 @@ type ISchema interface {
 	// DbDeleteRaw deletes a raw resource, given by a pointer, no events are emitted
 	DbDeleteRaw(filter Filter, context Context) error
 
-	// RegisterEventHandler registers an event handler for a named event with given priority
-	RegisterEventHandler(event string, schemaHandler SchemaHandler, priority int)
+	// RegisterResourceEventHandler registers an event handler with resource for a named event with given priority
+	RegisterResourceEventHandler(event ResourceEvent, schemaHandler SchemaHandler, priority int)
+
+	// RegisterCustomEventHandler registers an event handler without resource for a custom event with given priority
+	RegisterCustomEventHandler(event CustomEvent, handler Handler, priority int)
```
morrisson: Why is this not schemaHandler? Custom action is defined on the schema.

p-kozlowski: In the case of custom actions we specifically do _not_ want to pass the `Resource` parameter (as it will always be null). These methods do the separation: the CRUD actions (that, apart from `show`, provide the resource) go to `RegisterResourceEventHandler` and custom actions (without the resource) go to `RegisterCustomEventHandler`.

Are you referring to the naming of the parameter?

morrisson: Sorry, I overlooked this function is defined in ISchema. That should be okay.

PR #611, opened 2018-01-18
[BL-36] Add filter to fetch

jkmar:


extension/goplugin/schemas.go
```diff
@@ -262,12 +262,28 @@ func (schema *Schema) StateFetchRaw(id string, requestContext goext.Context) (go
 	return tx.StateFetch(goext.GetContext(requestContext), schema, goext.Filter{"id": id})
 }
 
+// FetchFilterRaw returns a pointer to raw resource, containing db annotations
+func (schema *Schema) FetchFilterRaw(filter goext.Filter, requestContext goext.Context) (interface{}, error) {
```
morrisson: I have two comments related to this function.

1. Probably, you added new function here with consideration of backward compatibility. However, I think it's ok to break the compatibility so far.
1. Even if we decide to add new FetchFilterRaw function, we need `id` as a parameter then fill it in goext.Filter .

extension/goplugin/schemas.go
```diff
@@ -281,26 +297,45 @@ func (schema *Schema) fetchImpl(id string, requestContext goext.Context, fetch f
...
 	}
 	xRaw := reflect.ValueOf(fetched)
 	return schema.rawToResource(xRaw)
 }
 
+// Fetch fetches a resource by id.
+// Schema, Logger, Environment and pointer to raw resource are required fields in the resource
+func (schema *Schema) Fetch(id string, context goext.Context) (interface{}, error) {
+	return schema.fetch(func() (interface{}, error) {
+		return schema.FetchRaw(id, context)
```
p-kozlowski: calling a public function from another public function looks suspicious. why don't we extract the common fetch implementation and parametrize it with conver/no convert and lock/no lock policy?

extension/goplugin/schemas.go
jkmar: There are some use cases where both versions could be useful.
If you want to find related object you could filter by id without having to write
```goext.Filter{"id": id}```.
Fetch with filter is more general and can be used instead of List with limit 1.

morrisson:
@jkmar Could you elaborate on the difference between `ListRaw` and `FetchFilterRaw`?

jkmar:
@morrisson 
If there are resources that satisfy given filter `FetchFilterRaw` is equivalent to `ListRaw` with limit 1.
If there are no such resources `FetchFilterRaw` returns an error, `ListRaw` returns an empty list.
The main reason to use filter instead of fetch is that if you need only one resource you don't have to take it from list.

PR #612, opened 2018-01-18
[BL-35] add delete id

jkmar:


morrisson:
Very similar comment to https://github.com/cloudwan/gohan/pull/611 , but do you have any reason to have another `DeleteFilterRaw` function?

extension/goext/schemas.go
```diff
@@ -162,11 +162,17 @@ type ISchema interface {
...
+	// DeleteRaw deletes a raw resource, given by id
+	DeleteRaw(id string, context Context) error
 
-	// DbDeleteRaw deletes a raw resource, given by a pointer, no events are emitted
-	DbDeleteRaw(filter Filter, context Context) error
+	// DbDeleteRaw deletes a raw resource, given by id, no events are emitted
+	DbDeleteRaw(id string, context Context) error
+
+	// DeleteFilterRaw deletes a raw resource, given by a filter
+	DeleteFilterRaw(filter Filter, context Context) error
```
p-kozlowski: `DeleteFilter` family may be very dangerous - calling these functions with an empty filter will delete all resources. It is a valid use case, but I'd rather be safe that sorry.
I was thinking of returning an error/panic()ing when the filter is empty and introducing `DeleteAll(context Context) error` to be explicit about the intent. On the other hand, this could complicate the code when a filter is build dynamically, eg.
```
f := goext.Filter{}
maybeAddSomeFields(f)
maybeAddMoreFields(f)
if len(f) == 0 {
    schema.DeleteAll()
} else {
    schema.DeleteFilterRaw()
}
```
Reviewers, what are your opinions about it?

zimnx: As discussed offline, IMO user should explicitly tell what he wants to delete. So returning error on empty filter is expected behavior.

PR #613, opened 2018-01-23
Add Limit 1 to fetch

jkmar:
added limit 1 to fetch and fixed out of range panic in file.go

### DIFF
--- a/db/file/file.go
+++ b/db/file/file.go
@@ -207,6 +207,13 @@ func (tx *Transaction) Delete(s *schema.Schema, resourceID interface{}) error {
 	return nil
 }
 
+func min(a, b uint64) uint64 {
+	if a < b {
+		return a
+	}
+	return b
+}
+
 type byPaginator struct {
 	data []*schema.Resource
 	pg   *pagination.Paginator
@@ -303,7 +310,7 @@ func (tx *Transaction) List(s *schema.Schema, filter transaction.Filter, options
 		if pg != nil {
 			sort.Sort(byPaginator{list, pg})
 			if pg.Limit > 0 {
-				list = list[:pg.Limit]
+				list = list[:min(pg.Limit, uint64(len(list)))]
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


db/file/file.go
```diff
@@ -303,7 +310,7 @@ func (tx *Transaction) List(s *schema.Schema, filter transaction.Filter, options
 		if pg != nil {
 			sort.Sort(byPaginator{list, pg})
 			if pg.Limit > 0 {
-				list = list[:pg.Limit]
+				list = list[:min(pg.Limit, uint64(len(list)))]
```
michalzurawski: please add a test to this scenario

morrisson:
I want to include @nati and @yudai as reviewers. I'm wondering if gohan should return error if more than one results are returned rather than add a limit one.

jkmar:
@morrisson 
I didn't change the behavior of any of these functions. I think that returning error when more resources are fetched would break the backward compatibility .

morrisson:
@jkmar My point is that current fetch implementation is always called with id filter. 'id' property is always unique in gohan schema, so fetch function is **implicitly** assured that the number of results will be 0 or 1.

However, with the fix #611 , `FetchFilter` function potentially has more than one result with the flexibility of the filter. This means, the result of `FetchFilter` may be different even in the same transaction. This may occur the unexpected error in the extension logic.

Therefore, I believe `FetchFilter` must be used to the combination of properties where the number of results will be 0 or 1, IMO. More concretely, each property must have unique attribute in JSON schema.

FYI, fetch you implemented is different from SQL's fetch definition https://dev.mysql.com/doc/refman/5.7/en/fetch.html . I think this difference itself is ok because Gohan is not SQL system.

jkmar:
@morrisson I think that fetch should succeed only if there is exactly one matching resource. This could be implemented either by changing the limit to 2 or by changing decodeRows for fetch to use next only 2 times. What do you think about this solution?

morrisson:
The direction looks good, but could you elaborate on what you mean "by changing decodeRows for fetch to use next only 2 times"?

jkmar:
@morrisson 
Currently `FetchContext ` in `sql.go` uses `ListContext` which on `decodeRows` calls `Next` on `rows` until all rows are read. We could change that for Fetch, so that it only asks for two rows.

PR #614, opened 2018-01-26
[ESI-16750] defer rollback in withinDetached

jkmar:


extension/goext/database.go
```diff
@@ -67,13 +67,25 @@ func withinJoinable(tx ITransaction, fn func(tx ITransaction) error) error {
...
+
 func withinDetached(db IDatabase, context Context, txBegin func() (ITransaction, error), fn func(tx ITransaction) error) error {
 	opts := db.Options()
 	retryTxCount := opts.RetryTxCount
 	retryTxInterval := opts.RetryTxInterval
 	var tx ITransaction
 	var err error
 
+	defer func() {
+		closeTx(tx)
```
morrisson: for safety, `tx.Closed` should be checked?

jkmar: @morrisson 
I've added check in closeTx.

extension/goext/database.go
```diff
@@ -67,13 +67,28 @@ func withinJoinable(tx ITransaction, fn func(tx ITransaction) error) error {
...
+}
+
 func withinDetached(db IDatabase, context Context, txBegin func() (ITransaction, error), fn func(tx ITransaction) error) error {
 	opts := db.Options()
 	retryTxCount := opts.RetryTxCount
 	retryTxInterval := opts.RetryTxInterval
 	var tx ITransaction
 	var err error
 
+	defer func() {
```
marcin-ptaszynski: This logic is weird. IMO, we shouldn't defer `tx.Close` in root function, but right after tx creation (for loop iteration, e.g. by wrapping for loop iteration body in `func() error`) and doing defer there. Current pitfalls:
- TX is nil, because it failed to create - this code will panic on nil pointer dereference
- some TXes might remain after between retries if err checking logic is insufficient. Not this case, but it's really easy to break.

p-kozlowski: Good points. In fact we've got a logic duplication with https://github.com/cloudwan/gohan/blob/master/db/db.go#L67 and we're fixing the same issues twice. In my opinion we should extract the common part.

p-kozlowski: Good points. In fact we've got a logic duplication with https://github.com/cloudwan/gohan/blob/master/db/db.go#L67 and we're fixing the same issues twice. In my opinion we should extract the common part.

jkmar:
@marcin-ptaszynski @morrisson @p-kozlowski 
I've moved goext.Within to database in goplugin and removed duplication (within now uses  within from db). I've also refactored that function - it now commits a transaction if fn was successful, so commit in fn is no longer needed (we didn't need commit in goext within).

marcin-ptaszynski:
@jkmar , awesome, LGTM.

PR #615, opened 2018-01-26
[BL-56] Add schema names to converter

jkmar:
Converter now generates const with schema ids.
Additionally schema id type is changed from string to separate type "SchemaID"

### DIFF
--- a/extension/goext/schemas.go
+++ b/extension/goext/schemas.go
@@ -20,6 +20,9 @@ import (
 	"errors"
 )
 
+// SchemaID is a type for schema ID
+type SchemaID string
+
 // LockPolicy indicates lock policy
 type LockPolicy int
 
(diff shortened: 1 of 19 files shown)


morrisson:
What is "schema names"?

jkmar:
@morrisson 
Schema names was list of schema ids, but I've changed this pr to include SchemaID as separate type.

extension/goext/schemas.go
```diff
@@ -20,6 +20,9 @@ import (
 	"errors"
 )
 
+// SchemaID is a type for schema ID
+type SchemaID string
```
marcin-ptaszynski: What's the benefit of creating new type in this case?

p-kozlowski: type safety

jkmar:
@marcin-ptaszynski 
Could you take a look at this pr again?

PR #617, opened 2018-01-30
Added OR condition to policy

michalzurawski:
This commit provides additional filters to the Schema's policy - it adds "AND" and "OR" operator which can be nested. Unlike the previous filters from policy, this one does affect the query. Currently, only equal operator is provided for comparison, however, this could be easily extended in future. Moreover this changes might be used to extend other SQL methods.

### DIFF
--- a/db/sql/sql_test.go
+++ b/db/sql/sql_test.go
@@ -27,6 +27,7 @@ import (
 	. "github.com/cloudwan/gohan/db/sql"
 	"github.com/cloudwan/gohan/db/transaction"
 	"github.com/cloudwan/gohan/schema"
+	"github.com/lann/squirrel"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
@@ -36,9 +37,10 @@ var _ = Describe("Sql", func() {
 	const testFixtures = "test_fixture.json"
 
 	var (
-		conn    string
-		tx      transaction.Transaction
-		sqlConn *DB
+		conn       string
+		tx         transaction.Transaction
+		sqlConn    *DB
+		testSchema *schema.Schema
 	)
 
 	BeforeEach(func() {
@@ -65,6 +67,11 @@ var _ = Describe("Sql", func() {
 
 		tx, err = dbc.Begin()
 		Expect(err).ToNot(HaveOccurred())
+
+		var ok bool
+		testSchema, ok = manager.Schema("test")
+		Expect(ok).To(BeTrue())
+		Expect(testSchema).ToNot(BeNil())
 	})
 
 	AfterEach(func() {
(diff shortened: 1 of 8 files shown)


server/server_test.go
```diff
@@ -479,6 +479,59 @@ var _ = Describe("Server package test", func() {
...
+			expectedToContainTest(4, res)
+			res = testURL("GET", testPluralURL, powerUserTokenID, nil, http.StatusOK)
+			expectedToContainTest(3, res)
+
+			testURL("PUT", testPluralURL+"/"+testUpLevel0ID, memberTokenID, testUpdate, http.StatusOK)
+			res = testURL("GET", testPluralURL, memberTokenID, nil, http.StatusOK)
+			expectedToContainTest(4, res)
+			res = testURL("GET", testPluralURL, powerUserTokenID, nil, http.StatusOK)
+			expectedToContainTest(4, res)
+		})
```
morrisson: This is testing the scenario "tenant A resources can be read tenant A and other tenants if conditions match".

Can current implementation cover another scenario "some tenant A resources can be read by tenant A but cannot be updated or deleted even by tenant A (owner) if some conditions match"? This is admin-managed user resources scenario.

michalzurawski: Good idea - added. I've also added `not equal` operator.
For more operators, we have to change our squirrel vendor to https://github.com/Masterminds/squirrel - `lann`'s project is deprecated.

db/sql/sql_test.go
```diff
@@ -353,8 +360,189 @@ var _ = Describe("Sql", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				resSql, param, err := res.ToSql()
+				Expect(err).ToNot(HaveOccurred())
+
+				expectedQuery = expectedQuery.Where(squirrel.And{squirrel.Eq{"`test_string`": "123"}, squirrel.Eq{"`test_number`": 42}})
+				expectedSql, expectedParam, _ := expectedQuery.ToSql()
+				Expect(resSql).To(Equal(expectedSql))
+				Expect(param).To(Equal(expectedParam))
+			})
+			It("should create disjunction of conjunctions for more then one parameter", func() {
```
p-kozlowski: how does operator precedence work?
for example what does `filter := {"test_string": "123", "test_number": 42, "__or__": {{"property": ...}}` produce? is it
`(test_string = 123 AND test_number = 42) OR (...)` (which makes sense) or `(test_string = 123) AND (test_number = 42 OR (...))` (which does not to make sense due to Golangs map ordering)?
please add a test case for that

db/sql/sql_test.go
```diff
@@ -353,8 +360,189 @@ var _ = Describe("Sql", func() {
...
+
+				Expect(err).ToNot(HaveOccurred())
+				resSql, param, err := res.ToSql()
+				Expect(err).ToNot(HaveOccurred())
+				expectedSql, expectedParam, _ := expectedQuery.ToSql()
+				Expect(resSql).To(Equal(expectedSql))
+				Expect(param).To(Equal(expectedParam))
+			})
+			It("should create select query with one parameter", func() {
+				filter := map[string]interface{}{"test_string": "123"}
```
p-kozlowski: is it possible to express `test_string != "123"` using the new API? how to do it? do we need to do sth like `{"__and__": {"property":"test_string", "type":"neq", "value":123}}`?
please add a test case for it

db/sql/sql_test.go
michalzurawski: none of this true.
in your example `filter := {"a": 1, "b": 2, "__or__": {{"property": "c", "value":3, "type":"eq"},{"property": "d", "value":4, "type":"neq"}}` it would produce `a = 1 AND b = 2 AND (c=3 OR d <> 4)`.
If you want `(a=1 and b=2) OR (c=3 OR d<> 4)` you just have to nest one operator in another i.e. 
```filter := {
"__or__":
  {"__and__":{
    {"property": "a", "value" 1, "type":"eq"},
    {"property": "b", "value" 2, "type":"eq"},
  },
  "__or__": {
    {"property": "c", "value":3, "type":"eq"},
    {"property": "d", "value":4, "type":"neq"}
  },
}
```

server/resources/resource_management.go
```diff
@@ -435,15 +436,25 @@ func CreateOrUpdateResource(
...
-		_, fetchErr := preTransaction.Fetch(resourceSchema, filter, nil)
-		exists = fetchErr == nil
+		_, fetchErr = preTransaction.Fetch(resourceSchema, filter, nil)
+		if fetchErr != nil {
+			return fetchErr
+		}
 		return nil
 	}); preTxErr != nil {
-		return false, preTxErr
+		return false, ResourceError{preTxErr, preTxErr.Error(), Unauthorized}
```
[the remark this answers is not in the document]
michalzurawski: @p-kozlowski I understand your concern - but could you please propose any other error code which should be returned? The problem is `PUT` behaves as both `UPDATE` and `CREATE` - when one tenant tries to create an object using `PUT` with `ID` which have been already used by another tenant our API will try to update existing object which is not possible.
And, of course, this tenant cannot create an object with ID which already exists in DB even when it was created by another tenant.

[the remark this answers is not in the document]
p-kozlowski: this answers the tenant isolation case, perhaps the DB architecture is broken - the primary key should be (id, tenant_id) pair. but solving this is definitely out of scope of this PR. taking this into consideration, I think we could live with 401 Unauthorized.
point 1 (general DB errors) remains

server/resources/resource_management.go
```diff
@@ -435,15 +436,25 @@ func CreateOrUpdateResource(
...
+		exists = true
+
 		if tenantIDs != nil {
 			filter["tenant_id"] = tenantIDs
 		}
+		policy.AddCustomFilters(filter, auth.TenantID())
 
-		_, fetchErr := preTransaction.Fetch(resourceSchema, filter, nil)
-		exists = fetchErr == nil
+		_, fetchErr = preTransaction.Fetch(resourceSchema, filter, nil)
```
p-kozlowski: inefficiency: is it possible to stay with 1 SQL query instead of 2?

michalzurawski: true - previous version with one select was slightly incorrect as discussed with previous comments - when one tenant created resource with ID "A" and the second tries to create resource with same ID using PUT it would get DB error that ID already exists in DB.
We can have just SELECT w/o WHERE clause at all and then manually check if tenant ID is correct - however, it will require a lot of changes to do the same with other properties defined with "OR" or "AND".

I can instead swap this select statements - in a typical case when tenant updates own resource there would be just one select. Only when a tenant would try to create resource with PUT or would try to update resource which does not belong to him there would be two selects.

db/sql/sql_test.go
michalzurawski: exactly, added test

michalzurawski:
@p-kozlowski @morrisson  i've applied changes and rebased

michalzurawski:
@morrisson would you kindly take a second look at this PR?

PR #619, opened 2018-02-13
fix index sql stmt execution

marcin-ptaszynski:


morrisson:
One test case is needed.

PR #620, opened 2018-02-20
Fixed error message for converter

michalzurawski:


cli/cli.go
```diff
@@ -508,7 +508,7 @@ ARGUMENTS:
 				RawSuffix:        c.String("raw-suffix"),
 				InterfaceSuffix:  c.String("interface-suffix"),
 			}); err != nil {
-				println(err)
+				fmt.Printf("%+v\n", err)
```
zimnx: nit: %s is more appropriate since error interface requires String() method.

PR #622, opened 2018-03-01
[BL-103] add support for nested slice conversion

marcin-ptaszynski:


extension/goplugin/util.go
```diff
@@ -309,35 +316,65 @@ func (util *Util) ResourceToMap(resource interface{}) map[string]interface{} {
...
+		v = v.Elem()
 	}
+	if v.Kind() == reflect.Map {
+		return v.Interface().(map[string]interface{})
+	} else {
+		return util.ResourceToMap(v.Interface())
+	}
+}
 
+func (util *Util) sliceToMap(v reflect.Value) interface{} {
```
p-kozlowski: How about:
```
func (util *Util) sliceToMap(v reflect.Value) interface{} {
   if isPrimitiveKind() {
       return convertPrimitiveKind()
   } else if isNestedSlice() {
       return convertNestedSlice()
   } else {
       return convertMappable()
   }
}
```

marcin-ptaszynski: more or less I've extracted code as requested.

PR #623, opened 2018-03-02
[BL-94] Added hiddenProperties for resource in action policy

jkmar:
Now it is possible to specify, which fields are forbidden for given action.

### DIFF
--- a/docs/policy.md
+++ b/docs/policy.md
@@ -202,3 +202,42 @@ policies:
 In the above example, the access to favicon is always granted and never requires an authorization.
 This feature is useful for web browsers and it is a good practice to set this policy.
 In the second policy, no-authorization access is granted to all member resources defined by a path wildcard.
+
+## Resource properties
+
+It is possible to filter allowed fields using `properties` or `blacklistProperties` properties.
+It is not possible to use both `properties` and `blacklistProperties` at the same time.
+In case of having both types of properties error is returned on the server start.
+- `properties` defines properties exposed to the user, the other properties are forbidden.
+- `blacklistProperties` defines forbidden properties, the other properties are exposed.
+
+```yaml
+- action: read
+  effect: allow
+  id: visible_properties_read
+  principal: Visible
+  condition:
+  - type: property
+    match:
+      is_public:
+      - true
+  resource:
+    path: /v2.0/visible_properties_test.*
+    properties:
+    - a
+- action: read
+  effect: allow
+  id: visible_properties_read
+  principal: Hidden
+  condition:
+  - type: property
+    match:
+      is_public:
+      - true
+  resource:
+    path: /v2.0/visible_properties_test.*
+    blacklistProperties:
+    - id
+    - a
+    - is_public
+```
--- a/schema/policy.go
+++ b/schema/policy.go
@@ -116,8 +116,8 @@ type Policy struct {
 
 //ResourcePolicy describes target resources
 type ResourcePolicy struct {
-	Properties []interface{}
-	Path       *regexp.Regexp
+	PropertiesFilter *Filter
+	Path             *regexp.Regexp
 }
 
 //Authorization interface
@@ -251,15 +251,39 @@ func NewPolicy(raw interface{}) (*Policy, error) {
 		return nil, fmt.Errorf(onlyOneOfTenantIDTenantNameError)
 	}
 
-	properties, ok := resourceData["properties"]
-	resource.Properties = nil
-	if ok {
-		resource.Properties = properties.([]interface{})
+	filterFactory := FilterFactory{}
+	if resource.PropertiesFilter, err = filterFactory.CreateFilterFromProperties(
+		getStringSliceFromMap(resourceData, "properties"),
+		getStringSliceFromMap(resourceData, "blacklistProperties"),
+	); err != nil {
+		return nil, err
 	}
+
 	if err := policy.precomputeConditions(); err != nil {
 		return nil, err
 	}
 	return policy, nil
+}
+
+func getStringSliceFromMap(data map[string]interface{}, key string) []string {
+	switch slice := data[key].(type) {
+	case []string:
+		return slice
+	case []interface{}:
+		return getStringSliceFromRawSlice(slice)
+	default:
+		return nil
+	}
+}
+
+func getStringSliceFromRawSlice(data []interface{}) []string {
+	var result []string
+	for _, rawItem := range data {
+		if item, ok := rawItem.(string); ok {
+			result = append(result, item)
+		}
+	}
+	return result
 }
 
 func (p *Policy) precomputeConditions() error {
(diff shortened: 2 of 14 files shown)


schema/property_filter.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	return result
+}
+
+func (f *Filter) IsForbidden(key string) bool {
+	return !f.filterStrategy.Filter(key)
+}
+
+func (f *Filter) AllowsAll() bool {
+	_, ok := f.filterStrategy.(*includeAllFilter)
```
zimnx: closing to initial implementation. Also AllowsAll should take input and validates whether all passes otherwise I don't understand what is should do.

schema/policy.go
```diff
@@ -251,17 +251,33 @@ func NewPolicy(raw interface{}) (*Policy, error) {
...
+
 	if err := policy.precomputeConditions(); err != nil {
 		return nil, err
 	}
 	return policy, nil
 }
 
+func getStringSliceFromMap(data map[string]interface{}, key string) []string {
+	var result []string
+	rawSlice, ok := data[key].([]interface{})
```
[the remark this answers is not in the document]
jkmar: Why? In previous version it was only []interface{}.

schema/property_filter.go
jkmar: AllowsAll checks if given filter allows all keys it is only possible for includeAllFilter.

zimnx: no, this function checks whether given filter is of type includeAllFIlter. Name doesn't explain semantic.

jkmar: There is only one usage of this function: in `Check` in `policy.go` to write a debug message.
Checking all properties here would be redundant, since they are checked again anyway, so I'd rather removed this function completely and also the debug message. What do you think about this solution?

schema/property_filter_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Property filter tests", func() {
+	var filterFactory = &FilterFactory{}
+
+	Describe("FilterFactory tests", func() {
```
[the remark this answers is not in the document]
jkmar: done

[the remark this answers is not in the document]
jkmar: done

schema/property_filter.go
```diff
@@ -0,0 +1,133 @@
...
+	return ok
+}
+
+type FilterStrategy interface {
```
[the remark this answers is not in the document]
jkmar: done

schema/property_filter.go
```diff
@@ -0,0 +1,133 @@
...
+func (f *FilterFactory) createFilterStrategy() (FilterStrategy, error) {
+	switch f.getFilterType() {
+	case All:
+		return &includeAllFilter{}, nil
+	case Visible:
+		return &visibleFilter{properties: sliceToSet(f.visible)}, nil
+	case Hidden:
+		return &hiddenFilter{properties: sliceToSet(f.hidden)}, nil
+	default:
+		return nil, errors.New("Cannot have filter with both visible and hidden properties")
```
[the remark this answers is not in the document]
jkmar: done

schema/policy.go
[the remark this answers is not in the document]
zimnx: just to be precautious

schema/property_filter.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	return result
+}
+
+func (f *Filter) IsForbidden(key string) bool {
+	return !f.predicate.Predicate(key)
+}
+
+type Predicate interface {
+	Predicate(string) bool
```
zimnx: Validate?

schema/policy.go
```diff
@@ -451,16 +419,11 @@ func (p *Policy) Check(action string, authorization Authorization, data map[stri
 		}
 	}
 
-	properties := p.Resource.Properties
-	if properties == nil {
```
zimnx: now this logic is missing

schema/policy.go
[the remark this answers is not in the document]
jkmar: done

schema/property_filter.go
jkmar: done

schema/policy.go
jkmar: I've restored this logic, but in my opinion this is unnecessary and only makes code uglier, since this check was only for debug purposes.

jkmar:
@zimnx I've applied all the comments please review again.

schema/property_filter.go
zimnx: Branch we are talking about is not only displaying log message, but also does an early return. I just don't want to rely on includeAllPredicate implementation. AllowsAll should not rely on type casting and believing on that Validate method always returns true. You can remove this early check - will cause small performance drop or change interface to delegate all/any/none predicate checking to predicate. In this solution you will be able preserve function complexity by implementing this semantic to these interface methods. For example includeAllPredicate::All(properties) would always return true no matter what is passed to function.

docs/policy.md
```diff
@@ -202,3 +202,41 @@ policies:
 In the above example, the access to favicon is always granted and never requires an authorization.
 This feature is useful for web browsers and it is a good practice to set this policy.
 In the second policy, no-authorization access is granted to all member resources defined by a path wildcard.
+
+## Resource properties
+
+It is possible to filter allowed fields using `properties` or `hiddenProperties` properties.
+It is not possible to use both `properties` and `hiddenProperties` at the same time.
+- `properties` defines visible properties, the other properties are hidden.
+- `hiddenProperties` defines hidden properties, the other properties are visible.
```
marcin-ptaszynski: IMO, name `hiddenProperties` is confusing and works only with `read`. Consider changing it to `blacklistProperties` or maybe something else.

server/server_test.go
```diff
@@ -479,6 +480,45 @@ var _ = Describe("Server package test", func() {
...
+				testResource := map[string]interface{}{
+					"id": resourceID,
+					"a":  "a",
+					"b":  "b",
+				}
+				testURL("POST", visibilityTestPluralURL, adminTokenID, testResource, http.StatusCreated)
+			})
+
+			It("Should see visible fields", func() {
+				res := testURL("GET", url, visibleTokenID, nil, http.StatusOK)
```
p-kozlowski: please add tests for create and update when `hiddenProperties` are non-empty in create and update policies

schema/policy.go
jkmar: I've run some performance tests to confirm that removing that logic doesn't hurt the performance. More details in the comments for this issue.

docs/policy.md
jkmar: I've change the name to `blacklistProperties`.

server/server_test.go
jkmar: I've added tests for create and update.

schema/property_filter.go
jkmar: Please check the comment about removing this logic.

server/server_test.go
jkmar: @marcin-ptaszynski @zimnx @p-kozlowski I've applied all the comment please review again.

jkmar:
@marcin-ptaszynski @zimnx @p-kozlowski I've applied all the comment please review again.

docs/policy.md
```diff
@@ -202,3 +202,41 @@ policies:
 In the above example, the access to favicon is always granted and never requires an authorization.
 This feature is useful for web browsers and it is a good practice to set this policy.
 In the second policy, no-authorization access is granted to all member resources defined by a path wildcard.
+
+## Resource properties
+
+It is possible to filter allowed fields using `properties` or `blacklistProperties` properties.
+It is not possible to use both `properties` and `blacklistProperties` at the same time.
```
morrisson: If an user wrongly uses both at the same time, what will happen? Error happens on server start, or either property is preferred?

server/server_test.go
```diff
@@ -479,6 +481,79 @@ var _ = Describe("Server package test", func() {
...
+					testResource := map[string]interface{}{
+						"id":     resourceID,
+						property: property,
+					}
+					testURL("POST", visibilityTestPluralURL, token, testResource, status)
+				},
+				Entry("Should create exposed with correct fields", "a", visibleTokenID, http.StatusCreated),
+				Entry("Should not create exposed with incorrect fields", "b", visibleTokenID, http.StatusUnauthorized),
+				Entry("Should create forbidden with correct fields", "b", hiddenTokenID, http.StatusCreated),
+				Entry("Should not create forbidden with correct fields", "a", hiddenTokenID, http.StatusUnauthorized),
```
p-kozlowski: incorrect?

jkmar: I've fixed that, also the descriptions for the other tests were also incorrect.

docs/policy.md
jkmar: Error happens on the server start. I've updated the docs to contain that information.

jkmar: I've also added test for that case.

PR #626, opened 2018-03-13
Extracted GetHttpRequestBody to goplugin::Util

zimnx:
Also changed ReadJSON method to not consume request body.
Duplicated buffer will be stored under request to allow multiple reads from it

### DIFF
--- a/extension/goplugin/util.go
+++ b/extension/goplugin/util.go
@@ -20,8 +20,11 @@ import (
 	"reflect"
 	"strings"
 
+	"net/http"
+
 	"github.com/cloudwan/gohan/db/transaction"
 	"github.com/cloudwan/gohan/extension/goext"
+	gohan_util "github.com/cloudwan/gohan/util"
 	"github.com/golang/mock/gomock"
 	"github.com/jmoiron/sqlx/reflectx"
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


extension/goplugin/util.go
```diff
@@ -56,6 +59,11 @@ func (util *Util) GetTransaction(context goext.Context) (goext.ITransaction, boo
 	return contextGetTransaction(context)
 }
 
+func (util *Util) GetHttpRequestBody(context goext.Context) (map[string]interface{}, error) {
+	httpRequest := context["http_request"].(*http.Request)
```
morrisson: Is proper error handling needed?

util/util.go
```diff
@@ -355,3 +356,20 @@ func Index(slice []string, elem string) int {
...
 	return -1
 }
+
+//ReadJSON reads JSON from http request
+func ReadJSON(r *http.Request) (map[string]interface{}, error) {
+	var data interface{}
+	var reqData bytes.Buffer
+	tee := io.TeeReader(r.Body, &reqData)
+	decoder := json.NewDecoder(tee)
+	r.Body = ioutil.NopCloser(&reqData)
```
p-kozlowski: looks fishy: we're modifying the requests internal state in a utility function. instead I'd rather store (a copy of) the `Body` in the `context`. as a result you wouldn't need to provide `GetHttpRequestBody` in `goext`

zimnx: You already keep request inside context, lets not duplicate this buffer. Reading from r.Body is changing internal state since r.Body is a stream which is consumed upon read. This function gives guarantee that r.Body will be readable multiple times.

zimnx:
I don't need it anymore. This would also affect design dependency. Gohan tries to separate HTTP server from extensions by handling HTTP stuff on his own way. In this PR we would have multiple request body handlers. Having only single one is IMO better

PR #627, opened 2018-03-14
Added handling of Maybe type assigned to interface{}

michalzurawski:


extension/goplugin/util.go
```diff
@@ -431,6 +431,9 @@ func setMaybeState(v reflect.Value, state goext.MaybeState) {
 }
 
 func primitiveMaybeToMap(fieldsMap map[string]interface{}, fieldName string, value reflect.Value) {
+	if value.Kind() == reflect.Interface {
```
p-kozlowski: even though it's good to support all possible cases, honestly I do not think we should use plain `interface{}`s unless absolutely necessary. let's merge this PR, but let's try to avoid usage of `interace{}` in the extensions

PR #628, opened 2018-03-16
CW-125 allow empty list in WHERE ... IN () clauses

p-kozlowski:


db/sql/sql_test.go
```diff
@@ -632,6 +632,68 @@ var _ = Describe("Sql", func() {
 				Expect(resSql).To(Equal(expectedSql))
 				Expect(param).To(Equal(expectedParam))
 			})
+			It("should handle empty list in WHERE ... IN ()", func() {
```
jkmar: Since all of these tests have the same structure:
```
filter := map[string]interface{}{...}

res, err := AddFilterToQuery(testSchema, query, filter, false)

resSql, param, err := res.ToSql()

expectedQuery = expectedQuery.Where(...)
expectedSql, expectedParam, _ := expectedQuery.ToSql()
Expect(resSql).To(Equal(expectedSql))
Expect(param).To(Equal(expectedParam))
```
I would consider grouping them into `DescribeTable`

p-kozlowski: sure, done

morrisson:
Please squash commits before merge

p-kozlowski:
squashed, waiting for CI

PR #629, opened 2018-03-19
Supply API request data to context

zimnx:


zimnx:
Tests added

server/api.go
```diff
@@ -184,10 +184,11 @@ func fillInContext(context middleware.Context, db db.DB,
 	r *http.Request, w http.ResponseWriter,
 	s *schema.Schema, p martini.Params, sync sync.Sync,
 	identityService middleware.IdentityService,
-	queue *job.Queue) {
+	queue *job.Queue, requestData map[string]interface{}) {
 	context["path"] = r.URL.Path
 	context["http_request"] = r
 	context["http_response"] = w
+	context["request_data"] = requestData
```
p-kozlowski: [there are places](https://github.com/cloudwan/gohan/blob/master/server/resources/resource_management.go#L773) where the dataMap is modified, so the content of `context["request_data"]` may be modified after initial assignment. is it OK?
if not, we'd need a deep copy

server/server_test.go
```diff
@@ -896,6 +896,24 @@ var _ = Describe("Server package test", func() {
 			testURL("POST", responderPluralURL, adminTokenID, responder, http.StatusCreated)
 		})
 
+		It("Request data not available in context on GET action", func() {
+			testURL("GET", responderPluralURL+"/r1", adminTokenID, nil, http.StatusOK)
+		})
+
+		It("Request data not available in context on DELETE action", func() {
```
p-kozlowski: please add tests for create and update, too

zimnx: other existing tests covers create and update

server/api.go
zimnx: Deepcopy added

morrisson:
`extension.md` should be updated.

zimnx:
@morrisson I've added documentation.

PR #630, opened 2018-03-23
[CW-123] Fixed error code for deleting resource

michalzurawski:
Return 401 when a tenant can "GET" resource but not delete. 404 when a resource does not exist at all.

### DIFF
--- a/server/resources/resource_management.go
+++ b/server/resources/resource_management.go
@@ -49,6 +49,7 @@ const (
 	CreateFailed
 	UpdateFailed
 	Unauthorized
+	Forbidden
 	ForeignKeyFailed
 )
 
@@ -284,7 +285,6 @@ func GetMultipleResources(context middleware.Context, dataStore db.DB, resourceS
 	if err != nil {
 		return ResourceError{err, err.Error(), WrongQuery}
 	}
-	context["policy"] = policy
 
 	environmentManager := extension.GetManager()
 	environment, ok := environmentManager.GetEnvironment(resourceSchema.ID)
@@ -326,7 +326,6 @@ func GetSingleResource(context middleware.Context, dataStore db.DB, resourceSche
 	if err != nil {
 		return err
 	}
-	context["policy"] = policy
 
 	environmentManager := extension.GetManager()
 	environment, ok := environmentManager.GetEnvironment(resourceSchema.ID)
@@ -452,7 +451,7 @@ func CreateOrUpdateResource(
 			return err
 		}
 		if exists {
-			return ResourceError{transaction.ErrResourceNotFound, "", Unauthorized}
+			return ResourceError{transaction.ErrResourceNotFound, "", Forbidden}
 		}
 		return nil
 	}); preTxErr != nil {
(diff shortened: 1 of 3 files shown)


server/resources/resource_management.go
```diff
@@ -851,9 +851,33 @@ func DeleteResource(context middleware.Context,
...
 		policy.AddCustomFilters(filter, auth.TenantID())
 		resource, fetchErr = preTransaction.Fetch(resourceSchema, filter, nil)
+		if fetchErr != nil {
+			switch fetchErr {
+			case transaction.ErrResourceNotFound:
+				policy, err := loadPolicy(context, "read", strings.Replace(resourceSchema.GetSingleURL(), ":id", resourceID, 1), auth)
+				if err != nil {
+					return err
+				}
+				context["policy"] = policy
```
jkmar: I'm not sure about this line, since context is not used in this function after that assignment.

server/resources/resource_management.go
```diff
@@ -851,9 +851,33 @@ func DeleteResource(context middleware.Context,
...
+				filter := transaction.IDFilter(resourceID)
```
jkmar: There is a duplication (861-865), (848-852)

morrisson:
I do not think `401` is good for this case because the user is authenticated to access for some other resources. `403 forbidden` to this resource should be applied.

michalzurawski:
@morrisson currently when tenant can read resource and tries to update it without proper permission it returns 401. Should i also change it to 403?

server/resources/resource_management.go
```diff
@@ -888,6 +885,28 @@ func DeleteResource(context middleware.Context,
 	return nil
 }
 
+func fetchResourceForAction(action string, auth schema.Authorization, resourceID string, resourceSchema *schema.Schema,
+	tx transaction.Transaction, context middleware.Context) (*schema.Resource, error) {
+	policy, err := loadPolicy(context, action, strings.Replace(resourceSchema.GetSingleURL(), ":id", resourceID, 1), auth)
+	if err != nil {
+		return nil, err
+	}
+	context["policy"] = policy
```
morrisson: It has already been done in `loadPolicy`

server/resources/resource_management.go
```diff
@@ -834,45 +835,41 @@ func DeleteResource(context middleware.Context,
...
 	if errPreTx := db.Within(dataStore, func(preTransaction transaction.Transaction) error {
-		tenantIDs := policy.GetTenantIDFilter(schema.ActionDelete, auth.TenantID())
-		filter := transaction.IDFilter(resourceID)
-		if tenantIDs != nil {
-			filter["tenant_id"] = tenantIDs
+		resource, fetchErr = fetchResourceForAction(schema.ActionDelete, auth, resourceID, resourceSchema, preTransaction, context)
+		if fetchErr != nil {
+			switch fetchErr {
+			case transaction.ErrResourceNotFound:
+				_, err := fetchResourceForAction(schema.ActionRead, auth, resourceID, resourceSchema, preTransaction, context)
```
morrisson: Just an idea. Can we verify "resource exists but no privilege for the action" in one fetch operation? It's just a wasteful cost to access DB just for error code tweak.

server/resources/resource_management.go
michalzurawski: good point it is added to context in few places after loadPolicy same as here - i'll remove all such occurrences

server/resources/resource_management.go
michalzurawski: in theory, we can - we have to fetch resource without any filter and if it exists we can manually check (using reflection) that it meets the policy. in practice, it involves much work to do such check (consider new policy allowing complex conditions with 'or'/'and') and this is not the common case when user tries to delete non-existing resource or resource that he's not allowed to delete.

morrisson: got it

PR #632, opened 2018-04-09
Add cached sql transactions

JDCodi:
Add wrapper for transaction.Transaction that caches select requests results and returns results from cache instead of making another call to the db. All operations that could influence requests results clear the cache, so we make another call to the db when needed.

### DIFF
--- a/db/sql/cached_transaction.go
+++ b/db/sql/cached_transaction.go
@@ -1,0 +1,206 @@
+// Copyright (C) 2018 NTT Innovation Institute, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+// implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sql
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/cloudwan/gohan/db/pagination"
+	"github.com/cloudwan/gohan/db/transaction"
+	"github.com/cloudwan/gohan/schema"
+	// Import mysql lib
+	_ "github.com/go-sql-driver/mysql"
+	// Import go-sqlite3 lib
+	_ "github.com/mattn/go-sqlite3"
+	// Import go-fakedb lib
+	"github.com/cloudwan/gohan/metrics"
+	"github.com/mitchellh/hashstructure"
+	_ "github.com/nati/go-fakedb"
+)
+
+type CachedState struct {
+	list     []*schema.Resource
+	total    uint64
+	isLocked bool
+}
+
+type CachedTransaction struct {
+	TxInterface
+	QueryCache map[string]CachedState
+}
+
+func MakeCachedTransaction(transx TxInterface) TxInterface {
+	cachedTransaction := &CachedTransaction{transx, nil}
+	cachedTransaction.ClearCache()
+	return cachedTransaction
+}
+
+func (tx *CachedTransaction) Create(resource *schema.Resource) error {
+	return tx.CreateContext(context.Background(), resource)
+}
+
+func (tx *CachedTransaction) CreateContext(ctx context.Context, resource *schema.Resource) error {
+	tx.ClearCache()
+	return tx.TxInterface.CreateContext(ctx, resource)
+}
+
+func (tx *CachedTransaction) Update(resource *schema.Resource) error {
+	return tx.UpdateContext(context.Background(), resource)
+}
+
+func (tx *CachedTransaction) UpdateContext(ctx context.Context, resource *schema.Resource) error {
+	tx.ClearCache()
+	return tx.TxInterface.UpdateContext(context.Background(), resource)
+}
+
+func (tx *CachedTransaction) StateUpdate(resource *schema.Resource, state *transaction.ResourceState) error {
+	return tx.StateUpdateContext(context.Background(), resource, state)
+}
+
+func (tx *CachedTransaction) StateUpdateContext(ctx context.Context, resource *schema.Resource, state *transaction.ResourceState) error {
+	tx.ClearCache()
+	return tx.TxInterface.StateUpdateContext(context.Background(), resource, state)
+}
+
+func (tx *CachedTransaction) Delete(s *schema.Schema, resourceID interface{}) error {
+	return tx.DeleteContext(context.Background(), s, resourceID)
+}
+
+func (tx *CachedTransaction) DeleteContext(ctx context.Context, s *schema.Schema, resourceID interface{}) error {
+	tx.ClearCache()
+	return tx.TxInterface.DeleteContext(context.Background(), s, resourceID)
+}
+
+func (tx *CachedTransaction) List(s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions, pg *pagination.Paginator) (list []*schema.Resource, total uint64, err error) {
+	return tx.ListContext(context.Background(), s, filter, options, pg)
+}
+
+//List resources in the db
+func (tx *CachedTransaction) ListContext(ctx context.Context, s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions, pg *pagination.Paginator) (list []*schema.Resource, total uint64, err error) {
+	sc := listContextHelper(s, filter, options, pg)
+
+	list, total, wasCached, _, err := tx.getCached(s.ID, sc)
+
+	if err != nil {
+		return
+	}
+
+	if !wasCached {
+		metrics.UpdateCounter(1, "tx.%s.cache.miss", s.ID)
+		list, total, err = tx.TxInterface.ListContext(ctx, s, filter, options, pg)
+		tx.saveCache(s.ID, sc, list, total, false, err)
+	} else {
+		metrics.UpdateCounter(1, "tx.%s.cache.hit", s.ID)
+	}
+	return
+}
+
+func (tx *CachedTransaction) createKey(schemaID string, sc *selectContext) (string, error) {
+	filterHash, err := hashstructure.Hash(sc.filter, nil)
+	if err != nil {
+		return "", err
+	}
+	res := fmt.Sprintf("%s %v %v %v", schemaID, sc.join, filterHash, sc.fields)
+	if sc.paginator != nil {
+		res = fmt.Sprintf("%s %v", res, *sc.paginator)
+	}
+	return res, err
+}
+
+func (tx *CachedTransaction) getCached(schemeID string, sc *selectContext) (list []*schema.Resource, total uint64, wasCached bool, wasLocked bool, err error) {
+	key, err := tx.createKey(schemeID, sc)
+	if err != nil {
+		return nil, 0, false, false, err
+	}
+
+	res, present := tx.QueryCache[key]
+	if present {
+		return res.list, res.total, true, res.isLocked, nil
+	} else {
+		return nil, 0, false, false, nil
+	}
+}
+
+func (tx *CachedTransaction) saveCache(schemeID string, sc *selectContext, list []*schema.Resource, total uint64, wasLocked bool, errList error) error {
+	if errList != nil {
+		return nil
+	}
+	key, err := tx.createKey(schemeID, sc)
+	if err != nil {
+		return err
+	}
+	tx.QueryCache[key] = CachedState{list, total, wasLocked}
+	return nil
+}
+
+func (tx *CachedTransaction) ClearCache() {
+	tx.QueryCache = make(map[string]CachedState)
+}
+
+func (tx *CachedTransaction) LockList(s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions, pg *pagination.Paginator, lockPolicy schema.LockPolicy) (list []*schema.Resource, total uint64, err error) {
+	return tx.LockListContext(context.Background(), s, filter, options, pg, lockPolicy)
+}
+
+func (tx *CachedTransaction) LockListContext(ctx context.Context, s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions, pg *pagination.Paginator, lockPolicy schema.LockPolicy) (list []*schema.Resource, total uint64, err error) {
+
+	sc := lockListContextHelper(s, filter, options, pg, lockPolicy)
+
+	list, total, wasCached, wasLocked, err := tx.getCached(s.ID, sc)
+	if err != nil {
+		return
+	}
+
+	if !wasCached || !wasLocked {
+		if wasCached {
+			metrics.UpdateCounter(1, "tx.%s.cache.notLocked", s.ID)
+		}
+		metrics.UpdateCounter(1, "tx.%s.cache.missLock", s.ID)
+
+		list, total, err = tx.TxInterface.LockListContext(ctx, s, filter, options, pg, lockPolicy)
+		tx.saveCache(s.ID, sc, list, total, true, err)
+	} else {
+		metrics.UpdateCounter(1, "tx.%s.cache.hitLock", s.ID)
+	}
+	return
+}
+
+func (tx *CachedTransaction) Query(s *schema.Schema, query string, arguments []interface{}) (list []*schema.Resource, err error) {
+	return tx.QueryContext(context.Background(), s, query, arguments)
+}
+
+func (tx *CachedTransaction) QueryContext(ctx context.Context, s *schema.Schema, query string, arguments []interface{}) (list []*schema.Resource, err error) {
+	tx.ClearCache()
+	return tx.TxInterface.QueryContext(context.Background(), s, query, arguments)
+}
+
+func (tx *CachedTransaction) Fetch(s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions) (*schema.Resource, error) {
+	return tx.FetchContext(context.Background(), s, filter, options)
+}
+
+func (tx *CachedTransaction) FetchContext(ctx context.Context, s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions) (*schema.Resource, error) {
+	list, _, err := tx.ListContext(ctx, s, filter, options, nil)
+	return fetchContextHelper(list, err, filter)
+}
+
+func (tx *CachedTransaction) LockFetch(s *schema.Schema, filter transaction.Filter, lockPolicy schema.LockPolicy, options *transaction.ViewOptions) (*schema.Resource, error) {
+	return tx.LockFetchContext(context.Background(), s, filter, lockPolicy, options)
+}
+
+func (tx *CachedTransaction) LockFetchContext(ctx context.Context, s *schema.Schema, filter transaction.Filter, lockPolicy schema.LockPolicy, options *transaction.ViewOptions) (*schema.Resource, error) {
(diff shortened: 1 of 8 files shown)


db/transaction/transaction_test.go
```diff
@@ -50,4 +57,131 @@ var _ = Describe("Transaction", func() {
...
+			Expect(count).To(Equal(1))
+		})
+
+		It("List calls get cached", func() {
+			Expect(count).To(Equal(0))
+			cachedTx.List(netSchema, tx.Filter{}, nil, nil)
+			Expect(count).To(Equal(1))
+			cachedTx.List(netSchema, tx.Filter{}, nil, nil)
+			Expect(count).To(Equal(1))
+			cachedTx.LockList(netSchema, tx.Filter{}, nil, nil, schema.LockRelatedResources)
```
p-kozlowski: First call to Lock* methods should never be cached - gohan relies on MySQL locks for synchronization

db/transaction/transaction_test.go
```diff
@@ -50,4 +57,131 @@ var _ = Describe("Transaction", func() {
...
+			cachedTx.LockFetchContext(nil, netSchema, tx.Filter{}, schema.LockRelatedResources, nil)
+			Expect(count).To(Equal(1))
+			cachedTx.LockFetchContext(nil, netSchema, tx.Filter{}, schema.SkipRelatedResources, nil)
+			Expect(count).To(Equal(1))
+		})
+
+		It("Updates / inserts clear cache", func() {
+			Expect(count).To(Equal(0))
+			cachedTx.List(netSchema, tx.Filter{}, nil, nil)
+			Expect(count).To(Equal(1))
```
p-kozlowski: sorry, but I can't follow what this tests is actually testing: please extract subfunctions for each step

db/sql/sql.go
```diff
@@ -932,6 +998,72 @@ func (tx *Transaction) ListContext(ctx context.Context, s *schema.Schema, filter
...
+	sqlFormat := strings.Replace(sql, "?", "%s", -1)
+	query := fmt.Sprintf(sqlFormat, args...)
+	res := fmt.Sprintf("%s %s %v %v %v",schemeID, query, sc.join, sc.filter, sc.fields);
+	return hash(res)
+}
+
+func (tx *CachedTransaction) getCached(schemeID string, sc *selectContext, sql string, args []interface{}) (list []*schema.Resource, total uint64, errList error, errCached error) {
+	key := tx.createKey(schemeID, sc, sql, args)
+	res, present := tx.QueryCache[key]
+	if present {
```
p-kozlowski: I'd rather go the defensive way and check for hash collisions

db/sql/sql.go
```diff
@@ -932,6 +998,72 @@ func (tx *Transaction) ListContext(ctx context.Context, s *schema.Schema, filter
 	return tx.executeSelect(ctx, sc, sql, args)
 }
 
+func (tx *CachedTransaction) List(s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions, pg *pagination.Paginator) (list []*schema.Resource, total uint64, err error) {
+	return tx.ListContext(context.Background(), s, filter, options, pg)
+}
+
+//List resources in the db
+func (tx *CachedTransaction) ListContext(ctx context.Context, s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions, pg *pagination.Paginator) (list []*schema.Resource, total uint64, err error) {
+	sc := &selectContext{
```
p-kozlowski: code duplication with `(tx *Transaction) ListContext`, please extract a helper method

db/sql/sql.go
```diff
@@ -984,10 +1120,56 @@ func (tx *Transaction) LockListContext(ctx context.Context, s *schema.Schema, fi
 	return tx.executeSelect(ctx, sc, sql, args)
 }
 
+// LockList locks resources in the db
+func (tx *CachedTransaction) LockListContext(ctx context.Context, s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions, pg *pagination.Paginator, lockPolicy schema.LockPolicy) (list []*schema.Resource, total uint64, err error) {
+
+	policyJoin := shouldJoin(lockPolicy)
```
p-kozlowski: code duplication with `(tx *Transaction) LockListContext`, please extract a helper method.

db/sql/sql.go
```diff
@@ -984,10 +1120,56 @@ func (tx *Transaction) LockListContext(ctx context.Context, s *schema.Schema, fi
...
+	if options != nil {
+		sc.fields = normFields(options.Fields, s)
+		sc.join = policyJoin && options.Details
+	}
+
+	sql, args, err := buildSelect(sc)
+	if err != nil {
+		return nil, 0, err
+	}
+
```
p-kozlowski:
```	
if tx.db.sqlType == "mysql" {
    sql += " FOR UPDATE"
}
```
is missing

db/sql/sql.go
```diff
@@ -932,6 +998,72 @@ func (tx *Transaction) ListContext(ctx context.Context, s *schema.Schema, filter
...
+	return
+}
+
+func hash(s string) uint32 {
+	h := fnv.New32a()
+	h.Write([]byte(s))
+	return h.Sum32()
+}
+
+func (tx *CachedTransaction) createKey(schemeID string, sc *selectContext, sql string, args []interface{}) uint32 {
```
p-kozlowski: minor: `schema`, not `scheme`

db/sql/sql.go
```diff
@@ -586,6 +610,12 @@ func (tx *Transaction) logQuery(sql string, args ...interface{}) {
 	log.Debug("[%p] Executing SQL query '%s'", tx.transaction, query)
 }
 
+func (tx *CachedTransaction) logQuery(sql string, args ...interface{}) {
```
p-kozlowski: copy-paste of `(tx *Transaction) logQuery`

db/sql/sql.go
```diff
@@ -1115,6 +1316,17 @@ func (tx *Transaction) LockFetchContext(ctx context.Context, s *schema.Schema, f
...
+	list, _, err := tx.LockListContext(ctx, s, filter, nil, nil, lockPolicy)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to fetch and lock %s: %s", filter, err)
+	}
+	if len(list) < 1 {
+		return nil, transaction.ErrResourceNotFound
+	}
+	return list[0], nil
+}
+
```
p-kozlowski: this file is more than 1500 lines now. could we move `CachedTransaction` to a separate file?

db/sql/sql.go
```diff
@@ -1115,6 +1316,17 @@ func (tx *Transaction) LockFetchContext(ctx context.Context, s *schema.Schema, f
 	return list[0], nil
 }
 
+func (tx *CachedTransaction) LockFetchContext(ctx context.Context, s *schema.Schema, filter transaction.Filter, lockPolicy schema.LockPolicy, options *transaction.ViewOptions) (*schema.Resource, error) {
+	list, _, err := tx.LockListContext(ctx, s, filter, nil, nil, lockPolicy)
```
p-kozlowski: code duplication with `(tx *Transaction) LockFetchContext`

db/sql/sql.go
```diff
@@ -1097,10 +1283,25 @@ func (tx *Transaction) FetchContext(ctx context.Context, s *schema.Schema, filte
 	return list[0], nil
 }
 
+func (tx *CachedTransaction) FetchContext(ctx context.Context, s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions) (*schema.Resource, error) {
```
p-kozlowski: code duplication with `(tx *Transaction) FetchContext`. don't be shy to refactor the existing code if needed - we've got extensive unit tests to assist you :)

db/sql/sql.go
```diff
@@ -686,10 +734,19 @@ func (tx *Transaction) UpdateContext(ctx context.Context, resource *schema.Resou
...
+	return tx.StateUpdateContext(context.Background(), resource, state)
+}
+
 func (tx *Transaction) StateUpdate(resource *schema.Resource, state *transaction.ResourceState) error {
 	return tx.StateUpdateContext(context.Background(), resource, state)
 }
 
+func (tx *CachedTransaction) StateUpdateContext(ctx context.Context, resource *schema.Resource, state *transaction.ResourceState) error {
+	tx.ClearCache()
+	return tx.StateUpdateContext(context.Background(), resource, state)
```
p-kozlowski: infinite recursion: should be `tx.TxInterface.StateUpdateContext`

db/sql/sql.go
JDCodi: Ok, lets assume that we have a collision, what should we do? Only decent solution I can think of is to add the original query to the cache and check it always, after extracting anything from cache. And what then? we can just assume that we only cache one query with the given key or add the key at the end of the query to create another key and store the new request under it.

db/sql/sql.go
JDCodi: Yes, will fix

db/sql/sql.go
JDCodi: Yes, will fix

db/sql/sql.go
JDCodi: will fix

db/sql/sql.go
JDCodi: It's hard to avoid as not every transaction.Transaction needs it but maybe will add it there.

db/sql/sql.go
```diff
@@ -932,6 +998,72 @@ func (tx *Transaction) ListContext(ctx context.Context, s *schema.Schema, filter
...
+		filter:    filter,
+		join:      true,
+		paginator: pg,
+	}
+	if options != nil {
+		sc.fields = normFields(options.Fields, s)
+		sc.join = options.Details
+	}
+
+	sql, args, err := buildSelect(sc)
```
zimnx: This probably brings performance impact since all SQL queries are generated twice. This operation requires memory allocations and iterations. Also flow is very frequent. I think we should calculate hash for parameters of TxTransaction methods, not SQL query.

db/sql/sql.go
p-kozlowski: in fact we could just use the plain-text query as a key instead of the hash in the `QueryCache` (which is a hashmap itself, so apart from small memory optimization there's not much added benefit)

db/sql/sql.go
JDCodi: I did it because i wasn't sure if there could be any changes in schema that could cause generating different query for the same parameters. But i can investigate it further

db/sql/sql.go
```diff
@@ -72,6 +73,19 @@ type Transaction struct {
 	isolationLevel transaction.Type
 }
 
+type CachedState struct {
+	list           []*schema.Resource
+	total          uint64
+	errList        error
```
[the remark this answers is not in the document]
JDCodi: Right, will fix

db/sql/sql.go
```diff
@@ -365,17 +380,22 @@ func (db *DB) Begin() (tx transaction.Transaction, err error) {
...
 		isolationLevel: transaction.RepeatableRead,
 	}
+
 	if os.Getenv("FUZZY_DB_TX") == "true" {
 		log.Notice("FUZZY_DB_TX is enabled")
-		tx = &transaction.FuzzyTransaction{Tx: tx}
+		tx = &transaction.FuzzyTransaction{Tx: transaction.Transaction(&transx)}
+	} else {
+		cachedTransaction := &CachedTransaction{&transx, nil }
+		cachedTransaction.ClearCache()
```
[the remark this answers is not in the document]
JDCodi: Will fix

db/sql/sql.go
```diff
@@ -932,6 +998,72 @@ func (tx *Transaction) ListContext(ctx context.Context, s *schema.Schema, filter
...
+	return hash(res)
+}
+
+func (tx *CachedTransaction) getCached(schemeID string, sc *selectContext, sql string, args []interface{}) (list []*schema.Resource, total uint64, errList error, errCached error) {
+	key := tx.createKey(schemeID, sc, sql, args)
+	res, present := tx.QueryCache[key]
+	if present {
+		return res.list, res.total, res.errList, nil
+	} else {
+		return nil, 0, nil, fmt.Errorf("NoCachedValue!")
```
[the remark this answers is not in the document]
JDCodi: Will fix

db/sql/sql.go
JDCodi: right, will fix.

db/sql/sql.go
JDCodi: memory + speed as map operations on long strings can be slow... but we won't have too many records there so it should be fine anyway

db/sql/sql.go
JDCodi: results will be the same, even if we skip "for update".

db/sql/cached_transaction.go
```diff
@@ -0,0 +1,189 @@
...
+	_ "github.com/nati/go-fakedb"
+)
+
+type CachedState struct {
+	list           []*schema.Resource
+	total          uint64
+	isLocked       bool
+}
+
+//type TxInterface transaction.Transaction
```
p-kozlowski: commented code

db/sql/cached_transaction.go
```diff
@@ -0,0 +1,189 @@
...
+
+	if !wasCached {
+		list, total, err = tx.TxInterface.ListContext(ctx, s, filter, options, pg)
+		tx.saveCache(s.ID, sc, list, total, false, err)
+	}
+	return
+}
+
+func (tx *CachedTransaction) createKey(schemaID string, sc *selectContext) string {
+	res := fmt.Sprintf("%s %v %v %v",schemaID, sc.join, sc.filter, sc.fields);
```
zimnx: This doesn't work for all cases: https://play.golang.org/p/iV-L-1tn0mv

db/sql/cached_transaction.go
```diff
@@ -0,0 +1,189 @@
...
+	return tx.TxInterface.DeleteContext(context.Background(), s, resourceID)
+}
+
+func (tx *CachedTransaction) List(s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions, pg *pagination.Paginator) (list []*schema.Resource, total uint64, err error) {
+	return tx.ListContext(context.Background(), s, filter, options, pg)
+}
+
+//List resources in the db
+func (tx *CachedTransaction) ListContext(ctx context.Context, s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions, pg *pagination.Paginator) (list []*schema.Resource, total uint64, err error) {
+	sc := &selectContext{
```
p-kozlowski: `listContextHelper`?

db/sql/cached_transaction.go
```diff
@@ -0,0 +1,189 @@
+// Copyright (C) 2015 NTT Innovation Institute, Inc.
```
zimnx: nit: 2018

db/sql/cached_transaction.go
zimnx: I think its worth to try https://github.com/mitchellh/hashstructure

db/sql/cached_transaction.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	if options != nil {
+		sc.fields = normFields(options.Fields, s)
+		sc.join = options.Details
+	}
+
+	list, total, wasCached, _ := tx.getCached(s.ID, sc)
+
+	if !wasCached {
+		list, total, err = tx.TxInterface.ListContext(ctx, s, filter, options, pg)
```
p-kozlowski: runtime metrics for cache hit/miss/miss due to lock could be useful

db/sql/sql.go
```diff
@@ -399,13 +406,19 @@ func (db *DB) BeginTx(ctx context.Context, options *transaction.TxOptions) (tx t
...
 		db:             db,
 		transaction:    rawTx,
 		closed:         false,
 		isolationLevel: options.IsolationLevel,
 	}
-	log.Debug("[%p] Created transaction %#v, isolation level %s", rawTx, rawTx, tx.GetIsolationLevel())
+	if transx.isolationLevel == transaction.RepeatableRead || transx.isolationLevel == transaction.Serializable {
+		tx = MakeCachedTransaction(&transx)
+	} else {
+		tx = TxInterface(&transx)
```
p-kozlowski: just `tx = &transx` is fine

db/sql/cached_transaction.go
```diff
@@ -93,28 +91,23 @@ func (tx *CachedTransaction) List(s *schema.Schema, filter transaction.Filter, o
...
 		tx.saveCache(s.ID, sc, list, total, false, err)
+	} else {
+		metrics.UpdateCounter(1, "tx.cache.hit")
 	}
 	return
 }
 
 func (tx *CachedTransaction) createKey(schemaID string, sc *selectContext) string {
-	res := fmt.Sprintf("%s %v %v %v",schemaID, sc.join, sc.filter, sc.fields);
+	filterHash, _ := hashstructure.Hash(sc.filter, nil)
```
zimnx: err is omitted, propagate it higher

db/sql/cached_transaction.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+//List resources in the db
+func (tx *CachedTransaction) ListContext(ctx context.Context, s *schema.Schema, filter transaction.Filter, options *transaction.ViewOptions, pg *pagination.Paginator) (list []*schema.Resource, total uint64, err error) {
+	sc := listContextHelper(s, filter, options, pg)
+
+	list, total, wasCached, _ := tx.getCached(s.ID, sc)
+
+	if !wasCached {
+		metrics.UpdateCounter(1, "tx.cache.miss")
```
p-kozlowski: sorry not to be explicit about it before, but including the schema name in the metric path would be nice (eg.  https://github.com/cloudwan/gohan/blob/master/db/sql/sql.go#L590). note that it's ok to pass format+args to UpdateCounter (it's a nice, but premature optimization: if the metrics are turned off, we don't pay for building the string)

JDCodi: Ok, will add it

db/sql/cached_transaction.go
```diff
@@ -0,0 +1,196 @@
...
+	if sc.paginator != nil {
+		res = fmt.Sprintf("%s %v", res, *sc.paginator)
+	}
+	return res, err
+}
+
+func (tx *CachedTransaction) getCached(schemeID string, sc *selectContext) (list []*schema.Resource, total uint64, wasCached bool, wasLocked bool) {
+	key, err := tx.createKey(schemeID, sc)
+	if err != nil {
+		return nil, 0, false, false
```
zimnx: err isn't propagated higher. User won't have any information why his tx operation failed.

db/transaction/transaction_test.go
```diff
@@ -50,4 +57,142 @@ var _ = Describe("Transaction", func() {
...
+		It("List calls get cached", func() {
+			Expect(count).To(Equal(0))
+			cachedTx.List(netSchema, tx.Filter{}, nil, nil)
+			Expect(count).To(Equal(1))
+			cachedTx.List(netSchema, tx.Filter{}, nil, nil)
+			Expect(count).To(Equal(1))
+			Expect(countLock).To(Equal(0))
+			cachedTx.LockList(netSchema, tx.Filter{}, nil, nil, schema.LockRelatedResources)
+			Expect(count).To(Equal(1))
+			Expect(count).To(Equal(1))
```
michalzurawski: countLock

db/sql/cached_transaction.go
```diff
@@ -0,0 +1,206 @@
...
+	cachedTransaction.ClearCache()
+	return cachedTransaction
+}
+
+func (tx *CachedTransaction) Create(resource *schema.Resource) error {
+	return tx.CreateContext(context.Background(), resource)
+}
+
+func (tx *CachedTransaction) CreateContext(ctx context.Context, resource *schema.Resource) error {
+	tx.ClearCache()
```
jkmar: Since create affects only one table I don't think it's necessary to clear all cache.

PR #634, opened 2018-04-13
Add preserving unspecified fields during update in internal schema objects

JDCodi:


schema/resource.go
```diff
@@ -125,13 +125,43 @@ func (resource *Resource) Update(updateData map[string]interface{}) error {
...
+				return err
+			}
 		}
 	}
 	return nil
 }
 
+func updateResourceRecursion(updateData interface{}, sourceData interface{}) (res interface{}, err error) {
+	switch updateData.(type) {
+	case map[string]interface{}:
```
p-kozlowski: does this code handle slices as well? in either case, please write a test for it

server/resources/resource_management_test.go
```diff
@@ -1292,6 +1305,27 @@ var _ = Describe("Resource manager", func() {
 					Expect(ok).To(BeTrue())
 					Expect(theResource).To(HaveKeyWithValue("test_string", "Ia, ia, HJPEV fhtang!"))
 				})
+
+				It("Should only modify updated fields", func() {
+					networkSchema, _ := manager.Schema("network")
+
+					err := resources.CreateResource(context, testDB, fakeIdentity, networkSchema, adminNetworkData)
+					Expect(err).ToNot(HaveOccurred())
+					result := context["response"].(map[string]interface{})
```
michalzurawski: result is unused - and i think it could be good idea to check its contents (i.e. `Expect(result).To(Equal))`) as it'll be easier to spot which part of network has been updated

server/resources/resource_management_test.go
```diff
@@ -1292,6 +1305,27 @@ var _ = Describe("Resource manager", func() {
 					Expect(ok).To(BeTrue())
 					Expect(theResource).To(HaveKeyWithValue("test_string", "Ia, ia, HJPEV fhtang!"))
 				})
+
+				It("Should only modify updated fields", func() {
```
p-kozlowski: name is misleading: you never actually check if the updated field is modified or not. please add an assert on `vlan_id`

server/resources/resource_management_test.go
```diff
@@ -1160,6 +1162,17 @@ var _ = Describe("Resource manager", func() {
 				"test_integer": 1,
 				"test_bool":    false,
 			}
+			adminNetworkData = map[string]interface{}{
```
p-kozlowski: these variables are used only in a single test case. could we move them to `BeforeEach` in `Context("As an admin"`?

server/resources/resource_management_test.go
JDCodi: good point, will fix

schema/resource.go
```diff
@@ -125,13 +125,43 @@ func (resource *Resource) Update(updateData map[string]interface{}) error {
...
+				updated, err := updateResourceRecursion(val, newSource[key])
+				if err == nil {
+					newSource[key] = updated
+				} else {
+					return nil, err
+				}
+			}
+			return newSource, nil
+		} else {
+			return nil, fmt.Errorf("Invalid update data")
```
p-kozlowski: The update data is indeed invalid, but this error message does not tell the user anything helpful to fix their mistake. Could we improve this error message?
By the way, is this error even possible? If so, please write a test for it, if not, it could be removed, greatly simplifying the logic of this function.

server/resources/resource_management_test.go
```diff
@@ -1292,6 +1305,27 @@ var _ = Describe("Resource manager", func() {
 					Expect(ok).To(BeTrue())
 					Expect(theResource).To(HaveKeyWithValue("test_string", "Ia, ia, HJPEV fhtang!"))
 				})
+
+				It("Should only modify updated fields", func() {
+					networkSchema, _ := manager.Schema("network")
+
+					err := resources.CreateResource(context, testDB, fakeIdentity, networkSchema, adminNetworkData)
```
p-kozlowski: a bit shorter version that reads as a sentence:
`Expect(resources.CreateResource(...)).To(Succeed())`

schema/resource.go
```diff
@@ -125,13 +125,43 @@ func (resource *Resource) Update(updateData map[string]interface{}) error {
...
+			}
 		}
 	}
 	return nil
 }
 
+func updateResourceRecursion(updateData interface{}, sourceData interface{}) (res interface{}, err error) {
+	switch updateData.(type) {
+	case map[string]interface{}:
+		newUpdate, foundUpdate := updateData.(map[string]interface{})
```
p-kozlowski: at this point you already know that `updateData` is a `map[string]interface{}`, so additional conversion is not needed. you could use `switch t := t.(type) {` syntax to simplify it. (cf. https://golang.org/doc/effective_go.html#type_switch)

schema/resource.go
```diff
@@ -125,13 +125,43 @@ func (resource *Resource) Update(updateData map[string]interface{}) error {
 	}
 	for _, property := range resource.schema.Properties {
 		id := property.ID
+
 		if val, ok := updateData[id]; ok {
-			data[id] = val
+			newVal, err := updateResourceRecursion(val, data[id])
+			if err == nil {
```
aleksander-bulanowski: I think we should use 
`if err != nil {
  return err
}
data[id] = newVal`

schema/resource.go
```diff
@@ -125,13 +125,36 @@ func (resource *Resource) Update(updateData map[string]interface{}) error {
 	}
 	for _, property := range resource.schema.Properties {
 		id := property.ID
+
 		if val, ok := updateData[id]; ok {
-			data[id] = val
+			newVal := updateResourceRecursion(val, data[id])
```
p-kozlowski: just `data[id] = updateResourceRecursion(val, data[id]`

p-kozlowski:
LGTM, please just add a test for overwriting an undefined value with an object and rebase

morrisson:
@JDCodi Please rebase.

zimnx:
@morrisson waiting for your approval

PR #636, opened 2018-04-19
Make requests with limit=0 return empty lists

JDCodi:


db/pagination/pagination.go
```diff
@@ -37,12 +37,12 @@ const (
...
 	Key    string
 	Order  string
-	Limit  uint64
+	Limit  int64
 	Offset uint64
 }
 
 //NewPaginator create Paginator
-func NewPaginator(s *schema.Schema, key, order string, limit, offset uint64) (*Paginator, error) {
+func NewPaginator(s *schema.Schema, key, order string, limit int64, offset uint64) (*Paginator, error) {
```
p-kozlowski: most of these parameters are in fact optional. could we refactor this code a bit?

morrisson:
I wonder what the use case for this change. I sometimes see `0 = unlimited` case in some system so I do not think the current design is bad. Cc: @nati

db/pagination/pagination.go
```diff
@@ -37,40 +37,82 @@ const (
 type Paginator struct {
 	Key    string
 	Order  string
-	Limit  uint64
+	Limit  int64
```
p-kozlowski: could we come back to unsigned? semantically, the limit could not be negative

zimnx:
@morrisson Paginator is structure used during SQL query building. IMO we should follow SQL logic for that, which is LIMIT = 0 means no records returned. 
Use case is dashboard showing only some statistics without downloading data. Gohan will return only header X-Total-Count with number of matching rows.

db/pagination/pagination.go
JDCodi: working on it, not trivial as uint64 max value won't work with js...

zimnx:
@morrisson I agree, HEAD is better option. Although GET with response removed still requires some unnecessary data round-trip over DB <->Gohan connection. If we would have option to eliminate data from DB response then we could implement HEAD as you described. This PR introduces this option.

morrisson:
@zimnx Ok. I understand.

JDCodi:
Came back to uint64. It was possible due to refactor in gohan_db.go (problem with js variable size...). Rebased, squashed.

db/sql/sql.go
```diff
@@ -876,8 +877,8 @@ func buildSelect(sc *selectContext) (string, []interface{}, error) {
 			}
 		}
 
-		if sc.paginator.Limit > 0 {
-			q = q.Limit(sc.paginator.Limit)
+		if sc.paginator.Limit != math.MaxUint64 {
```
morrisson: Same change should be necessary for `file.go` (sqlite3)

JDCodi: Right, fixed

extension/otto/gohan_db.go
```diff
@@ -439,60 +362,100 @@ func init() {
...
-		return []map[string]interface{}{}, fmt.Errorf("Error during gohan_db_lock_list: %s", err.Error())
+		return
 	}
 
-	return parseListResults(resources), nil
+	return schema, filter, pg, tx, needCommit, err
+}
+
+func parseListResults(resources []*schema.Resource) []map[string]interface{} {
+	resp := []map[string]interface{}{}
```
michalzurawski: nit: 
```
        resp := make([]map[string]interface{}, len(resources))
	for i, resource := range resources {
		resp[i] = resource.Data()
	}
```

db/file/file.go
```diff
@@ -21,6 +21,8 @@ import (
 	"sort"
 	"strconv"
 
+	"math"
```
michalzurawski: nit: goimport

PR #637, opened 2018-04-20
 Check for unrecognized parameters in a query 

JDCodi:


morrisson:
I'm pretty neutral to add this feature but may affect clients. @nati I will notify this to you because this change potentially affects your project so it would be grateful if you leave a comment.

PR #638, opened 2018-04-25
Add new schema formats for IPv4

JDCodi:


schema/formats.go
```diff
@@ -51,6 +55,41 @@ func (f cidrOrIPv4FormatChecker) IsFormat(input string) bool {
...
+}
+
+func (f ipv4AddressWithCidrFormatChecker) IsFormat(input string) bool {
+	hostIP, netIP, cidrErr := extractHostAndNet(input)
+	if cidrErr != nil {
+		return false
+	}
+
+	for i := range hostIP {
+		if hostIP[i] == 255 {
```
marcin-ptaszynski: this is not going to work for cases where prefix length isn't multiple of 8, e.g. Broadcast address for subnet 192.168.2.0/25 is 192.168.2.127.

schema/formats.go
```diff
@@ -51,6 +55,41 @@ func (f cidrOrIPv4FormatChecker) IsFormat(input string) bool {
 	return (cidrErr == nil && cidrIP.To4() != nil) || (ip != nil && ip.To4() != nil)
 }
 
+func (f ipv4NetworkFormatChecker) IsFormat(input string) bool {
+	hostIP, netIP, cidrErr := extractHostAndNet(input)
+	if cidrErr != nil {
+		return false
+	}
+	return reflect.DeepEqual(hostIP, netIP)
```
marcin-ptaszynski: [net.IP exposes Equal method](https://golang.org/pkg/net/#IP.Equal).

schema/formats.go
```diff
@@ -51,6 +55,41 @@ func (f cidrOrIPv4FormatChecker) IsFormat(input string) bool {
...
+
+	for i := range hostIP {
+		if hostIP[i] == 255 {
+			return false
+		}
+	}
+	return !reflect.DeepEqual(hostIP, netIP)
+}
+
+func extractHostAndNet(input string) (hostIP net.IP, netIP net.IP, err error) {
```
marcin-ptaszynski: this method loses cidr information, I think if would be called extractHostAndNetAddresses, it might be more to the point.

schema/formats.go
JDCodi: true, fixed

marcin-ptaszynski: This is still not fixed, please check https://play.golang.org/p/46MHiZHeuMs for correct method of computing broadcast address.

morrisson:
Let's merge after the rebase.

PR #639, opened 2018-04-27
Treat empty strings as undefined during unmarshalling

zimnx:
This is a workaround for issue which prohibits doing request validation in PreUpdate events using golang extensions. Structures provided to schema handlers which has fields with `string` type will be filled with empty string during allocation. When user doesn't provide this field in his request body, he don't intents to update it, but rather leave value as it is. Current code will try to update each of these fields to empty strings, which is not expected behavior. Additionally validation may fail because some fields may require that length is not 0.


zimnx:
It was unusable in tests. Need another workaround

PR #641, opened 2018-05-04
Clear zero values from dataMap before Update validation

zimnx:
This is a workaround for issue which prohibits doing request validation in PreUpdate events using golang extensions. Structures provided to schema handlers which has fields with string type will be filled with empty string during allocation. When user doesn't provide this field in his request body, he don't intents to update it, but rather leave value as it is. Current code will try to update each of these fields to empty strings, which is not expected behavior. Additionally validation may fail because some fields may require that length is not 0.

### DIFF
--- a/server/resources/resource_management.go
+++ b/server/resources/resource_management.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
@@ -51,6 +52,8 @@ const (
 	Unauthorized
 	Forbidden
 	ForeignKeyFailed
+
+	goValidationContextKey = "go_validation"
 )
 
 // ResourceError is created when an anticipated problem has occurred during resource manipulations.
@@ -586,7 +589,7 @@ func CreateResource(
 
 	//Validation
 
-	if _, ok := context["go_validation"]; ok {
+	if _, ok := context[goValidationContextKey]; ok {
 		err = resourceSchema.ValidateGoOnCreate(dataMap)
 		if err != nil {
 			return ResourceError{err, fmt.Sprintf("Validation error: %s", err), WrongData}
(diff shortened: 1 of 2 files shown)


morrisson:
@zimnx It looks this PR may occur really bad side effect. If an user wants to update with "zero-value" such as `"", 0, false`, this code blocks to update those values. Can you confirm that?

zimnx:
@morrisson You're right. None of our tests validated this use case. I've fixed implementation, and added missing testcase.

server/resources/resource_management_test.go
```diff
@@ -1428,6 +1429,57 @@ var _ = Describe("Resource manager", func() {
...
+				Context("pre_update inserts zero-value", func() {
+					var (
+						requestMap map[string]interface{}
+					)
+					BeforeEach(func() {
+						requestMap = map[string]interface{}{
+							"test_string": "Steloj ne estas en ordo.",
+							"test_bool":   true,
+						}
+						events["pre_update"] = `if (context.resource.test_bool === undefined) { context.resource.test_bool = false; }`
```
michalzurawski: unnecessary

server/resources/resource_management.go
```diff
@@ -24,6 +24,8 @@ import (
 	"strings"
 	"time"
 
+	"reflect"
```
michalzurawski: nit: goimport

server/resources/resource_management_test.go
zimnx: why? It simulates Goext behavoir in JS code.

server/resources/resource_management.go
```diff
@@ -795,6 +796,25 @@ func UpdateResourceInTransaction(
 		return ResourceError{err, "", Unauthorized}
 	}
 
+	if _, ok := context["go_validation"]; ok {
```
michalzurawski: "go_validation" appears already twice - maybe const?

zimnx: done

server/resources/resource_management.go
zimnx: done

PR #642, opened 2018-05-09
Cw 335 fixes for limit zero ver2

aleksander-bulanowski:


michalzurawski:
it's worth mention one drawback of this solution pointed out by @aleksander-bulanowski - a conversion between int and uint. It's possible to pass a negative value which then would be translated to uint

morrisson:
Please squash commits and rebase.

db/file/file.go
```diff
@@ -300,7 +301,7 @@ func (tx *Transaction) List(s *schema.Schema, filter transaction.Filter, options
 
 		if pg != nil {
 			sort.Sort(byPaginator{list, pg})
-			if pg.Limit > 0 {
+			if pg.Limit != math.MaxUint64 {
 				list = list[:pg.Limit]
```
jkmar: this will panic if `pg.Limit > len(list)`

extension/goext/schemas.go
```diff
@@ -53,7 +53,7 @@ type Filter map[string]interface{}
 type Paginator struct {
 	Key    string
 	Order  string
-	Limit  uint64
+	Limit  MaybeInt
```
jkmar: Since was uint64 before I'd rather made a new type for it: `MaybeUint64`.

morrisson:
@aleksander-bulanowski you are working on #643 so we can close this PR without the merge, right?

aleksander-bulanowski:
Since https://github.com/cloudwan/gohan/pull/643 was merged there is no need to merge this PR.

PR #643, opened 2018-05-10
Cw 335 fixes for limit zero ver3

aleksander-bulanowski:


extension/goext/schemas.go
```diff
@@ -57,6 +58,49 @@ type Paginator struct {
 	Offset uint64
 }
 
+// Below code is adapted from similar code in db/pagination/pagination.go, but
+// doesn't validate parameters.
+type OptionPaginator func(*Paginator)
```
jkmar: Unfortunately there is bit of duplication here, is it possible to convert `paginator` into an interface?

PR #653, opened 2018-06-06
[CW-485] Hide database error for foreign keys

michalzurawski:


server/resources/resource_management.go
```diff
@@ -660,7 +660,7 @@ func CreateResourceInTransaction(context middleware.Context, resourceSchema *sch
 		if isForeignKeyFailed(err) {
 			return ResourceError{
 				err,
-				fmt.Sprintf("Foreign resource not found: %v", err),
+				"Related resource does not exist",
```
morrisson: This fix enables to avoid exposing our backend to the user, but also hides the reason of error. We should let users know what input is wrong as much as possible.

- (Best) Extracts what property is not good input from `isForeignKeyFailed` function.
- If it's difficult, this error message should include an user input (`resource.properties`).
- Write a error log with MySQL raw error message for debugging purpose.

michalzurawski: that was my first idea (to extract it from db error message) - it's not a problem to extract it from mysql error, however, it's not possible for sqlite as it doesn't include any useful information in its error message

morrisson: Got it. Then, we should take second option. At least, we must expose raw error message in a log.

PR #655, opened 2018-06-20
[CW-316] Fixed validation for Go resources

michalzurawski:


server/resources/resource_management.go
```diff
@@ -593,20 +593,19 @@ func CreateResource(
...
-	if resourceData, ok := context["resource"].(map[string]interface{}); ok {
-		dataMap = resourceData
-	}
-
 	//Validation
-
 	if _, ok := context[goValidationContextKey]; ok {
-		err = resourceSchema.ValidateGoOnCreate(dataMap)
-		if err != nil {
+		if err := resourceSchema.ValidateOnCreate(dataMap); err != nil {
```
p-kozlowski: as discussed offline: UPDATE does the same thing [differently](https://github.com/cloudwan/gohan/blob/master/server/resources/resource_management.go#L807). as a result, fields set by PreUpdate handlers may be lost.

I think we should unify the way ValidateOnCreate and ValidateOnUpdate are called in this file.

db/db_test.go
```diff
@@ -328,9 +328,6 @@ var _ = Describe("Database operation test", func() {
 				})
 
 				It("Updates the resource properly", func() {
-					By("Not allowing to update some fields")
-					Expect(networkResource1.Update(map[string]interface{}{"id": "new_id"})).ToNot(Succeed())
```
morrisson: Why is this removed?

extension/goplugin/integration_tests/end_to_end_test.go
```diff
@@ -139,10 +139,17 @@ var _ = Describe("Environment", func() {
 				"id":            "testId",
```
morrisson: Context should change to "Resource creation" to "Resource update" (line 129).

db/db_test.go
michalzurawski: this logic has been moved from `Update` to `UpdateResourceInTransaction` which calls `Update`

extension/goplugin/integration_tests/end_to_end_test.go
michalzurawski: it tries to create resource three times two of them fails due to validation and third succeeds with status code `http.StatusCreated`. There isn't any update in this test

michalzurawski:
> It looks no test is found for creation with empty properties. Please add test case for object creation.

Added to `resource_managment_test`

server/resources/resource_management_test.go
```diff
@@ -1438,12 +1438,78 @@ var _ = Describe("Resource manager", func() {
...
+						context["request_data"] = deepcopy.Copy(requestMap).(map[string]interface{})
+
+						err := resources.CreateResource(context, testDB, fakeIdentity, currentSchema, requestMap)
+						result := context["response"].(map[string]interface{})
+						Expect(err).NotTo(HaveOccurred())
+						theResource, ok := result[schemaID]
+						Expect(ok).To(BeTrue())
+						Expect(theResource).To(HaveKeyWithValue("test_bool", false))
+						Expect(theResource).To(HaveKeyWithValue("test_string", ""))
+					})
```
morrisson: You should include an unit test of failing resource creation if required value has not been filled in Goext.

PR #659, opened 2018-06-27
[CW-423] Added version to format property validation

michalzurawski:


schema/formats.go
```diff
@@ -95,4 +102,5 @@ func registerGohanFormats(checkers gojsonschema.FormatCheckerChain) {
 	checkers.Add("port", portFormatChecker{})
 	checkers.Add("yaml", yamlFormatChecker{})
 	checkers.Add("text", textFormatChecker{})
+	checkers.Add("version", versionFormatChecker{})
```
p-kozlowski: please add it to the docs

michalzurawski: Done

michalzurawski:
Build fails due to missing semver in vendor. I'll wait for https://github.com/cloudwan/gohan/pull/658 with adding it

morrisson:
#658 has been merged but still semver needs to be included.

morrisson:
@michalzurawski As you may know, `github.com/Masterminds/semver` should be stored in `vendor` directory.

michalzurawski:
will wait with this PR as there is still pending discussion - as you can see in the tests semver allows using i.e. "1", "1.1" or "v1.1.1" which probably doesn't suit our case

PR #663, opened 2018-06-28
Broken Integration Test Fix After Revert PR 662

morrisson:
Relates to #662


morrisson:
Finally I found we need another revert.

PR #666, opened 2018-07-03
BL-208 [goext] BeginTx() does not time out gracefully 

BartoszKakol:
Work in progress. Tests will be soon.


extension/goplugin/database.go
```diff
@@ -64,7 +64,11 @@ func (db *Database) Begin() (goext.ITransaction, error) {
 // BeginTx starts a new transaction with options
 func (db *Database) BeginTx(ctx goext.Context, options *goext.TxOptions) (goext.ITransaction, error) {
 	opts := transaction.TxOptions{IsolationLevel: transaction.Type(options.IsolationLevel)}
-	t, err := db.raw.BeginTx(context.Background(), &opts)
+	newTimedContext := ctx.GetContext()
+	if _, ok := newTimedContext.Deadline();!ok{
```
zimnx: s/ok/hasDeadline

extension/goplugin/database.go
```diff
@@ -64,7 +64,11 @@ func (db *Database) Begin() (goext.ITransaction, error) {
...
+		newTimedContext = context.WithTimeout(newTimedContext,5* time.Second)
```
zimnx: timeout should be configurable

p-kozlowski: please run goimports to format the code before pushing it. in GoLand you could add it in File->Tools->File Watchers - it will run whenever you save a file

BartoszKakol: It is configurable if you specify deadline in context. 5 seconds is only used if someone used function without specifying deadline.

zimnx: I meant that this 5 seconds value should be configurable. Although I don't think that's the right place to set this deadline. Maybe we should do at the beginning of request in api.go to make sure context is always there. What do you think?

BartoszKakol: We can return error from BeginTx if Timeout is not set, to force person using the function to add Timeout to Context. We need some sort of protection against Context without timeout otherwise we risk ending up with the same errors.

p-kozlowski: @BartoszKakol I think what @zimnx meant was that the timeout should be configurable using the config file, not by harcoding 5sec (and he's right about it). In my opinion that would be unnecessary - API requests always provide a context with timeout:
https://github.com/cloudwan/gohan/blob/master/extension/goplugin/environment.go#L438 (though again I agree it should be a part of Gohan core, not goplugin). timeout won't be here iff a user hand-mades a context, but that's already shooting oneself in their foot, so I wouldn't care that much for it.

in other words: just pass `ctx.GetContext()` to `BeginTx` and we'll be fine.

side note: returning an error when timeout is not set is not the best solution available. it's an API misuse and we should preferably make it a compile error, or, if that's not possible, `panic()` (aka. `assert` in other languages) to make it very clear that something is wrong.

extension/goplugin/database_test.go
```diff
@@ -83,6 +85,20 @@ var _ = Describe("Database", func() {
 				Expect(err).To(HaveOccurred())
 				Expect(tx).To(BeNil())
 			})
+
+			It("should call inner function using Context with Timeout", func() {
+				tempCtxWithTimeout, _ := context.WithTimeout(context.Background(), 50*time.Millisecond)
+				ctx:= goext.Context{
```
p-kozlowski: `goimports` complain, but otherwise looks good

BartoszKakol:
@morrisson Hi, this PR is approved, can you merge it, so we can close the ticket?

p-kozlowski:
@BartoszKakol please squash the commits

PR #669, opened 2018-07-11
BL-155 Tenant isolation via policy

piotr-dulikowski:
**Work in progress**. Needs cleanup, doc update and - optionally - attachment policy checking when target resource is modified.

### DIFF
--- a/docs/policy.md
+++ b/docs/policy.md
@@ -8,7 +8,7 @@ A policy has following properties.
 - action: one of `create`, `read`, `update`, `delete` for CRUD operations
   on the resource or any custom actions defined by schema performed on a
   resource or `*` for all actions
-- effect : Allow API access or not. `Deny` keyword (incasesensitive) block access, any other option (including lack of this property) allows
+- effect : Allow API access or not. `Deny` keyword (case insensitive) block access, any other option (including lack of this property) allows
 - resource : target resource
   you can specify target resource using "path" and "properties"
 - condition : additional condition (see below)
@@ -262,4 +262,60 @@ In an example below admin has access to all methods in all paths except `delete`
   principal: admin
   resource:
     path: /v2.0/restricted_method.*
+```
+
+## Attach policy
+
+Attach policy defines allowed values for a relation property
+(a resource "attaches" to another resource by having a relation to it).
+Such policy defines restrictions for a property between a _source_ and a _target_ resource.
+The syntax is slightly different than in the case of a standard policy:
+
+- id, principal, resource, tenant_id: as in the case of a standard policy
+- action: Must be `attach`
+- effect: If the condition in the policy allows for the access, or denies it.
+  Possible values (case insensitive, default is `allow`):
+    - `allow`: policy denies if source or target conditions are false
+    - `deny`: policy denies if source and target conditions are true
+- resource: Like in the case of a standard policy, but considers the source resource
+- relation_property: Name of the relation property the policy is applied to.
+  Can be `*`, then it applies to all properties of the source resource
+- target_condition: Condition for the target resource (for syntax, see the 'Condition' section)
+
+Attach policies are checked during creation/update of the _source_ resource.
+An attach policy is applied only if relation field is non-nil.
+If any of the policies deny access, then modification is forbidden and a 404 response is returned.
+In contrast to standard policies, the fact that no attach policies apply to the resource does not restrict access to the resource.
+
+An example:
+
+```yaml
+- action: '__attach__'
+  id: attach_if_accessible
+  effect: allow
+  principal: Member
+  relation_property: some_relation_id
+  target_condition:
+  - or:
+    - is_owner
+    - match:
+        property: accessibility
+        type: eq
+        value: everybody
+  resource:
+    path: /v2.0/attacher.*
+- action: '__attach__'
+  id: deny_if_blocked
+  effect: deny
+  principal: Member
+  relation_property: some_relation_id
+  target_condition:
+  - and:
+    - is_owner
+    - match:
+        property: block_flag
+        type: eq
+        value: true
+  resource:
+    path: /v2.0/attacher.*
 ```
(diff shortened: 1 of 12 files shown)


tests/test_schema.yaml
```diff
@@ -370,6 +370,42 @@ policies:
...
+  id: allow_attach_target
+  principal: Member
+  resource:
+    path: /v2.0/attach_target.*
+- action: 'attach'
+  id: attach_if_accessible
+  effect: allow
+  principal: Member
+  source:
+    resource_schema_id: attacher
```
morrisson: source schema_id can be specified by resource.path (same policy can be applied to multiple types), so I do not think we need to add this property.

I will not stick to above idea, but it might be worth to consider.

schema/policy.go
```diff
@@ -260,12 +277,73 @@ func NewPolicy(raw interface{}) (*Policy, error) {
 		return nil, err
 	}
 
-	if err := policy.precomputeConditions(); err != nil {
-		return nil, err
+	// TODO: Move validation somewhere else?
+	if policy.Action == ActionAttach {
+		// Source property is required
+		source, hasSource := typeData["source"].(map[string]interface{})
```
morrisson: Maybe it's better to convert to struct.

docs/policy.md
```diff
@@ -263,3 +263,58 @@ In an example below admin has access to all methods in all paths except `delete`
...
+The syntax is slightly different than in the case of a standard policy:
+
+- id, principal, resource, tenant_id: as in the case of a standard policy
+- action: Must be "allow"
+- effect: If the condition in the policy allows for the access, or denies it.
+  Possible values (case insensitive, default is "allow"):
+    - "allow": policy denies if source or target conditions are false
+    - "deny": policy denies if source and target conditions are true
+- resource : Like in the case of a standard policy, but considers the source resource
+- relation_property: Name of the relation property the policy is applied to
```
marcin-ptaszynski: can we wildcard this value? Would be nice to have some defaults.

piotr-dulikowski:
This version is ready for review. Summary of the changes:
- Removed the "schema_id" field
- Removed the source condition field (its meaning was unclear to me, and don't think it is necessary for implementing tenant isolation)
- Updated documentation

tests/test_schema.yaml
piotr-dulikowski: I removed this field.

schema/policy.go
```diff
@@ -34,6 +35,8 @@ const (
 	ActionUpdate = "update"
 	// ActionDelete allows to delete a resource
 	ActionDelete = "delete"
+	// ActionAttach allows a resource to have a relation to another resource
+	ActionAttach = "attach"
```
marcin-ptaszynski: I'm afraid of potential custom action conflict - could we \_\_attach\_\_ instead?

docs/policy.md
piotr-dulikowski: Added possibility to specify a wildcard `relation_property`.

schema/policy.go
piotr-dulikowski: Changed "attach" to "__attach__".

server/resources/resource_management.go
```diff
@@ -1013,6 +1069,86 @@ func ActionResource(context middleware.Context, dataStore db.DB, identityService
...
+
+	errorNotFound := func() error {
+		return goext.NewErrorNotFound(errors.Errorf("Related resource of ID \"%s\" not found", relatedResourceID))
+	}
+
+	if err != nil {
+		if policy.IsDeny() {
+			return nil
+		} else {
+			return errorNotFound()
```
morrisson: No related resource means user's input is incorrect. Therefore, `BadRequest()` should be better.

server/resources/resource_management.go
```diff
@@ -1013,6 +1069,86 @@ func ActionResource(context middleware.Context, dataStore db.DB, identityService
...
+		}
+	}
+
+	err = otherCond.ApplyPropertyConditionFilter(schema.ActionAttach, relatedRes.Data(), nil)
+	if policy.IsDeny() {
+		if err == nil {
+			return errorNotFound()
```
morrisson: Resource is available but the request is forbidden due to the policy. Therefore, `Forbidden()` should be better. (However, `BadRequest()` might be safer because error code which is different from above may imply the resource existence.)

docs/policy.md
```diff
@@ -263,3 +263,59 @@ In an example below admin has access to all methods in all paths except `delete`
...
+
+## Attach policy
+
+Attach policy defines allowed values for a relation property
+(a resource "attaches" to another resource by having a relation to it).
+Such policy defines restrictions for a property between a _source_ and a _target_ resource.
+The syntax is slightly different than in the case of a standard policy:
+
+- id, principal, resource, tenant_id: as in the case of a standard policy
+- action: Must be "__attach__"
```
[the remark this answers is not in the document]
p-kozlowski: fixed, same in other places

server/resources/resource_management.go
p-kozlowski: changed, also modified the error construction, so that the callstack will point to the correct code path

server/resources/resource_management.go
p-kozlowski: let's go the secure way. changed to `BadRequest`

p-kozlowski:
@morrisson, remarks applied, squashed & rebased. please merge at your convenience

docs/policy.md
```diff
@@ -263,3 +263,59 @@ In an example below admin has access to all methods in all paths except `delete`
...
+
+## Attach policy
+
+Attach policy defines allowed values for a relation property
+(a resource "attaches" to another resource by having a relation to it).
+Such policy defines restrictions for a property between a _source_ and a _target_ resource.
+The syntax is slightly different than in the case of a standard policy:
+
+- id, principal, resource, tenant_id: as in the case of a standard policy
+- action: Must be `attach`
```
morrisson: nit: Must be `__attach__`

PR #671, opened 2018-07-13
Print trace ID in request logs & transaction logs

p-kozlowski:
The main point of this PR was to include the request `trace_id` (an UUID) in the request logs (eg. `Started GET...` or `Completed 200 OK`) and transaction logs (eg. `Created transaction...`, `Executing SQL query ...`). However, to make it feasible, some large-scale refactorings (not rewrites!) were necessary.

I suggest reviewing commit-by-commit.
1.  "Create context.Context and trace_id in middleware layers" is the essence: it moves creating trace ID and context from goext to middleware
2. "Database interface refactoring" removes dependencies from `db` package to avoid dependency cycles and introduces functional options paradigm for easier transaction parametrization
3. "Log trace_id in middleware and transaction" introduces a new type `tracingLogger` and uses it it goext and transactions
4. "Simplify Transaction interface" is a follow up of commit 2. - it further simplifies the `transaction` interface - about half of all methods has been removed
5. "Exec should clear the cache" fixes bugs in `cached_transaction` - the cache was not cleared on Exec and new contexts were created instead of using a context passed by parameter


db/db.go
```diff
@@ -16,31 +16,23 @@
...
 
-var errNoSchemasInManager = errors.New("No schemas in Manager. Did you remember to load them?")
-
 //DB is a common interface for handing db
 type DB interface {
 	Connect(string, string, int) error
 	Close()
-	Begin() (transaction.Transaction, error)
-	BeginTx(ctx context.Context, options *transaction.TxOptions) (transaction.Transaction, error)
+	Begin(options ...transaction.Option) (transaction.Transaction, error)
```
piotr-dulikowski: Nitpick: I think I prefer the name `BeginTx`; `Begin` does not really tell what it begins.

extension/goext/database.go
```diff
@@ -30,7 +30,7 @@ type DbOptions struct {
 type IDatabase interface {
 	// Begin starts a new transaction
 	Begin() (ITransaction, error)
-	// BeginTx starts a new transaction with options
+	// Begin starts a new transaction with options
```
piotr-dulikowski: Should be `BeginTx`?

extension/goplugin/environment.go
```diff
@@ -691,7 +619,7 @@ func (env *Environment) Clone() extension.Environment {
 		databaseImpl: env.databaseImpl.Clone(),
 
 		name:    env.name,
-		traceID: newTraceID(),
+		traceID: fmt.Sprintf("%s-clone", env.traceID),
```
morrisson: If env is cloned multiple times,  multiple `<root_trace_uuid>-clone` envs may be produced so it's hard to distinguish each other. How about `<root_trace_uuid>-clone-<some_rand>`?

db/db.go
p-kozlowski: sure, done

extension/goext/database.go
p-kozlowski: fixed

extension/goplugin/environment.go
```diff
@@ -435,12 +439,8 @@ func dumpStackTrace(logger goext.ILogger, err error) {
 }
 
 func handleEventForEnv(env IEnvironment, event string, requestContext map[string]interface{}) error {
-	if !hasInterrupt(requestContext) {
-		interrupt := newInterrupt(env, event, requestContext)
-		defer func() {
-			interrupt.Cleanup()
-		}()
-	}
+	cleanup := addTimeout(env, event, requestContext)
```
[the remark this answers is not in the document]
p-kozlowski: What was actually done is:
1. Create a context.Context in the middleware layers, cancel it when peer disconnects.
2. For each event, get the timeout and decorate the parent context `WithTimeout`

Effectively we're building a tree(*) of contexts. Canceling any of those will cancel the whole subtree: cancels propagate down, but not up. Contexts are thread-safe, so canceling a parent and a child is not an issue.

The code passed automatic tests in https://github.com/cloudwan/gohan/blob/master/extension/goplugin/environment_test.go#L339.

@morrisson have you got any specific concerns to address? If so, I'll be happy to add automatic tests for those cases.

Side note: contexts are not perfect, but I think it's the best we've got (I'd kind of agree with https://faiface.github.io/post/context-should-go-away-go2/), definitely better and simpler to analyze than hand-crafted solutions (cf. https://github.com/cloudwan/gohan/blob/master/extension/otto/gohan_sync.go#L66)

(*) if no goroutines are started it's actually brought down to a list, but the argument holds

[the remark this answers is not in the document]
morrisson: Thank you for your clarification. I got it.

extension/goplugin/environment.go
morrisson: How about this?

p-kozlowski: sorry, didn't manage to complete it yesterday. 3 random decimal digits added

p-kozlowski:
will squash & merge after getting one more approve

PR #681, opened 2018-08-30
Domain-scoped tokens

piotr-dulikowski:
~~**Please do not merge yet**, this is still work in progress.~~

### DIFF
--- a/schema/policy.go
+++ b/schema/policy.go
@@ -20,6 +20,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/cloudwan/gohan/extension/goext/filter"
 	"github.com/cloudwan/gohan/util"
 	"github.com/pkg/errors"
 )
@@ -48,34 +49,52 @@ const (
 	globalRegexp = ".*"
 
 	onlyOneOfTenantIDTenantNameError = "Only one of [tenant_id, tenant_name] should be specified"
+
+	adminRole = "admin"
 )
 
 // AllActions are all possible actions
 var AllActions = []string{ActionCreate, ActionRead, ActionUpdate, ActionDelete}
 
-func newTenant(tenantID, tenantName string) tenant {
+var adminPolicy = func() *Policy {
+	policy, err := NewPolicy(map[string]interface{}{
+		"action": "*",
+		"effect": "allow",
+		"id": "admin_statement",
+		"principal": "admin",
+		"resource": map[string]interface{}{
+			"path": ".*",
+		},
+	})
+	if err != nil {
+		panic(errors.Errorf("Failed to create the admin policy: %v", err))
+	}
+	return policy
+}()
+
+func newTenantMatcher(tenantID, tenantName string) tenantMatcher {
 	tenantIDRegexp, _ := getRegexp(tenantID)
 	tenantNameRegexp, _ := getRegexp(tenantName)
-	return tenant{ID: tenantIDRegexp, Name: tenantNameRegexp}
+	return tenantMatcher{ID: tenantIDRegexp, Name: tenantNameRegexp}
 }
 
-// Tenant ...
-type tenant struct {
+// tenantMatcher matches given tenant
+type tenantMatcher struct {
 	ID   *regexp.Regexp
 	Name *regexp.Regexp
 }
 
-func (t *tenant) equal(t2 tenant) bool {
+func (t *tenantMatcher) equal(t2 tenantMatcher) bool {
 	idMatch := t.ID.MatchString(t2.ID.String()) || t2.ID.MatchString(t.ID.String())
 	nameMatch := t.Name.MatchString(t2.Name.String()) || t2.Name.MatchString(t.Name.String())
 	return idMatch && nameMatch
 }
 
-func (t *tenant) notEqual(t2 tenant) bool {
+func (t *tenantMatcher) notEqual(t2 tenantMatcher) bool {
 	return !t.equal(t2)
 }
 
-func (t tenant) String() string {
+func (t tenantMatcher) String() string {
 	return fmt.Sprintf("%s (%s)", t.Name.String(), t.ID.String())
 }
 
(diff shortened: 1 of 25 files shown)


schema/policy.go
```diff
@@ -137,36 +137,64 @@ type resourceFilter struct {
...
+		roles:       roles,
+		authToken:   authToken,
+		catalog:     catalog,
+	}
 }
 
-//NewAuthorization is a constructor for auth info
-func NewAuthorization(tenantID, tenantName, authToken string, roleIDs []string, catalog []*Catalog) Authorization {
+//NewScopedToDomainAuthorization constructs auth info when user is scoped to domain
+func NewScopedToDomainAuthorization(domain Domain, authToken string, roleIDs []string, catalog []*Catalog) Authorization {
```
p-kozlowski: code duplication, can we extract the common part? maybe with Functional Options, or a Builder?

schema/policy.go
```diff
@@ -671,25 +741,32 @@ func (policy *Policy) GetOtherResourceCondition() *ResourceCondition {
...
-func addCustomFilters(f map[string]interface{}, tenantId string, conditionFilters *conditionFilter) {
+func addCustomFilters(f map[string]interface{}, auth Authorization, conditionFilters *conditionFilter) {
 	if conditionFilters == nil {
 		return
 	}
 	filters := make([]map[string]interface{}, 0)
 	if conditionFilters.isOwner {
-		filters = append(filters, map[string]interface{}{"property": "tenant_id", "type": "eq", "value": tenantId})
+		switch auth.ScopingType() {
+		case ScopedToTenant:
```
p-kozlowski: design issue: `if`s (or this `switch`) on `ScopedToTenant` are present in several places. instead of `if`ing, could we take advantage of polymorphism or decorate Domain scoped authorization with additional Tenant scope features?

schema/policy.go
```diff
@@ -137,36 +137,64 @@ type resourceFilter struct {
...
+	ScopedToDomain ScopingType = "to_domain"
+)
+
 //Authorization interface
 type Authorization interface {
 	TenantID() string
 	TenantName() string
+	DomainID() string
+	DomainName() string
 	AuthToken() string
```
p-kozlowski: it looks like `AuthToken` and `Catalog` are never actually read (to be confirmed). Could we remove these fields and any associated logic?

server/resources/resources_suite_test.go
```diff
@@ -42,6 +42,8 @@ const (
 	adminTenantID     = "fc394f2ab2df4114bde39905f800dc57"
 	memberTenantID    = "fc394f2ab2df4114bde39905f800dc57"
```
p-kozlowski: adminTenantID is the same as memberTenantID. my strong guess is that the values does not really matter, so could we use more human-friendly consts here? I propose just `adminTenantID = "adminTenantID"`. Rationale: if we see a human friendly name in a test error log it'll be easier to debug

p-kozlowski:
please run `goimports`, it complains in many places

server/middleware/cached_test.go
```diff
@@ -29,10 +29,14 @@ var _ = ginkgo.Describe("Cached identity service", func() {
...
-		tenantName = "tenant-name"
+		tenant := schema.Tenant{
+			ID:   "tenant-id",
+			Name: "tenant-name",
+		}
 		token = "token"
-		auth = schema.NewAuthorization(tenantID, tenantName, token, []string{}, []*schema.Catalog{})
+		auth = schema.NewAuthorizationBuilder().
+			WithTenant(tenant).
+			BuildScopedToDomain()
```
morrisson: Should this be `ScopedToTenant`?

server/resources/resource_management_test.go
```diff
@@ -32,40 +32,64 @@ import (
...
 		ctx = context_pkg.Background()
 
-		adminAuth = schema.NewAuthorization(adminTenantID, "admin", adminTokenID, []string{"admin"}, nil)
-		memberAuth = schema.NewAuthorization(memberTenantID, "demo", memberTokenID, []string{"Member"}, nil)
+		domainA := schema.Domain{
+			ID:   domainAID,
+			Name: "domainA",
+		}
+
+		adminAuth = schema.NewAuthorizationBuilder().
```
morrisson: If you mean `admin` here as cloud administrator or super-user to see resources across domains, we should use `admin_project` concept here for v3 auth. https://developer.openstack.org/api-ref/identity/v3/index.html#authentication-and-token-management

server/resources/resource_management_test.go
```diff
@@ -417,8 +459,8 @@ var _ = Describe("Resource manager", func() {
 					result := context["response"].(map[string]interface{})
 					number := context["total"].(uint64)
 					Expect(err).NotTo(HaveOccurred())
-					Expect(number).To(Equal(uint64(2)))
-					Expect(result).To(HaveKeyWithValue("tests", ConsistOf(adminResourceData, memberResourceData)))
+					Expect(number).To(Equal(uint64(3)))
+					Expect(result).To(HaveKeyWithValue("tests", ConsistOf(adminResourceData, memberResourceData, otherDomainResourceData)))
```
morrisson: If `adminAuth` means super-user, this result should be expected. However, if `adminAuth` is not super-user, the result should be only `adminResourceData` here, because token is tenant-scoped.

server/resources/resource_management_test.go
```diff
@@ -1035,19 +1101,65 @@ var _ = Describe("Resource manager", func() {
...
+
+				It("Should not create resource in other domain", func() {
+					err := resources.CreateResource(
+						context, testDB, fakeIdentity, currentSchema, otherDomainResourceData)
 					Expect(err).To(HaveOccurred())
 					_, ok := err.(resources.ResourceError)
 					Expect(ok).To(BeTrue())
 				})
+
+				It("Should fill in the domain_id", func() {
```
morrisson: Maybe we should also add "Should not create resource without tenant_id".

PR #685, opened 2018-09-24
BL-179 Version API

p-kozlowski:


morrisson:
We should update a doc about configuration and REST API.

PR #686, opened 2018-09-25
Add is_domain_owner property to policy

piotr-dulikowski:


schema/policy.go
```diff
@@ -56,22 +56,6 @@ const (
 // AllActions are all possible actions
 var AllActions = []string{ActionCreate, ActionRead, ActionUpdate, ActionDelete}
 
-var adminPolicy = func() *Policy {
```
piotr-dulikowski: Removed it, as such policy can be defined by adding "tenant_name: <tenant admin name>" to it and specifying it and putting it into external schema.

docs/policy.md
```diff
@@ -13,6 +13,10 @@ A policy has following properties.
   you can specify target resource using "path" and "properties"
 - condition : additional condition (see below)
 - tenant_id : regexp matching the tenant, defaults to ``.*``
+- is_domain_owner: type of the token's scope.
```
p-kozlowski: @morrisson these are the current ways to modify resource access:

1. schema/property/view
2. schema/property/permission
3. policy/effect
4. policy/properties vs. policy/blacklistProperties
5. policy/condition, where a user can specify additional SQL criteria based on custom schema properties
6. policy/condition/is_owner (besides, I initially thought that is_domain_owner is similar to this one)
7. operation/admin_role_as_super_user
8. is_domain_owner

To be honest, I think that this is already unmaintainable and impossible to reason about. For example, given a random property in a random schema, I could not image how a developer is supposed to tell if this property is visible/updateble to any given tenant without knowing the full system configuration.

In my opinion, we should break the backward compatibility (and break it hard, there's still 1.0.0 tag for any legacy users) and clean it once and for all. 

What do you think?

schema/policy.go
```diff
@@ -445,6 +446,20 @@ func NewPolicy(raw interface{}) (*Policy, error) {
 		return nil, fmt.Errorf(onlyOneOfTenantIDTenantNameError)
 	}
 
+	if isDomainOwnerRaw, ok := typeData["is_domain_owner"]; ok {
```
p-kozlowski: `NewPolicy` is now ~90 lines long. can we split it into smaller parts?

schema/policy.go
```diff
@@ -623,6 +638,12 @@ func (p *Policy) match(action, path string, auth Authorization) *Role {
 		return nil
 	}
 
+	if p.domainOwnerCondition == requireDomainOwner && !auth.isDomainScoped() {
```
p-kozlowski: Tell, Don't Ask. can we hide this logic inside `Auth` implementations?

schema/policy_test.go
```diff
@@ -204,6 +204,34 @@ var _ = Describe("Policies", func() {
...
+			})
+
+			It("should set to requireDomainOwner when is_domain_owner is true", func() {
+				testPolicy["is_domain_owner"] = true
+				policy, err := NewPolicy(testPolicy)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(policy.domainOwnerCondition).To(Equal(requireDomainOwner))
+			})
+
+			It("should set to requireDomainOwner when is_domain_owner is true", func() {
```
p-kozlowski: same TC name duplicated

schema/policy_test.go
```diff
@@ -421,7 +449,7 @@ var _ = Describe("Policies", func() {
 				It("should allow access be default", func() {
```
p-kozlowski: b**y**, besides this TC is misleading: the test code does not verify if access is indeed granted - now it depends on "policy1". by the way: what is policy1? can we come up with a better name?

docs/policy.md
morrisson: Hmm, I do not agree on your points.

We should not include point 1 and point 7 in this discussion, point 1 is related to gohan_webui handling and point 7 is our internal.

Without point 2, it can be controlled in policies. That's enough simple and monolithic, IMO. In this sense, I can agree on that point 2 has a duplication of point 4 so removing point 2 might make it simpler.

morrisson: @piotr-dulikowski As @p-kozlowski commented in point 6, I also expected that `is_domain_owner` is a condition keyword like `is_owner`.

Can you change the design? Or, if you think it's necessary to place `is_domain_owner` at the top level of the policy, can you elaborate on the reason?

piotr-dulikowski: @morrisson,
Semantics of `is_owner` are not similar to `is_domain_owner`.
- Properties like `principal`, `tenant_id`/`tenant_name`, `is_domain_owner` determine _if given policy is used_. This logic can be found in [policy.match](https://github.com/cloudwan/gohan/pull/686/files#diff-4166c24150929a3ed94d148ea2a87b80R655).
- The `condition` property defines _visibility_ of resources. If a given policy is decided to be used, `is_owner` may deny access to a resource, but it will not cause a different policy to be used instead. Note that, depending on the token scope, `is_owner` may filter resources either:
	- by both tenant_id and domain_id (in case of scoping to project),
	- only by domain_id (in case of scoping to domain),
	- not at all (in case of scoping to admin project).

piotr-dulikowski:
Superseded by https://github.com/cloudwan/gohan/pull/687

PR #689, opened 2018-10-05
[BL-267] Handle blacklisted tenant_id when creating resource

cierpuchaw:
BL-267: handle blacklisted tenant_id when creating resource.

We would always fail to create a resource when tenant_id is
blacklisted in a matching policy. This was due to the fact that we
filled the tenant_id field if it was not provided and was also defined
in the schema. That way we always matched the blacklist filter.

This fix is implemented by splitting Policy.Check method into two new
methods: CheckAccess and CheckPropertiesFilter. CheckAccess receives
the filled in tenant_id/domain_id as it requires that info to perform
authorization. CheckPropertiesFilter receives all the original
parameters without those additionally filled in. So if tenant_id was
not provided, it will not match against the blacklist/hidden filter.
Updating resource was not performing any tenant_id/domain_id filling,
so it still used the Check method (which is now implemented using the
extracted methods) as it leads to simpler code. And to smaller diff.

Added HasPropertyId method to Schema as a simple wrapper of
GetPropertyById, to simplify those call sites that do not need to
access the property, just to check its existence. Added tests for both
while at it.

### DIFF
--- a/server/resources/resource_management_test.go
+++ b/server/resources/resource_management_test.go
@@ -1270,22 +1270,47 @@ var _ = Describe("Resource manager", func() {
 					auth = memberAuth
 				})
 
-				It("Should create own resource when tenant_id isn't provided and not fill it in",
+				unpackResponseFrom := func(context middleware.Context) map[string]interface{} {
+					result := context["response"].(map[string]interface{})
+					response, ok := result[schemaID]
+					Expect(ok).To(BeTrue())
+					return response.(map[string]interface{})
+				}
+
+				It("Should create own resource when tenant_id isn't provided and fill it in DB",
 					func() {
 						err := resources.CreateResource(
 							context, testDB, fakeIdentity, currentSchema, map[string]interface{}{
 								"id":        memberResourceID,
 								"domain_id": domainAID,
 							})
-
 						Expect(err).NotTo(HaveOccurred())
 
-						result := context["response"].(map[string]interface{})
-						resource, ok := result[schemaID]
-						Expect(ok).To(BeTrue())
-						Expect(resource).To(HaveKeyWithValue("id", memberResourceID))
-						Expect(resource).To(HaveKeyWithValue("domain_id", domainAID))
-						Expect(resource).NotTo(HaveKeyWithValue("tenant_id", auth.TenantID()))
+						delete(context, "response")
+						delete(context, "resource")
+
+						err = resources.GetSingleResource(context, testDB, currentSchema,
+							memberResourceID)
+						Expect(err).NotTo(HaveOccurred())
+
+						response := unpackResponseFrom(context)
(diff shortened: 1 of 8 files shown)


server/resources/resource_management_test.go
```diff
@@ -1252,6 +1259,66 @@ var _ = Describe("Resource manager", func() {
...
+							})
+
+						Expect(err).NotTo(HaveOccurred())
+
+						result := context["response"].(map[string]interface{})
+						resource, ok := result[schemaID]
+						Expect(ok).To(BeTrue())
+						Expect(resource).To(HaveKeyWithValue("id", memberResourceID))
+						Expect(resource).To(HaveKeyWithValue("domain_id", domainAID))
+						Expect(resource).NotTo(HaveKeyWithValue("tenant_id", auth.TenantID()))
```
morrisson: `tenant_id` should be *filled* based on token after the policy check.

server/resources/resource_management_test.go
```diff
@@ -1252,6 +1259,66 @@ var _ = Describe("Resource manager", func() {
...
+							"id":        memberResourceID,
+							"tenant_id": memberTenantID,
+							"domain_id": domainAID,
+						})
+
+					Expect(err).To(HaveOccurred())
+
+					resErr, ok := err.(resources.ResourceError)
+					Expect(ok).To(BeTrue())
+					Expect(resErr.Problem).To(Equal(resources.Unauthorized))
```
zimnx: shouldn't it be a bad request?

server/resources/resource_management_test.go
piotr-dulikowski: `tenant_id` should be filled and saved in DB, but not returned in the response, because it is blacklisted in the policy for Member. It can be checked by fetching it directly from db, or by adding a policy for e.g. admin and using it to get the resource.

server/resources/resource_management_test.go
cierpuchaw: This test is simply documenting already existing behavior.

Changing this is trivial but results in 2 failures in `server/server_test.go`. If changing those expectations is fine, I'll do it. Waiting for go-ahead.

Failing tests:
server_test.go:553:"Should not create exposed with incorrect fields"
server_test.go:555:"Should not create forbidden with incorrect fields"

server/resources/resource_management_test.go
cierpuchaw: Added a new test to show that `tenant_id` is in fact filled in and saved to the DB.

morrisson: Oh, I see. Thanks for the clarification.

server/resources/resource_management_test.go
cierpuchaw: So I guess this is not that important. If it's still desired I can create another PR with this simple change.

cierpuchaw:
Not a big fan of squashing, but done as requested. Thanks for the review guys!

PR #692, opened 2018-10-24
Remove vendored golang.org/x/net/trace

zimnx:
This package is not used by Gohan. 
Vendoring it causes that extensions that imports gprc package panics on startup because of https://github.com/etcd-io/etcd/issues/9357

```
~/work/src/github.com/cloudwan/gohan on  remove-x-net-trace-from-vendor ⌚ 15:12:18         
$ git grep "net/trace"                                                                       
vendor/github.com/coreos/etcd/embed/serve.go:   "golang.org/x/net/trace"                     
vendor/github.com/coreos/etcd/etcdserver/api/v3client/v3client.go:// (e.g., x/net/trace), causing ctx.String() to throw data races.                                                       
vendor/google.golang.org/grpc/clientconn.go:    "golang.org/x/net/trace"                     
vendor/google.golang.org/grpc/server.go:        "golang.org/x/net/trace"                     
vendor/google.golang.org/grpc/stream.go:        "golang.org/x/net/trace"                     
vendor/google.golang.org/grpc/trace.go: "golang.org/x/net/trace"                             
vendor/google.golang.org/grpc/trace.go:// EnableTracing controls whether to trace RPCs using the golang.org/x/net/trace package.                                                          


```


tools/dev_setup.sh
```diff
@@ -5,4 +5,4 @@ go get -u golang.org/x/lint/golint
 go get -u github.com/gordonklaus/ineffassign
 go get -u github.com/jteeuwen/go-bindata/go-bindata
 go get -u github.com/client9/misspell/cmd/misspell
-go get -u github.com/vektra/mockery/cmd/...
+go get github.com/golang/mock/gomock
```
jkmar: why not -u?

PR #695, opened 2018-10-30
is_domain_owner condition implementation

piotr-dulikowski:
This implements is_domain_owner condition, which allows filtering resources by current user's domain id.

### DIFF
--- a/schema/policy.go
+++ b/schema/policy.go
@@ -40,6 +40,7 @@ const (
 	ActionAttach = "__attach__"
 
 	conditionIsOwner       = "is_owner"
+	conditionIsDomainOwner = "is_domain_owner"
 	conditionTypeBelongsTo = "belongs_to"
 	conditionProperty      = "property"
 	conditionOr            = "or"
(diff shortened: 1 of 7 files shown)


schema/policy_test.go
```diff
@@ -928,6 +984,96 @@ var _ = Describe("Policies", func() {
...
+
+			It("Should omit creating a filter for is_domain_owner when domain_id is missing from schema", func() {
+				tenant := Tenant{ID: "test", Name: "test"}
+				testAuth = NewAuthorizationBuilder().
+					WithTenant(tenant).
+					WithRoleIDs("admin").
+					BuildScopedToTenant()
+
+				schema := getSchema("network")
+				testPolicy["condition"] = []interface{}{
```
cierpuchaw: This is the same condition as in the test above. Can we extract and name it?

schema/policy.go
```diff
@@ -324,12 +334,19 @@ func (auth *DomainScopedAuthorization) IsAdmin() bool {
...
 }
 
-func (auth *DomainScopedAuthorization) getResourceFilters(schema *Schema) map[string]interface{} {
-	return maybeEmptyAndFilter(getFilterByPropertyIfPresent(schema, "domain_id", auth.DomainID())...)
+func (auth *DomainScopedAuthorization) getDomainCustomFilter(schema *Schema) []map[string]interface{} {
+	return getFilterByPropertyIfPresent(schema, "domain_id", auth.DomainID())
+}
+
+func (auth *DomainScopedAuthorization) getTenantCustomFilter(schema *Schema) []map[string]interface{} {
+	return []map[string]interface{}{}
```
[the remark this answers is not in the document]
piotr-dulikowski: Done, changed to filter.FilterElem

schema/policy_test.go
piotr-dulikowski: Done

PR #696, opened 2018-11-13
Introduce admin_statement for admin-scoped token into gohan.json

piotr-dulikowski:
With scope property support added, admin_statement policy can now be reintroduced into the gohan.json file.

### DIFF
--- a/schema/policy_test.go
+++ b/schema/policy_test.go
@@ -525,6 +525,44 @@ var _ = Describe("Policies", func() {
 			})
 		})
 
+		Describe("Built-in policies in gohan.json", func() {
+			const resourcePath = "/any/path/without/a/custom/policy"
+			var adminAuth, domainAdminAuth, memberAuth Authorization
+
+			BeforeEach(func() {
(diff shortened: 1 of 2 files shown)


schema/policy_test.go
```diff
@@ -525,6 +525,43 @@ var _ = Describe("Policies", func() {
...
+					WithDomain(Domain{ID: "domainID", Name: "domainName"}).
+					BuildScopedToDomain()
+				memberAuth = NewAuthorizationBuilder().
+					WithRoleIDs("Member").
+					WithTenant(Tenant{ID: demoTenantID, Name: "demo"}).
+					BuildScopedToTenant()
+			})
+
+			It("should allow operations for admin-scoped token", func() {
+				p, _ := manager.PolicyValidate("read", "/v2.0/networks", adminAuth)
```
p-kozlowski: Is `/v2.0/networks` somehow meaningful in this test? If not, I'd go for something more descriptive, like `/any/path/without/a/custom/policy`

PR #697, opened 2018-11-14
CW-870 Always add a context when creating a Context

p-kozlowski:


extension/goext/schemas.go
```diff
@@ -103,7 +103,9 @@ func OptionOffset(offset uint64) OptionPaginator {
 
 // MakeContext creates an empty context
```
cierpuchaw: It's not empty anymore :) Maybe `default`?

extension/goext/schemas.go
```diff
@@ -103,7 +103,9 @@ func OptionOffset(offset uint64) OptionPaginator {
...
 func MakeContext() Context {
-	return make(map[string]interface{})
+	return map[string]interface{}{
+		"context": context.Background(),
```
p-kozlowski: Reviewers, I'd like to hear your voices about this fix. This PR solves the issue with `goext.Contexts` not having `context.Contexts` and `panic`ing when trying to access it. But there's more to it: 

- if a user creates a `goext.Context` by hand (eg. `c := goext.Context{}`) the issue remains.
- creating a `goext.Context` using `MakeContext()` when a `context.Context` is at hand is a mistake: the parent `context.Context` won't be used, so `HandlingEvent` in case of a timeout may take longer then expected (child timeout will be used, but not the parent timeout)

I'm a bit uncertain if we should merge this as-is.

michalzurawski: @zimnx what's the reason of passing multiple context.Context? especially if you want panic when there are multiple - in such case you moving this check from static compiler analysis to runtime

cierpuchaw: I'm also in favor of having this checked at compile time. Since golang does not support function overloading, maybe `MakeContext()` (or `MakeDefaultContext()`) and `MakeContextFrom(ctx)` would do the job. Either way, we do have to go through all call sites...

PR #699, opened 2018-11-17
Tenant isolation of nested properties

piotr-dulikowski:


schema/property.go
```diff
@@ -83,36 +88,64 @@ func NewPropertyFromObj(id string, rawTypeData interface{}, required bool) *Prop
...
 		nullable = true
 	}
 	sqlType, _ := typeData["sql"].(string)
 	indexed, _ := typeData["indexed"].(bool)
-	Property := NewProperty(id, title, description, typeID, format, relation, relationColumn, relationProperty,
-		sqlType, unique, nullable, cascade, properties, defaultValue, indexed)
-	return &Property
+
+	var items *Property
+	if itemsRaw, hasItems := typeData["items"]; hasItems {
```
morrisson: It looks `typeID == "array"` check is also required.

schema/property.go
```diff
@@ -83,36 +88,64 @@ func NewPropertyFromObj(id string, rawTypeData interface{}, required bool) *Prop
...
+		items = NewPropertyFromObj("[]", itemsRaw, true)
```
morrisson: `"[]"` should be `const`

server/resources/resource_management.go
```diff
@@ -1170,25 +1166,89 @@ func extendFilterByTenantAndDomain(
...
+			if currChildren[i].ID == key {
+				currProp = &currChildren[i]
+				currChildren = currProp.Properties
+
+				for j, dataPiece := range dataSet {
+					dataPieceMap, _ := dataPiece.(map[string]interface{})
+					dataSet[j], _ = dataPieceMap[key]
+				}
+
+				continue outer
```
michalzurawski: IMHO changing first `continue outer` to `continue`, this one to `childFound = true` and `break` to `if !childFound { break }` would be clearer

schema/property.go
```diff
@@ -58,7 +65,7 @@ func NewProperty(id, title, description, typeID, format, relation, relationColum
 
 //NewPropertyFromObj make Property  from obj
 func NewPropertyFromObj(id string, rawTypeData interface{}, required bool) *Property {
-	typeData := rawTypeData.(map[string]interface{})
+	typeData, _ := rawTypeData.(map[string]interface{})
```
michalzurawski: nit: there's no point in consuming this boolean value and then ignore it. If the reason for this change was consistency please consider removing it from the rest rather than adding it here.

schema/property.go
```diff
@@ -28,14 +31,17 @@ type Property struct {
...
 }
 
 //PropertyMap is a map of Property
 type PropertyMap map[string]Property
 
+const ItemPropertyID = "[]"
+
 //NewProperty is a constructor for Property type
-func NewProperty(id, title, description, typeID, format, relation, relationColumn, relationProperty, sqlType string, unique, nullable, onDeleteCascade bool, properties map[string]interface{}, defaultValue interface{}, indexed bool) Property {
+func NewProperty(id, title, description, typeID, format, relation, relationColumn, relationProperty, sqlType string, unique, nullable, onDeleteCascade bool, properties []Property, items *Property, defaultValue interface{}, indexed bool) Property {
```
michalzurawski: 16 parameters is 'bit' too much, especially that there are often filled with `nil`. Please consider removing all but this necessary (like id, typeID) and some builder instead

piotr-dulikowski: Removed in favor of the `PropertyBuilder`

schema/property.go
piotr-dulikowski: Done

schema/property.go
piotr-dulikowski: Done

schema/property.go
piotr-dulikowski: Reverted to how it was before

server/resources/resource_management.go
piotr-dulikowski: Refactored into something akin to iterator pattern

PR #702, opened 2018-12-06
Gohan template improvements

piotr-dulikowski:
- Add schemaMap as a gohan template argument
- Include array properties in GetAllPropertiesFullyQualifiedMap function
- Add filtering of policies by token scope in gohan template command

### DIFF
--- a/cli/template.go
+++ b/cli/template.go
@@ -181,6 +181,7 @@ func doTemplate(c *cli.Context) {
 		}
 	}
 	schemas := manager.OrderedSchemas()
+	schemasMap := manager.Schemas()
 
 	if err != nil {
 		util.ExitFatal(err)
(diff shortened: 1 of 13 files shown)


cli/template.go
```diff
@@ -363,16 +370,31 @@ func isMatchingPolicy(action schema.Action, policy *schema.Policy) bool {
...
+
+		matchesScope := false
+		for _, tokenType := range scope {
+			if policy.HasTokenType(tokenType) {
+				matchesScope = true
+				break
+			}
+		}
+
+		if !matchesScope {
```
p-kozlowski: let's simplify the control flow a little bit:
```go
func matches(policy, principal, scopes) bool {
  if !matchesPrincipal(policy, principal) {
    return false
  }
  if !matchesScopes(policy, scopes) {
    return false
  }
  return true
}

func filterPolicies(...) {
  ...
  for _, policy := range policies {
    if matches(policy, principal, scopes) {
      append(...)
    }
}
```

cli/template.go
```diff
@@ -363,16 +370,31 @@ func isMatchingPolicy(action schema.Action, policy *schema.Policy) bool {
...
 	var matchedPolicies []*schema.Policy
 	for _, policy := range policies {
-		if policy.Principal == principal {
-			matchedPolicies = append(matchedPolicies, policy)
+		if policy.Principal != principal {
+			continue
 		}
+
+		matchesScope := false
+		for _, tokenType := range scope {
```
p-kozlowski: naming: we're mixing "scope" and "tokenType". I suggest `for _, scope := range scopes`

morrisson:
but need an investigation of circleci failure

morrisson:
Please squash commits before merge

PR #703, opened 2018-12-13
Repair broken "scopes" option in gohan template

piotr-dulikowski:
During review of "Gohan template improvements" PR, I considered changing "scope" option to "scopes". I forgot to change it back in one place, which made "scopes" be always empty, and all policies were filtered out.

### DIFF
--- a/cli/template.go
+++ b/cli/template.go
@@ -16,6 +16,20 @@ import (
 	"github.com/vattle/sqlboiler/strmangle"
 )
 
+const (
+	configFileFlagName           = "config-file"
+	templateFlagName             = "template"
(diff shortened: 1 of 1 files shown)


cli/template.go
```diff
@@ -194,7 +194,7 @@ func doTemplate(c *cli.Context) {
 	}
 
 	scopes := []schema.Scope{}
-	for _, tokenType := range c.StringSlice("scopes") {
+	for _, tokenType := range c.StringSlice("scope") {
```
p-kozlowski: how about a const for this magic string?

PR #706, opened 2019-01-02
[CW-928] Check policy before create or update

jkmar:


server/resources/policy_filter_test.go
```diff
@@ -0,0 +1,266 @@
+package resources
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Policy filter", func() {
+	DescribeTable("Apply filter to resource",
```
cierpuchaw: I like those test cases. But 2 largest ones barely fit on a screen. Can we compact them a bit?

For example:
```
{
    "property": "a",
    "type":     "eq",
    "value":    1,
}
```
could become:
```
equals("a", 1)
```

server/resources/policy_filter.go
```diff
@@ -0,0 +1,114 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func applyPropertyFilter(resource map[string]interface{}, filter map[string]interface{}) bool {
+	actualValue, ok := resource[filter["property"].(string)]
+	if !ok {
+		return false
```
cierpuchaw: There is no test case for this.

PR #707, opened 2019-01-03
CW-997 Ip format validate 00 and 000

kuba2:


schema/formats_test.go
```diff
@@ -90,6 +90,83 @@ var _ = Describe("format checkers", func() {
 			result := formatChecker.IsFormat("::1/129")
 			Expect(result).To(Equal(false))
 		})
+
+		It("Should not pass - two zeros in octet IPv4", func() {
```
zimnx: Change all of these tests to Table tests. ref: https://onsi.github.io/ginkgo/#table-driven-tests

schema/formats.go
```diff
@@ -46,22 +47,53 @@ func (f macFormatChecker) IsFormat(input string) bool {
...
-	_, _, err := net.ParseCIDR(input)
-	return err == nil
+	ip, _, err := net.ParseCIDR(input)
+	if err != nil {
+		return false
+	}
+	if ip.To4() != nil {
+		return matchYangIPv4AddressPattern(input)
+	}
+	return true
```
zimnx: shouldn't it be false?

schema/formats.go
```diff
@@ -46,22 +47,53 @@ func (f macFormatChecker) IsFormat(input string) bool {
 	return match
 }
 
+func matchYangIPv4AddressPattern(input string) bool {
+	res, err := regexp.MatchString(`^(([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])(%[\p{N}\p{L}]+)?`, input)
```
zimnx: Have you handcrafted this regexp or took from some rfc?

schema/formats.go
kuba2: This checker works for IPv6 and IPv4. Returning false would break IPv6 format checking.

schema/formats.go
p-kozlowski: i suggest https://golang.org/pkg/regexp/#MustCompile for small efficiency gain

schema/formats.go
```diff
@@ -46,22 +47,53 @@ func (f macFormatChecker) IsFormat(input string) bool {
...
+
+	if err != nil {
+		return false
+	}
+	return res
+}
+
+func (f ipv4FormatChecker) IsFormat(input string) bool {
+	ip := net.ParseIP(input)
+	if !((ip != nil) && (ip.To4() != nil)) {
```
p-kozlowski: for simplicity in code reading, I suggest helper functions + changing the `if` condition a little bit:

```Go
func isValid(ip net.IP) bool {
   return ip != nil
}

func isV4(ip net.IP) bool {
   return ip.To4() != nil
}

// in this function
if !isValid(ip) || !isV4(ip) {
   return false
}
```

schema/formats.go
```diff
@@ -46,22 +47,53 @@ func (f macFormatChecker) IsFormat(input string) bool {
...
+	}
+	return true
 }
 
 func (f cidrOrIPv4FormatChecker) IsFormat(input string) bool {
 	cidrIP, _, cidrErr := net.ParseCIDR(input)
 	ip := net.ParseIP(input)
-	return (cidrErr == nil && cidrIP.To4() != nil) || (ip != nil && ip.To4() != nil)
+
+	if !((cidrErr == nil && cidrIP.To4() != nil) || (ip != nil && ip.To4() != nil)) {
```
p-kozlowski: again - can we simplify this condition?

schema/formats_test.go
p-kozlowski: this file is a bad example, but prefer test names that read like sentences. now it's "CIDR format checker / Should not pass - two zeros in octet IPv4". I suggest "CIDR format checker / should reject two zeros in octet IPv4" instead

schema/formats_test.go
```diff
@@ -90,6 +90,83 @@ var _ = Describe("format checkers", func() {
...
+			Expect(result).To(Equal(false))
+		})
+
+		It("Should not pass - three zeros in octet IPv4", func() {
+			result := formatChecker.IsFormat("127.000.0.1/24")
+			Expect(result).To(Equal(false))
+		})
+
+		It("Should not pass - two zeros in first octet IPv4", func() {
+			result := formatChecker.IsFormat("000.0.0.1/24")
```
p-kozlowski: test data does not match test name (should be "**00**.0...")

schema/formats.go
```diff
@@ -46,22 +47,53 @@ func (f macFormatChecker) IsFormat(input string) bool {
...
+		return matchYangIPv4AddressPattern(input)
+	}
+	return true
 }
 
 func (f cidrOrIPv4FormatChecker) IsFormat(input string) bool {
 	cidrIP, _, cidrErr := net.ParseCIDR(input)
 	ip := net.ParseIP(input)
-	return (cidrErr == nil && cidrIP.To4() != nil) || (ip != nil && ip.To4() != nil)
+
```
p-kozlowski: could we reuse other format checkers here? sth. like

```Go
func (f cidrOrIPv4FormatChecker) IsFormat(input string) bool {
    ipv4 := ipv4FormatChecker{}
    cidr := ipv4NetworkFormatChecker{}
    return ipv4.IsFormat(input) || cidr.IsFormat(input)
}
```

schema/formats.go
kuba2: Done

schema/formats.go
kuba2: I reused format checkers and the condition is no longer here.

schema/formats_test.go
kuba2: Good point. I changed descriptions to be more sentence like.

schema/formats.go
kuba2: Good point. Looks simpler that way.

schema/formats.go
kuba2: checkers are reused now.

schema/formats_test.go
kuba2: done

schema/formats.go
```diff
@@ -46,22 +47,53 @@ func (f macFormatChecker) IsFormat(input string) bool {
 	return match
 }
 
+func matchYangIPv4AddressPattern(input string) bool {
+	r := regexp.MustCompile(`^(([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])(%[\p{N}\p{L}]+)?`)
```
p-kozlowski: this regexp will be compiled whenever a validator is invoked, possibly many times during single request processing. can we make it a package-level variable?

schema/formats_test.go
```diff
@@ -51,158 +51,158 @@ var _ = Describe("format checkers", func() {
...
+			Entry("IPv4 cidr", "127.0.0.1/16"),
+			Entry("IPv4 network", "127.0.0.0/16"),
+			Entry("IPv6 cidr", "::1/16"),
+			Entry("IPv6 network", "fe80::9e17/64"),
+			Entry("wrong IPv4", "218.108.149.379"),
+			Entry("wrong IPv4 with three octets only", "218.149.179"),
+			Entry("IPv4 with two zeros in octet ", "127.00.0.1"),
+			Entry("IPv4 with three zeros in octet ", "127.000.0.1"),
+			Entry("IPv4 with two zeros in the first octet ", "00.3.0.1"),
+			Entry("IPv4 wiht three zeros in the first octet ", "000.4.0.1"),
```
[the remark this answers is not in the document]
kuba2: fixed

schema/formats.go
kuba2: made a package level variable

schema/formats.go
```diff
@@ -41,27 +42,59 @@ type textFormatChecker struct{}
...
+}
+
 func (f cidrFormatChecker) IsFormat(input string) bool {
-	_, _, err := net.ParseCIDR(input)
-	return err == nil
+	ip, _, err := net.ParseCIDR(input)
+	if err != nil {
+		return false
+	}
+	if isV4(ip) {
```
jkmar: there is also yang for ipv6:
https://tools.ietf.org/html/rfc6991#page-19

PR #710, opened 2019-01-09
CW-948 Expose enum values in Schema properties

p-kozlowski:


extension/goext/schemas.go
```diff
@@ -267,18 +267,21 @@ type ISchema interface {
...
 type Property struct {
-	ID       string
-	Title    string
-	Relation SchemaID
-	Type     string
+	ID         string
+	Title      string
+	Relation   SchemaID
+	Type       string
+	Enum       []string
```
morrisson: Just a note. I do not think we need to change this type, but enum is parsed in yaml parser then some of items will not be "string" type (e.g., integer, bool or nil). So, at least unit test should contain the schema yaml file which includes non-string values to check if no error happens.

p-kozlowski: It's a valid point. in fact, one of the test schemas uses ints for enum, this is why I'm converting all values using `fmt.Sprint`.
Let me change the field type to `[]interface{}`and let the clients do the appropriate conversion.

schema/property.go
```diff
@@ -199,6 +209,10 @@ func parseItems(itemsRaw interface{}) *Property {
 	}
 }
 
+func parseEnum(enumRaw interface{}) []interface{} {
+	return enumRaw.([]interface{})
```
zimnx: Is it always an array? What if someone specifies single value for enum? I don't know if it's possible, a test would be great.

p-kozlowski: enum is required to be an [array](https://github.com/cloudwan/gohan/blob/master/etc/schema/gohan.json#L197), nevertheless I'll add a test because AFAIR validation of nested properties may be broken in some places

extension/goplugin/schemas_test.go
```diff
@@ -154,32 +154,56 @@ var _ = Describe("Schemas", func() {
 	Context("Properties", func() {
 		It("Should get correct properties", func() {
 			Expect(testSchema.Properties()).To(Equal(
```
cierpuchaw: Nit: not a big fan of checking complex maps using one expectation.

p-kozlowski:
@tomasz-michalski mock fixed. in fact, the script for generating mocks needed an update as well

extension/goplugin/schemas_test.go
p-kozlowski: what do you suggest instead?

extension/goplugin/integration_tests/end_to_end_test.go
```diff
@@ -139,6 +140,18 @@ var _ = Describe("Environment", func() {
 		})
 	})
 
+	matchTestObject := func(actualResponse interface{}, expectedResponse map[string]interface{}) {
+		Expect(actualResponse).To(HaveKey("test"))
+		response := actualResponse.(map[string]interface{})["test"].(map[string]interface{})
+
+		// gomega does not handle nested slices comparison gracefully, we compare it manually
```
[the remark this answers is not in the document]
p-kozlowski: previously (when enum was a top-level string-typed), gomaga matched neither `nil` nor `[]string{}`. this was a quick-and-dirty (but at least documented) workaround.

according to other review remarks, I changed the type to `[]interface{}` and somehow now the matcher works... 

let's just revert to the previous, clean solution.

extension/goplugin/schemas_test.go
cierpuchaw: To check each "subtree" separately. That way we won't have to decode and compare large portions of text if something fails.
If you want to be sure that map is equal to sum of those "subtrees" and there are no unexpected ones, just check it's key count.

p-kozlowski: there you go

PR #712, opened 2019-01-09
Go 1.11 support and test stability

cierpuchaw:
This PR improves test stability: on my machine, server and otto tests failed all the time before because they used the same db file. It also fixes one issue reported by `go test` when using go 1.11.


cierpuchaw:
@zimnx and @piotr-dulikowski please notice that I've updated golang version in CI config after your approvals.

PR #713, opened 2019-01-10
[BL-314] Add possibility to load all schemas in tests

michalzurawski:


cli/cli.go
```diff
@@ -331,6 +331,7 @@ documentation for detail information about writing tests.`,
 			cli.StringFlag{Name: "run-test, r", Value: "", Usage: "Run only tests matching specified regex"},
 			cli.IntFlag{Name: "parallel, p", Value: runtime.NumCPU(), Usage: "Allow parallel execution of test functions"},
 			cli.StringFlag{Name: "type, t", Value: "", Usage: "Run only specific types of tests from a comma separated list (js,go); if not specified, all types of tests are run"},
+			cli.StringFlag{Name: "schemas-file, s", Value: "gohan.yaml", Usage: "File containing schemas to load"},
```
morrisson: For the generality, can we simply use config-file to get schemas?

extension/framework/framework.go
```diff
@@ -87,7 +87,11 @@ func TestExtensions(context *cli.Context) {
 	}
 	if runSoExt {
 		testFiles := getTestFiles(context.Args(), "so")
-		if err := gorunner.NewTestRunner(testFiles, context.Bool("verbose") || config != nil, context.String("run-test"), context.Int("parallel")).Run(); err != nil {
+		printAllLogs := context.Bool("verbose") || config != nil
+		testFilter := context.String("run-test")
+		workers := context.Int("parallel")
+		schemasFilePath := context.String("schemas-file")
+		if err := gorunner.NewTestRunner(testFiles, printAllLogs, testFilter, workers, schemasFilePath).Run(); err != nil {
```
cierpuchaw: Shouldn't all those string literals be named constants?

extension/framework/goplugin/runner/runner.go
```diff
@@ -265,3 +251,22 @@ func (testRunner *TestRunner) runSingle(t ginkgo.GinkgoTestingT, reporter *Repor
...
+	}
+
+	fn, ok := fnRaw.(func() []string)
+
+	if !ok {
+		return nil, fmt.Errorf("invalid signature of 'Schemas' export")
+	}
+	schemas := fn()
+	for i := range schemas {
+		schemas[i] = path + "/" + schemas[i]
```
cierpuchaw: `filepath.Join()`?

cli/cli.go
michalzurawski: true, thanks

extension/framework/framework.go
michalzurawski: done

extension/framework/framework.go
```diff
@@ -81,13 +90,18 @@ func TestExtensions(context *cli.Context) {
...
 			os.Exit(ret)
 		}
 	}
 	if runSoExt {
 		testFiles := getTestFiles(context.Args(), "so")
-		if err := gorunner.NewTestRunner(testFiles, context.Bool("verbose") || config != nil, context.String("run-test"), context.Int("parallel")).Run(); err != nil {
+		printAllLogs := context.Bool(FlagVerbose) || config != nil
+		testFilter := context.String(FlagRunTest)
+		workers := context.Int(FlagParallel)
+		schemas := config.GetStringList("schemas", []string{})
```
cierpuchaw: `config` is `nil` when tests are run without config flag and this panics.

michalzurawski: thanks, done

PR #716, opened 2019-01-23
Add health check tool

michalzurawski:
It runs on separate port and allows checking condition of etcd and db.

### DIFF
--- a/healthcheck/healthcheck.go
+++ b/healthcheck/healthcheck.go
@@ -1,0 +1,105 @@
+package healthcheck
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/cloudwan/gohan/db"
+	"github.com/cloudwan/gohan/db/transaction"
+	"github.com/cloudwan/gohan/log"
+	"github.com/cloudwan/gohan/sync"
+	"github.com/cloudwan/gohan/sync/etcdv3"
+	"github.com/cloudwan/gohan/util"
+)
+
+const (
+	healthCheckKey        = "healthcheck"
+	healthCheckEnabledKey = healthCheckKey + "/enabled"
+	healthCheckEtcdKey    = healthCheckKey + "/etcd_key"
+	healthCheckAddressKey = healthCheckKey + "/address"
+	healthCheckTimeoutKey = healthCheckKey + "/timeout"
+	defaultEtcdKey        = "/gohan"
+	defaultTimeout        = 5 * time.Second
+)
+
+type HealthCheck struct {
+	DataStore db.DB
+	Etcd      sync.Sync
+	etcdKey   string
+	address   string
+	timeout   time.Duration
+	logger    log.Logger
+}
+
+func NewHealthCheck(db db.DB, etcd sync.Sync, serverAddress string, config *util.Config) *HealthCheck {
+	if !config.GetBool(healthCheckEnabledKey, false) {
+		return nil
+	}
+	healthCheckAddress, err := getHealthCheckAddress(serverAddress, config)
+	if err != nil {
+		panic(err)
+	}
+	timeout := config.GetDuration(healthCheckTimeoutKey, defaultTimeout)
+	return &HealthCheck{
+		DataStore: db,
+		Etcd:      etcd,
+		address:   healthCheckAddress,
+		timeout:   timeout,
+		etcdKey:   config.GetString(healthCheckEtcdKey, defaultEtcdKey),
+		logger:    log.NewLogger(log.ModuleName(healthCheckKey)),
+	}
+}
+
+func (healthCheck *HealthCheck) Run() {
+	if healthCheck == nil {
+		return
+	}
+	healthCheckHandler := func(w http.ResponseWriter, req *http.Request) {
+		if err := healthCheck.isHealthy(); err == nil {
+			w.WriteHeader(http.StatusOK)
+		} else {
+			healthCheck.logger.Error("Health Check error: %s", err.Error())
+			w.WriteHeader(http.StatusServiceUnavailable)
+		}
+	}
+	http.HandleFunc("/healthcheck", healthCheckHandler)
+	go func() {
+		for {
+			err := http.ListenAndServe(healthCheck.address, nil)
+			if err != nil {
+				healthCheck.logger.Critical("Health Check server error %+v. Restarting", err)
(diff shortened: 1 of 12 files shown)


server/server.go
```diff
@@ -398,6 +401,18 @@ func NewServer(configFile string) (*Server, error) {
...
+	var gohanPort int
+	addressAndPort := strings.Split(server.address, ":")
+	if len(addressAndPort) != 2 {
+		panic("Incorrect gohan server address: " + server.address)
+	}
+	gohanPort, err = strconv.Atoi(addressAndPort[1])
+	if err != nil {
+		panic("Incorrect gohan server address: expected port number got " + addressAndPort[1])
+	}
+	healthCheckAddress := config.GetString("healtcheck/address", fmt.Sprintf("%s:%d", addressAndPort[0], gohanPort+1))
```
morrisson: comment 1: typo `healtcheck`

comment 2: How should we handle the case where healthcheck port is configured as the same value as server port?

healthcheck/healthcheck.go
```diff
@@ -0,0 +1,59 @@
...
+			healthCheck.logger.Error("healthcheck error: %s", err.Error())
+			w.WriteHeader(http.StatusServiceUnavailable)
+		}
+
+	}
+	http.HandleFunc("/healthcheck", healthCheckHandler)
+	go func() {
+		for {
+			err := http.ListenAndServe(healthCheck.address, nil)
+			if err != nil {
```
p-kozlowski: please add a small delay between retries, I think `<-time.After(5 * time.Second)` will do

server/server.go
```diff
@@ -398,6 +401,18 @@ func NewServer(configFile string) (*Server, error) {
 			SkipLogging: true,
 		}))
 	}
+	etcdKey := config.GetString("healthcheck/etcd_key", "/gohan")
```
p-kozlowski: `NewServer` is more than 200 lines long. Could we move this config parsing inside `NewHealthCheck`?

server/server.go
```diff
@@ -398,6 +401,18 @@ func NewServer(configFile string) (*Server, error) {
...
+	addressAndPort := strings.Split(server.address, ":")
+	if len(addressAndPort) != 2 {
+		panic("Incorrect gohan server address: " + server.address)
+	}
+	gohanPort, err = strconv.Atoi(addressAndPort[1])
+	if err != nil {
+		panic("Incorrect gohan server address: expected port number got " + addressAndPort[1])
+	}
+	healthCheckAddress := config.GetString("healtcheck/address", fmt.Sprintf("%s:%d", addressAndPort[0], gohanPort+1))
+	server.Healthcheck = healthcheck.NewHealthCheck(server.db, server.sync, etcdKey, healthCheckAddress, log)
```
p-kozlowski: Now there's no way to disable the healthcheck. It's an unauthenticated endpoint, so it's a bit serious. I think we should add a configuration knob with disabled as default.

server/server.go
michalzurawski: done, added error

healthcheck/healthcheck.go
```diff
@@ -0,0 +1,102 @@
...
+	if len(addressAndPort) != 2 {
+		return "", errors.New("Incorrect gohan server address: " + serverAddress)
+	}
+	gohanPort, err := strconv.Atoi(addressAndPort[1])
+	if err != nil {
+		return "", errors.New("Incorrect gohan server address: expected port number got " + addressAndPort[1])
+	}
+	healthCheckAddress := config.GetString(healthCheckAddressKey, fmt.Sprintf("%s:%d", addressAndPort[0], gohanPort+1))
+	if healthCheckAddress == serverAddress {
+		return "", errors.New("HealthCheck address must be different then server address " + serverAddress)
```
cierpuchaw: different *than* server address

healthcheck/healthcheck.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (healthCheck *HealthCheck) Run() {
+	healthCheckHandler := func(w http.ResponseWriter, req *http.Request) {
+		if err := healthCheck.isHealthy(); err == nil {
+			w.WriteHeader(http.StatusOK)
+		} else {
+			healthCheck.logger.Error("Health Check error: %s", err.Error())
+			w.WriteHeader(http.StatusServiceUnavailable)
+		}
+
```
cierpuchaw: Empty line between ending curly brackets.

healthcheck/healthcheck.go
```diff
@@ -0,0 +1,102 @@
...
+	DataStore db.DB
+	Etcd      sync.Sync
+	etcdKey   string
+	address   string
+	timeout   time.Duration
+	logger    log.Logger
+}
+
+func NewHealthCheck(db db.DB, etcd sync.Sync, serverAddress string, config *util.Config) *HealthCheck {
+	if enabled := config.GetBool(healthCheckEnabledKey, false); !enabled {
```
jkmar: this `enabled` seems redundant

cierpuchaw: ` if config.GetBool() {` should be enough.

server/server.go
```diff
@@ -420,6 +423,9 @@ func (server *Server) SetRunning(running bool) {
 
 //Start starts GohanAPIServer
 func (server *Server) Start() (err error) {
+	if server.HealthCheck != nil {
```
p-kozlowski: let's make `nil` meaningful (https://goplay.space/#GjkirrWI53b)

healthcheck/healthcheck.go
```diff
@@ -0,0 +1,101 @@
...
+
+func NewHealthCheck(db db.DB, etcd sync.Sync, serverAddress string, config *util.Config) *HealthCheck {
+	if !config.GetBool(healthCheckEnabledKey, false) {
+		return nil
+	}
+	healthCheckAddress, err := getHealthCheckAddress(serverAddress, config)
+	if err != nil {
+		panic(err)
+	}
+	timeout := time.Duration(config.GetInt(healthCheckTimeoutKey, defaultTimeoutInMs)) * time.Millisecond
```
p-kozlowski: actually, I won't mind if you borrow `GetDuration` from https://github.com/cloudwan/gohan/pull/715/files#diff-d891f1defb2f0e4c0d8b9494144c191aR167 ;)

michalzurawski: nice - stolen

PR #718, opened 2019-01-28
Switch back to upstream gophercloud

zimnx:


cli/client/client_test.go
```diff
@@ -211,7 +211,10 @@ var _ = Describe("CLI functions", func() {
 				os.Unsetenv("OS_USERNAME")
 				provider, err = getProviderClient()
 				Expect(provider).To(BeNil())
-				Expect(err).To(MatchError("Missing one of the following environment variables [OS_USERNAME, OS_USERID]"))
```
jkmar: How is this test related to the external library change?

zimnx: Upstream version recently merged [this change](https://github.com/gophercloud/gophercloud/commit/bb1ef8ce758cba00104a697a03ab07c77940f87e#diff-4c715a45568f9447e5182cd78f385357R89). Message is almost the same, order of missing env variables is different.

PR #721, opened 2019-02-05
Take advantage of reflectx.Mapper's internal cache.

cierpuchaw:
Building a StructMap is an expensive operation but fortunately there
is a cache for that inside reflectx's Mapper struct. Unfortunately we
were not using it because for every ResourceToMap call we created a
new mapper instance. This commit changes that to always use one global
mapper instance.

It's hard to measure exact performance improvements for ResourceToMap
run times, because they vary depending on how complex input objects
were. We might say that it is now a couple of times faster. What's
more, our project based on gohan, sees 20% runtime improvement when
running all tests, just because this one function is heavily used
there.

Removed unused Clone method while at it.

### DIFF
--- a/extension/goplugin/util.go
+++ b/extension/goplugin/util.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"sync"
 
 	"github.com/cloudwan/gohan/db/transaction"
 	"github.com/cloudwan/gohan/extension/goext"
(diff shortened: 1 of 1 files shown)


extension/goplugin/util.go
```diff
@@ -263,12 +260,20 @@ func sliceToMap(context map[string]interface{}, fieldName string, field reflect.
 	return nil
 }
 
+var (
+	jsonMapperInitOnce sync.Once
+	jsonMapper         *reflectx.Mapper
```
p-kozlowski: I think 
```Go
var (
    jsonMapper = reflectx.NewMapper("json")
)
```
will do

cierpuchaw: Done. This was too obvious for me to see ;)

PR #724, opened 2019-02-06
CW-1070 Add more metrics for auth

p-kozlowski:
We also drop support for Keystone API v2


p-kozlowski:
docs updated, commits squashed. will merge after CI approval

PR #650, opened 2018-05-24
[BL-168] Fixed sync for etcd v3

michalzurawski:
Sync fetch for etcdv3 behaves differently than in version 2 when multiple keys could be obtained. Previously only one key was returned

PR #725, opened 2019-02-08
[CW-1065][BL-355] fix sync watcher races

p-kozlowski:
Reviewers, this PR is going to be large. I'd like to ask you for a quick review of smaller functional parts already.


p-kozlowski:
Reviewers, this code is ready for a review.
One thing I noted is missing is handling of ETCD compaction. It's not directly in the scope of this PR, nevertheless I'll add it here tomorrow.

p-kozlowski:
Compaction issues resolved. Code is ready for review. Before merging, I'll give it a round of testing - I'll post results here once I'm done.

extension/goext/sync.go
```diff
@@ -47,3 +47,13 @@ type ISync interface {
 	// Update updates a path with given json
 	Update(ctx context.Context, path string, json string) error
 }
+
+type ErrCompacted struct {
```
p-kozlowski: reviewers, I've put this type in goext to avoid converting back and forth from internal types (eg. sync.ErrCompacted) to external types (eg. goext.ErrCompacted).
In fact the problem is more general, we already have duplication of some types, eg. DbOptions, (log) Level, LockPolicies etc..

Note that goext pakacge depends only on the standard library (and pkg/errors), so I started to think that goext may the best package to put definition of common interfaces!

p-kozlowski:
end-to-end testing done, works fine

extension/goext/sync.go
zimnx: Actually this is the only place where you can put any struct/interface which should be shared with Go extension. I wonder if this restriction applies since we began using Go modules.

server/path_watcher.go
```diff
@@ -0,0 +1,309 @@
...
+	sb.WriteString(watcher.path)
+	sb.WriteString(")")
+	return sb.String()
+}
+
+func (watcher *PathWatcher) Run(ctx context.Context, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	select {
+	case <-time.After(time.Duration(watcher.priority*masterTTL) * time.Second):
```
zimnx: why do we need those sleeps?

extension/goext/sync.go
cierpuchaw: >  I wonder if this restriction applies since we began using Go modules.

Even if it doesn't it's still a good thing to have a clear distinction on what's private and what's not. If we actually start allowing extensions to use anything from outside of `goext` then we will have a much harder time modifying anything here.

zimnx: I know, I'm just curious ;)

server/path_watcher.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+func (watcher *PathWatcher) consumeEvent(event *gohan_sync.Event) (err error) {
+	// each event & recovery must not be interrupted,
+	// otherwise we could easily end up in an inconsistent state
+	// if ctx is canceled either in running extensions or just before recovery
+	ctx := context.Background()
+
+	defer func() {
+		watcher.tryRecover(ctx, err, event)
```
zimnx: In case when inconsistency or compaction error won't be resolved consumeEvents() will iterate over all other events, and probably all of them will fail in same way. We can save few iterations by returning error here if recovery failed.
 What do you think?

sync/etcdv3/etcd_test.go
```diff
@@ -277,88 +233,140 @@ func TestWatchWithRevision(t *testing.T) {
...
+
+	resp := <-responseChan
+	if resp.Key != path || resp.Data["version"].(string) != "1" || resp.Revision != firstRevision {
+		t.Fatalf("mismatch response: %+v, expecting version==1, revision==%d", resp, firstRevision)
 	}
-	if nodes.Children[1].Key != "/path/to" {
-		t.Fatalf("expected node to has key /path/to has %s instead", nodes.Children[1].Key)
+
+	resp = <-responseChan
+	if resp.Key != path || resp.Data["version"].(string) != "2" || resp.Revision != secondRevision {
```
cierpuchaw: We seem to always check for key, version and revision. Can we extract this pattern into a function? `verifyResponse(resp, path, "2", secondRevision)`?

sync/etcdv3/etcd_test.go
```diff
@@ -277,88 +233,140 @@ func TestWatchWithRevision(t *testing.T) {
...
 	if resp.Key != path+"/third" || resp.Data["existing"].(bool) != false || resp.Revision != thirdRevision {
 		t.Fatalf("mismatch response: %+v, expecting /third, existing==false, revision==%d", resp, thirdRevision)
 	}
-
 }
 
-func TestFetchMultipleNodes(t *testing.T) {
-	ctx := context.Background()
+func TestShouldReturnAllValuesWhenWatchingAtCurrentRevision(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
zimnx: consider context.WithTimeout(), if sync would hang test will be interrupted after 10 minutes.

server/path_watcher.go
```diff
@@ -0,0 +1,309 @@
...
+		return
+	}
+
+	for {
+		err := watcher.run(ctx)
+
+		switch err {
+		case errLockFailed:
+			watcher.updateCounter(1, "lock.failed")
+			log.Debug("%s failed to acquire lock, retrying...", watcher)
```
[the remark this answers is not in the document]
p-kozlowski: it's because `log` refers to a package-scope [variable](https://github.com/cloudwan/gohan/blob/CW-1065-fix-sync-watcher-races/server/log.go#L22), not a package name

server/path_watcher.go
```diff
@@ -0,0 +1,309 @@
...
+		case context.Canceled:
+			// Do nothing, normal shutdown
+		default:
+			watcher.updateCounter(1, "error")
+			log.Error("%s aborted, retrying...: %s", watcher, err)
+		}
+
+		select {
+		case <-time.After(time.Duration(watcher.priority*masterTTL+1) * time.Second):
```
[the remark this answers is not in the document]
p-kozlowski: my _guess_ is that in a sunny day scenario:
1. the prefered node gets priority 0
2. we do not want to delay startup, so the node with priority 0 does not sleep at the beginning, instead it starts competing for the lock immediately
3. if we got here, there has (probably) been an error in `run()` and we do not restart immediately, instead we wait at least 1 second

server/path_watcher.go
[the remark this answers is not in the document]
cierpuchaw: Found it. Package `server` defines `var log = l.NewLogger()` and `path_watcher.go` happens to belong to that package. It seems we have 2 methods of logging in this file...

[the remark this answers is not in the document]
cierpuchaw: Scratch all that, `l.Panic(log)` is actually panic-recovery, name not misleading at all.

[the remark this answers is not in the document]
zimnx: I had to lookup it too during first glance. Can you change name to `LogPanic(log)`? Also 'l' is declared too far from scope where its used.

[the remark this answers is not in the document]
cierpuchaw: It's kind of funny that `Panic(log)` does something entirely different than `log.Panic()` ;)

server/path_watcher.go
```diff
@@ -0,0 +1,309 @@
...
+			err = ctx.Err()
+			return
+		}
+	}
+}
+
+func (watcher *PathWatcher) consumeEvent(event *gohan_sync.Event) (err error) {
+	// each event & recovery must not be interrupted,
+	// otherwise we could easily end up in an inconsistent state
+	// if ctx is canceled either in running extensions or just before recovery
```
cierpuchaw: On the other hand, if one of the functions called here never returns, we will never exit from `func (watcher *PathWatcher) run(` as it would wait forever on `doneCh`

server/path_watcher.go
```diff
@@ -0,0 +1,309 @@
...
+		watcher.updateCounter(1, "inconsistency_recovery.fetch_errors")
+		log.Critical("%s can't recover from inconsistent cluster state on %s: Fetch() failed: %s", watcher, event.Key, err)
+		return
+	}
+
+	if recovered, err := watcher.sync.CompareAndSwap(ctx, event.Key, node.Value, watcher.sync.ByValue(node.Value)); err != nil {
+		// some sync events could have be processed out-of-order and the recovery failed.
+		// incorrect data could be stored in DB. a user has to recover (resync) manually
+		watcher.updateCounter(1, "inconsistency_recovery.cas_errors")
+		log.Critical("%s can't recover from inconsistent cluster state on %s: notifying the current master failed: %s", watcher, event.Key, err)
```
cierpuchaw: Lots of very long lines until the end of this file.

server/path_watcher.go
```diff
@@ -0,0 +1,309 @@
...
+
+	return watcher.storeRevision(ctx, event.Revision)
+}
+
+func (watcher *PathWatcher) tryRecover(ctx context.Context, err error, event *gohan_sync.Event) {
+	if err == errInconsistentCluster {
+		watcher.tryRecoverInconsistentCluster(ctx, event)
+	} else if errCompacted, ok := err.(goext.ErrCompacted); ok {
+		watcher.tryRecoverCompaction(ctx, event, errCompacted.CompactRevision)
+	}
```
cierpuchaw: else log about unsupported recovery try?

server/path_watcher.go
```diff
@@ -0,0 +1,309 @@
...
+		"action":   response.Action,
+		"data":     response.Data,
+		"key":      response.Key,
+		"context":  ctx,
+		"trace_id": util.NewTraceID(),
+	}
+	if err := env.HandleEvent("notification", context); err != nil {
+		log.Error("%s extension error, last processed event may be lost: %s", watcher, err)
+	}
+	return
```
cierpuchaw: Redundant `return`

server/path_watcher.go
```diff
@@ -0,0 +1,309 @@
...
+	return fromRevision
+}
+
+// storeRevision puts a revision number for a path to the sync backend.
+func (watcher *PathWatcher) storeRevision(ctx context.Context, revision int64) error {
+	path := SyncWatchRevisionPrefix + watcher.path
+	value := strconv.FormatInt(revision, 10)
+	opts := make([]gohan_sync.CASCondition, 0, 1)
+	if watcher.previousProcessedRevision != int64(gohan_sync.RevisionCurrent) {
+		opts = append(opts, watcher.sync.ByValue(strconv.FormatInt(watcher.previousProcessedRevision, 10)))
```
cierpuchaw: Shouldn't this be `ByRevision`?

cierpuchaw: `ByRevision` isn't used anywhere outside of tests.

server/path_watcher.go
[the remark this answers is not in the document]
p-kozlowski: we can do better: the [util.go](https://github.com/cloudwan/gohan/blob/master/log/util.go) had two functions:

1. Panic() which is incorrectly named and is called only here, and
2. FatalPanic() which is called just once (in cron.go), but it's call both redundantly & incorrectly (it's actually a dead code)

let me sort it out for good.

sync/etcdv3/etcd_test.go
```diff
@@ -30,79 +32,46 @@ func TestNewSyncTimeout(t *testing.T) {
 func TestNonEmptyUpdate(t *testing.T) {
```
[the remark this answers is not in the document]
p-kozlowski: fixed

sync/etcdv3/etcd_test.go
```diff
@@ -121,19 +90,15 @@ func TestRecursiveUpdate(t *testing.T) {
 	checkNode(abed.Children[1], "/a/b/e/d/2", "test5", 0, t)
 	checkNode(abed.Children[2], "/a/b/e/d/3", "test6", 0, t)
 
-	node, err = syn.Fetch(ctx, "/a/b/c")
-	if err != nil {
-		t.Fatalf("unexpected error %s", err.Error())
-	}
-	checkNode(abc, "/a/b/c", "", 1, t)
+	sync.mustExist("/a/b/c", "", 1)
```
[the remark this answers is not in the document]
p-kozlowski: I think it actually guards against a regression, cf. #650

sync/etcdv3/etcd_test.go
p-kozlowski: sure, changed

sync/etcdv3/etcd_test.go
p-kozlowski: good point, changed

server/path_watcher.go
p-kozlowski: removed

server/path_watcher.go
p-kozlowski: as discussed offline: there's no easy way out: in a hypothetical application build on Gohan framework, executing extensions may take an arbitrary amount of time.

let me extend the comment to mention this issue

server/path_watcher.go
p-kozlowski: formatting of logs improved

server/path_watcher.go
p-kozlowski: sure thing

server/path_watcher.go
p-kozlowski: so let's just remove it

p-kozlowski:
@zimnx @cierpuchaw I've addressed your comment, please re-review.

PR #726, opened 2019-02-12
[CW-1077] Response ownership and allowing Cookie as a Auth method

zimnx:


morrisson:
Tests need to be fixed, but those changes look good.

PR #729, opened 2019-02-28
[CW-1186] Copy data during update

michalzurawski:
Previously the data was not copied which resulted in incorrect data in
the cache.

### DIFF
--- a/extension/goplugin/integration_tests/end_to_end_test.go
+++ b/extension/goplugin/integration_tests/end_to_end_test.go
@@ -218,6 +218,7 @@ var _ = Describe("Environment", func() {
 			Expect(result).To(HaveKeyWithValue("error", "Validation error: Json validation error:\n\tname: String length must be greater than or equal to 3,"))
 			resource["name"] = "abcd"
 			expectedResponse["name"] = "abcd"
+			delete(expectedResponse, "subobject")
 			result = testURL("PUT", baseURL+"/v0.1/tests/testId", adminTokenID, resource, http.StatusOK)
 			Expect(result).To(HaveKeyWithValue("test", expectedResponse))
 		})
--- a/schema/resource.go
+++ b/schema/resource.go
@@ -117,7 +117,7 @@ func (resource *Resource) String() string {
 }
 
 //Update resource data
-func (resource *Resource) Update(updateData map[string]interface{}) map[string]interface{} {
+func (resource *Resource) CloneWithUpdate(updateData map[string]interface{}) map[string]interface{} {
 	data := make(map[string]interface{}, len(resource.properties))
 	for key, value := range resource.properties {
 		data[key] = value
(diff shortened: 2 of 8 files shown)


server/resources/resource_management_test.go
```diff
@@ -1624,6 +1624,30 @@ var _ = Describe("Resource manager", func() {
 
 			})
 		})
+		Describe("Whether old data is not overridden in update_in_transaction", func() {
+			BeforeEach(func() {
+				javascriptCode := fmt.Sprintf(`
+					tx = context.transaction;
+					res = gohan_db_fetch(tx, "%s", "%s", "");
```
michalzurawski: Long live JS!
Sorry guys for that but that was easiest way for reproducing this issue in UT ;)

p-kozlowski: I'd dispute what I'd wish for the JS exts, but definitely it'll not be a "long live" ;)

You could try to `RegisterEnvironment` with a `MockEnvironment` with `DoAndReturn(func() { yourGoCode() }` registered on `HandleEvent`

michalzurawski:
@jkmar @zimnx @cierpuchaw @morrisson @tomasz-michalski @kuba2 kind reminder about this PR

schema/resource.go
```diff
@@ -117,16 +117,12 @@ func (resource *Resource) String() string {
...
 //Update resource data
-func (resource *Resource) Update(updateData map[string]interface{}) error {
-	data := resource.properties
-	for _, property := range resource.schema.Properties {
-		id := property.ID
-
-		if val, ok := updateData[id]; ok {
-			data[id] = updateResourceRecursion(val, data[id])
-		}
+func (resource *Resource) Update(updateData map[string]interface{}) map[string]interface{} {
```
cierpuchaw: This no longer updates the resource so maybe a new name is in order. Something like `MakeUpdatedCopy`, `CloneWithUpdate` etc.

extension/goplugin/integration_tests/end_to_end_test.go
```diff
@@ -218,6 +218,7 @@ var _ = Describe("Environment", func() {
 			Expect(result).To(HaveKeyWithValue("error", "Validation error: Json validation error:\n\tname: String length must be greater than or equal to 3,"))
 			resource["name"] = "abcd"
 			expectedResponse["name"] = "abcd"
+			delete(expectedResponse, "subobject")
```
cierpuchaw: Why this is not returned by PUT? We know from BeforeEach that resource has this field and I don't see anything being removed.

michalzurawski: actually, this was something which also confused me and i thought it was a bug but it seems is expected behavior. We do not return nested objects which are `nil` only when `*_in_transaction` handler is defined (previously there was only `pre_create` handler that's why `subobject` was returned)

cierpuchaw: Ack. Seems inconsistent and problematic but this is not a problem with this PR, approving.

PR #730, opened 2019-03-04
[BL-355] Do not send Get() request when starting revision is specified

p-kozlowski:
ETCD's API does allow Get() with a selected revision, but it's impossible to express "Get() with a revision equal or greater". When a Watch() is requested starting an given revision, we just omit the call to Get().

Previously the API will return the current revision of the given key, not the one the API user requested. Ie. the newly added test will fail with ```mismatch response: &{Action:get Key:/path/to/watch/with/starting/revision Data:map[version:2] Revision:56501 Err:<nil>}, expecting version==1, revision==56500``` (as seen in the [CI](https://circleci.com/gh/cloudwan/gohan/2613?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link))

### DIFF
--- a/sync/etcdv3/etcd.go
+++ b/sync/etcdv3/etcd.go
@@ -358,28 +358,34 @@ func eventsFromNode(ctx context.Context, action string, kvs []*pb.KeyValue, resp
 	}
 }
 
-//Watch keep watch update under the path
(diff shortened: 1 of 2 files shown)


p-kozlowski:
let me decline this PR and merge it with #725. the reason is that removing Get() will cause issues with etcd compaction. keeping the 2 PRs will require doing the same work twice as the affected code has been changed significantly in #725.

PR #733, opened 2019-03-08
Follow SELECT with SELECT COUNT only when required.

cierpuchaw:
After every SELECT we called SELECT COUNT to return total number of
entries in requested table. This can be different than number of
entries returned from SELECT only if we use OFFSET or LIMIT from
paginator.

### DIFF
--- a/db/sql/sql.go
+++ b/db/sql/sql.go
@@ -871,12 +871,16 @@ func (tx *Transaction) executeSelect(ctx context.Context, sc *selectContext, sql
 		return nil, 0, err
 	}
 
-	if sc.paginator != nil && (sc.paginator.Offset != 0 || sc.paginator.Limit != math.MaxUint64) {
+	if tx.isSelectPaginated(sc) {
 		total, err = tx.Count(ctx, sc.schema, sc.filter)
 	} else {
 		total = uint64(len(list))
 	}
 	return
+}
+
+func (tx *Transaction) isSelectPaginated(sc *selectContext) bool {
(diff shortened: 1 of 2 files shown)


db/sql/sql.go
```diff
@@ -870,7 +870,12 @@ func (tx *Transaction) executeSelect(ctx context.Context, sc *selectContext, sql
 	if err != nil {
 		return nil, 0, err
 	}
-	total, err = tx.Count(ctx, sc.schema, sc.filter)
+
+	if sc.paginator != nil && (sc.paginator.Offset != 0 || sc.paginator.Limit != 0) {
```
tomasz-michalski: Why `sc.paginator.Limit != 0` and not `sc.paginator.Limit != math.MaxUint64`? Default value for `paginator.Limit` is `math.MaxUint64`

Please make test with
`pagination.NewPaginator(pagination.OptionLimit(0))`

p-kozlowski: how about
```go
func (sc selectContext) hasPagination() bool
```
?

cierpuchaw: No idea how I've missed that, THANK YOU!

Added a test for the "LIMIT 0" case.

cierpuchaw: @p-kozlowski went with `func (tx *Transaction) isSelectPaginated`, because it's used only in `func (tx *Transaction) executeSelect` and `selectContext` has no methods at all, it's always used as a parameter.

PR #734, opened 2019-03-13
[BL-339]Search api

giyyanan:


db/sql/sql.go
```diff
@@ -1215,6 +1217,12 @@ func AddFilterToQuery(s *schema.Schema, q sq.SelectBuilder, filter map[string]in
 			column = quote(key)
 		}
 
+		substr,ok := value.(string)
+		if ok && strings.HasPrefix(substr,"~"){
+		 	q = q.Where(sq.Like{column: buildLikeValues(substr[1:])})
+		 	continue
+		}
```
morrisson: I got your idea but see my comment below.

server/resources/resource_management.go
```diff
@@ -316,6 +316,23 @@ func GetMultipleResources(context middleware.Context, dataStore db.DB, resourceS
 	currCond := policy.GetCurrentResourceCondition()
 
 	filter := FilterFromQueryParameter(resourceSchema, queryParameters)
+	if search_filter,ok := queryParameters["search_field"]; ok && resourceSchema.IsSubstringSeachEnabled(){
+			for _,column := range search_filter{
+				if search_value,ok := queryParameters[column]; ok {
+					delete(filter, column)
+					searchValue := strings.Builder{}
+					searchValue.WriteString("~")
```
morrisson: I think tweaking original input with magic prefix tilde is not good idea. It makes hard for other developers to understand how it works, and may occur some unexpected errors (e.g., users may use tilde as name field, description field and so on...). 

We should manage what properties are used for substr match in other fields. Actually, I like @michalzurawski 's idea, which is discussed offline, that enhance `paginator` below.

server/resources/resource_management.go
```diff
@@ -316,6 +316,23 @@ func GetMultipleResources(context middleware.Context, dataStore db.DB, resourceS
...
+					filter[column] = searchValue.String()
+				} else{
+					err := fmt.Errorf("search value for `%s` not available in URL ", column)
+					return ResourceError{err, err.Error(), WrongQuery}
+				}
+			}
+			log.Info("substring search is enabled")
+	}
+
+	extendFilterByTenantAndDomain(resourceSchema, filter, schema.ActionRead, currCond, auth)
```
morrisson: `extendFilterByTenantAndDomain` is doubled.

db/sql/sql_test.go
```diff
@@ -411,6 +411,29 @@ var _ = Describe("Sql", func() {
...
+			})
+			It("should create like query with two parameter", func() {
+				filter := map[string]interface{}{"test_string": "~123", "test_number": "~42"}
+
+				res, err := AddFilterToQuery(testSchema, query, filter, false)
+
+				Expect(err).ToNot(HaveOccurred())
+				resSql, param, err := res.ToSql()
+				Expect(err).ToNot(HaveOccurred())
+				checkAndStatement(resSql, "`test_string` LIKE ?", "`test_number` LIKE ?", param, []interface{}{"%42%", "%123%"})
```
cierpuchaw: `checkAndStatement` does not check order, only presence. But please be consistent and reverse what you pass as `expectedParams`.

server/resources/resource_management.go
cierpuchaw: +1 about `~` being problematic. Can you guys paste Michal's idea here?

server/server_test.go
```diff
@@ -1273,6 +1273,39 @@ var _ = Describe("Server package test", func() {
...
+		})
+
+		AfterEach(func() {
+			testURL("DELETE", getNetworkSingularURL("red"), adminTokenID, nil, http.StatusNoContent)
+			testURL("DELETE", getNetworkSingularURL("red1"), adminTokenID, nil, http.StatusNoContent)
+			testURL("DELETE", getNetworkSingularURL("red2"), adminTokenID, nil, http.StatusNoContent)
+			testURL("DELETE", getNetworkSingularURL("red3"), adminTokenID, nil, http.StatusNoContent)
+		})
+
+		DescribeTable("match all rows by substring search", func(query string) {
```
cierpuchaw: If you move `func(query string...` to next line, goimports/gofmt will format it a bit better. This is of course subjective.

server/server_test.go
```diff
@@ -1273,6 +1273,39 @@ var _ = Describe("Server package test", func() {
...
+		},
+			Entry("empty query", ""),
+			Entry("long prefix", "?id=networkred&search_field=id"),
+			Entry("substring at the start", "?id=net&search_field=id"),
+			Entry("substring at the middle", "?id=work&search_field=id"),
+			Entry("substring at the last", "?id=red&search_field=id"),
+			Entry("search on one parameter", "?description=red&search_field=description"),
+			Entry("more than one field to search","?id=red&search_field=id&description=red&search_field=description"),
+		)
+	})
```
cierpuchaw: Can you also add some tests to show that we perform exact matches by default?

cierpuchaw: `search_field=` does not really explain what it does.

server/resources/resource_management.go
```diff
@@ -316,6 +316,23 @@ func GetMultipleResources(context middleware.Context, dataStore db.DB, resourceS
 	currCond := policy.GetCurrentResourceCondition()
 
 	filter := FilterFromQueryParameter(resourceSchema, queryParameters)
+	if search_filter,ok := queryParameters["search_field"]; ok && resourceSchema.IsSubstringSeachEnabled(){
```
cierpuchaw: If we can enable substring searching on demand now, do we really need to have that also enabled in schemas?

server/resources/resource_management.go
cierpuchaw: Yep, this is already called 2 lines below.

schema/schema.go
```diff
@@ -278,6 +278,19 @@ func (schema *Schema) Init() error {
 	return nil
 }
 
+//Is substring search enabled for the schema
+func (schema *Schema) IsSubstringSeachEnabled() bool {
+	likeEnabled, ok := schema.Metadata["substr_search_enabled"]
+	if !ok {
+		return true
```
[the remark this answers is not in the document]
giyyanan: this one specifically for blocking sub string search on certain schema's like log tables

server/resources/resource_management.go
giyyanan: earlier this flag was disable_substring_search...earlier reviews suggested no to stick to negative logic ..so changed it to enable_substring_search..probably the enable might be confusing..will try to find a different name
https://github.com/cloudwan/gohan/pull/731#discussion_r262837149

db/sql/sql_test.go
```diff
@@ -411,6 +412,32 @@ var _ = Describe("Sql", func() {
...
+				number_search := search.NewSearchField("42")
+				filter := map[string]interface{}{"test_string": string_search, "test_number": number_search}
+
+				res, err := AddFilterToQuery(testSchema, query, filter, false)
+
+				Expect(err).ToNot(HaveOccurred())
+				resSql, param, err := res.ToSql()
+				Expect(err).ToNot(HaveOccurred())
+				checkAndStatement(resSql, "`test_string` LIKE ?", "`test_number` LIKE ?", param, []interface{}{ "%123%","%42%"})
+			})
```
morrisson: Let's add an unit test of SQL statement mixed with substr search enabled/disabled fields.

docs/schema.md
```diff
@@ -654,6 +659,23 @@ HTTP Status Code: 200
     }
   }
 ```
+GET http://$GOHAN/[$namespace_prefix/]$prefix/$plural/$id&search_field=$id
```
morrisson: `/$id` is an example of GETting single resource, not multiple resources, so it's not a good place to explain substr match.

morrisson:
@giyyanan  `go fmt` will help you to make your code well-formatted. I believe some plugins of auto `go fmt` should be available for your IDE.

server/server_test.go
giyyanan: @cierpuchaw  exact matches would be the get default get queries ...wouldn't existing test case suffice for that... 

search_field..i could change it to search_on_field or search_column

db/search/search.go
```diff
@@ -0,0 +1,36 @@
...
+const (
+	LIKE = "like"
+)
+
+type Search struct {
+	Value string
+}
+
+func NewSearchField(value string) Search {
+	searchValue := strings.Builder{}
```
p-kozlowski: this comment https://github.com/cloudwan/gohan/pull/731#discussion_r263075188 is still valid and needs to be resolved

db/sql/sql.go
```diff
@@ -1274,6 +1282,12 @@ func addToFilter(s *schema.Schema, q sq.SelectBuilder, filter interface{}, join
 
 			// TODO: add other operators
 			value := filter["value"]
+			substr, ok := value.(search.Search)
+			if ok {
+				sqlizer = append(sqlizer, sq.Like{column: substr.Value})
```
p-kozlowski: there are not tests for it (and it looks like it breaks the API). please see how the "extended" API looks like and add tests for it [this](https://github.com/cloudwan/gohan/blob/master/db/sql/sql_test.go#L425) block

server/server_test.go
```diff
@@ -1273,6 +1273,75 @@ var _ = Describe("Server package test", func() {
...
+
+		AfterEach(func() {
+			testURL("DELETE", getNetworkSingularURL("red"), adminTokenID, nil, http.StatusNoContent)
+			testURL("DELETE", getNetworkSingularURL("red1"), adminTokenID, nil, http.StatusNoContent)
+			testURL("DELETE", getNetworkSingularURL("red2"), adminTokenID, nil, http.StatusNoContent)
+			testURL("DELETE", getNetworkSingularURL("red3"), adminTokenID, nil, http.StatusNoContent)
+		})
+
+		DescribeTable("match all rows by substring search", func(query string) {
+			result := testURL("GET", networkPluralURL+query, adminTokenID, nil, http.StatusOK)
```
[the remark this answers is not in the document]
giyyanan: path seems to append a escape slash to an already escaped one..

networkURL is stored as string `http://localhost:19090` and path overtires it with `http:///localhost:19090` which causes test cases to fail...could use string builder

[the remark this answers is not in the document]
p-kozlowski: ah, I see. StringBuilder would be an overkill. let's keep it as-is

db/search/search.go
giyyanan: In case we are escaping the characters '%','_','/' with slash or something, we need to make sure that when the query is executed the escape character is handled in the queries 
i.e for instance with maria DB
https://mariadb.com/kb/en/library/like/ you specify the escape character while executing like queries

otherwise `%`+'/'+`%` can be executed as `%`+'//'+`%` and might result in incorrect results if not handled.

Since we are using Masterminds/squirrel package i don't see a place where escape characters are handled while framing the query strings

https://github.com/Masterminds/squirrel/blob/55303df43ec0bf2b0ce3f9c3d8c3ce07022ef76b/expr.go#L169

zimnx: By default, escape character is set to backslash. 'squirrel' doesn't have to support changing it, we can go with backslash.
https://mariadb.com/kb/en/library/like/
> If you need to match the characters _ or %, you must escape them. By default, you can prefix the wildcard characters the backslash character \ to escape them. The backslash is used both to encode special characters like newlines when a string is parsed as well as to escape wildcards in a pattern after parsing. Thus, to match an actual backslash, you sometimes need to double-escape it as "\\\\".

giyyanan: yes...mariadb supports backslash...gohan is DB agnostic right..isn't that the reason why squirrel is used ....don't we need to handle for other databases as well.. I tested with sqlite and it yielded wrong results, when escaped with backlash..

zimnx: Well, in this case you can always implement you own Where matcher similar to existing sq.Like, which supports ESCAPE clause. All you have to do is to implement sq.Sqlizer interface on your custom struct.

p-kozlowski: implementing sq.Sqlizer sound a bit like an overkill for me. maybe I'm missing something, but can we implement [driver.Valuer](https://github.com/Masterminds/squirrel/blob/v1/expr.go#L183) instead and stay with regular sq.Like? eg.

```Go
func (s *Search) Value() (driver.Value, error) {
	buf := strings.Builder{}
	buf.WriteString("%")
	buf.WriteString(escape(s.value))
	buf.WriteString("%")
	return buf.String()
}
```

db/sql/sql.go
```diff
@@ -1215,6 +1217,12 @@ func AddFilterToQuery(s *schema.Schema, q sq.SelectBuilder, filter map[string]in
 			column = quote(key)
 		}
 
+		substr, ok := value.(search.Search)
+		if ok {
+			q = q.Where(sq.Like{column: substr.Value})
```
p-kozlowski: if we go with implementing `driver.Valuer`, this will become simply
```Go
q = q.Where(sq.Like{column: substr})
```

examples/noauth/gohan.yaml
```diff
@@ -6,10 +6,10 @@
 database:
     # yaml, json, sqlite3 and mysql supported
     # yaml and json db is for schema development purpose
-    type: "sqlite3"
+    type: "mysql"
     # connection string
     # it is file path for yaml, json and sqlite3 backend
-    connection: "./gohan.db"
+    connection: "chilli:passwd@tcp(127.0.0.1:3306)/gohan_test"
```
p-kozlowski: private testing data?

schema/schema.go
```diff
@@ -278,6 +278,15 @@ func (schema *Schema) Init() error {
 	return nil
 }
 
+//Is substring search enabled for the schema
+func (schema *Schema) IsSubstringSearchEnabled() bool {
+
```
p-kozlowski: redundant whitespace

server/server_test.go
```diff
@@ -1273,6 +1273,76 @@ var _ = Describe("Server package test", func() {
...
+		DescribeTable("match and search combinations return two rows", func(query string) {
+			result := testURL("GET", networkPluralURL+query, adminTokenID, nil, http.StatusOK)
+			Expect(result).To(HaveKeyWithValue("networks", ConsistOf(
+				util.MatchAsJSON(getNetwork("red3", "blue")),
+				util.MatchAsJSON(getNetwork("red2", "blue")))))
+		},
+			Entry("match on id and search on description", "?id=networkred2&id=networkred3&description=red&search_field=description"),
+			Entry("match on tenant and search on id", "?id=red&tenant_id=blue&search_field=id"),
+			Entry("search on tenant", "?tenant_id=blue&search_field=tenant_id"),
+			Entry("non exclusive parameters","?id=re&search_field=id&tenant_id=bl&search_field=tenant_id"),
```
p-kozlowski: please run `goimports` (which is a superset of `go fmt`)

schema/schema.go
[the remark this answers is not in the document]
cierpuchaw: I don't get it. Why is substring searching **enabled** by default when schema does not enable or disable it explicitly? What's blocked here?

This function has had its contents changed completely but this property was retained.

examples/noauth/gohan.yaml
giyyanan: my bad..removed

schema/schema.go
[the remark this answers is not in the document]
giyyanan: the goal was to enable sub string search on all schema tables except for a few where the no of records run into millions i.e log tables..in such cases  we explicit block the search feature even if the API requests for it with the `search_field` parameter....

the naming is confusing but was changed as per previous comments...i would have preferred 'disable_substr_search'

db/sql/sql.go
giyyanan: except for injecting the squirrel.Like in some of the already existing test cases couldn't find any need for new test cases...

db/search/search.go
giyyanan: apart from this suggested change..i escaped characters and ran test with the server_test.go..still it fails in producing the correct results as i believe the test configuration is not mysql..so tests for the search feature fails with escaped chars..which brings to ..why aren't we testing the api's on a set of databases..rather than just sqlite3

zimnx: @p-kozlowski As far as I understand, sqlite doesn't support backslash escape character by default. To enable it, query must contain optional ESCAPE statement followed by escape character. In MariaDB, backslash is enabled by default. squirrel doesn't provide such query modifier, but exposes interace which could be implemented to add such field. 

@giyyanan i see that characters are escaped, and you wrote that some tests are failing. Those on CircleCI are passing. Does CI skips any tests?

giyyanan: I haven't checked in the code for unit test cases with special characters..

schema/schema.go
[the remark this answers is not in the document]
cierpuchaw: Got it!

server/server_test.go
```diff
@@ -1273,6 +1273,139 @@ var _ = Describe("Server package test", func() {
...
+				util.MatchAsJSON(getNetwork("_apple", "apple")),
+				util.MatchAsJSON(getNetwork("\\ball", "ball")),
+				util.MatchAsJSON(getNetwork("%cats", "cats")),
+				util.MatchAsJSON(getNetwork("a\\x%y_z", "xyz")))))
+		},
+			Entry("empty query", ""),
+			Entry("common occuring character", "?id=a&search_field=id"),
+		)
+
+		PDescribeTable("search should return single row with _", func(query string) {
```
zimnx: Instead marking tests as Pending, it's better to implement [my previous comment](https://github.com/cloudwan/gohan/pull/734/files#r271553551) to support character escaping in both supported DB engines.

morrisson:
@zimnx @p-kozlowski @cierpuchaw @michalzurawski @jkmar Please make a review again. Thanks.

db/sql/sql.go
```diff
@@ -1299,3 +1315,48 @@ func (db *DB) SetMaxOpenConns(maxIdleConns int) {
...
+		case driver.Valuer:
+			if val, err = v.Value(); err != nil {
+				return
+			}
+		}
+
+		if val == nil {
+			err = fmt.Errorf("cannot use null with like operators")
+			return
+		} else {
```
cierpuchaw: This `else {` is not needed, we always return from the function when taking the first branch.

db/sql/sql.go
```diff
@@ -1299,3 +1315,48 @@ func (db *DB) SetMaxOpenConns(maxIdleConns int) {
...
+			if isListType(val) {
+				err = fmt.Errorf("cannot use array or slice with like operators")
+				return
+			} else {
```
cierpuchaw: Same here.

db/sql/sql.go
```diff
@@ -1299,3 +1315,48 @@ func (db *DB) SetMaxOpenConns(maxIdleConns int) {
 func safeMysqlContext(_ context.Context) context.Context {
 	return context.Background()
 }
+
+type Like map[string]interface{}
+
+func (lk Like) ToSql() (sql string, args []interface{}, err error) {
+	var (
+		exprs []string
```
cierpuchaw: `exprs` is either empty (in case of error) or has `len(lk)` elements, we can preallocate memory here.

db/sql/sql_test.go
```diff
@@ -411,6 +412,57 @@ var _ = Describe("Sql", func() {
...
+
+				res, err := AddFilterToQuery(testSchema, query, filter, false)
+
+				Expect(err).ToNot(HaveOccurred())
+				resSql, param, err := res.ToSql()
+				Expect(err).ToNot(HaveOccurred())
+				expectedQuery = expectedQuery.Where(Like{"`test_string`": "%\\\\1\\%2\\_3\\_a\\\\b\\%c%"})
+				expectedSql, expectedParam, _ := expectedQuery.ToSql()
+				Expect(resSql).To(Equal(expectedSql))
+				Expect(param).To(Equal(expectedParam))
```
cierpuchaw: Above block of code is virtually identical to the one in the preceding test case. You can factor this out into a function.

db/sql/sql_test.go
```diff
@@ -448,6 +501,31 @@ var _ = Describe("Sql", func() {
...
+				res, err := AddFilterToQuery(testSchema, query, filter, false)
+
+				Expect(err).ToNot(HaveOccurred())
+				resSql, param, err := res.ToSql()
+				Expect(err).ToNot(HaveOccurred())
+
+				expectedQuery = expectedQuery.Where(squirrel.Or{Like{"`test_string`": "%123%"}, Like{"`test_number`": "%42%"}})
+				expectedSql, expectedParam, _ := expectedQuery.ToSql()
+				Expect(resSql).To(Equal(expectedSql))
+				Expect(param).To(Equal(expectedParam))
```
cierpuchaw: Another candidate for using a helper function.

db/sql/sql_test.go
```diff
@@ -473,6 +551,33 @@ var _ = Describe("Sql", func() {
...
+				res, err := AddFilterToQuery(testSchema, query, filter, false)
+
+				Expect(err).ToNot(HaveOccurred())
+				resSql, param, err := res.ToSql()
+				Expect(err).ToNot(HaveOccurred())
+
+				expectedQuery = expectedQuery.Where(squirrel.And{Like{"`test_string`": "%123%"}, Like{"`test_number`": "%42%"}})
+				expectedSql, expectedParam, _ := expectedQuery.ToSql()
+				Expect(resSql).To(Equal(expectedSql))
+				Expect(param).To(Equal(expectedParam))
```
cierpuchaw: Another candidate for using a helper function.

docs/schema.md
```diff
@@ -145,6 +145,10 @@ Developers can make a schema as abstract schema specifying type=abstract. The de
 
   We don't sync this resource for sync backend when this option is true.
 
+- enable_substring_search (boolean)
+
+  can be used to block substring matches while fetching data in GET API request calls. By defult substring search i enabled for all schemas
```
cierpuchaw: Tiny typo: `search i enabled` -> `search is enabled`

db/sql/sql.go
giyyanan: this just a copy of the function in squirrel..had to copy it here just for the ESCAPE part..intent of the function was to use as exactly as what squirrel library has and just inject ESCAPE for the queries..

michalzurawski: i guess Wojciech meant to change this line to `exprs = make([]string, 0, len(lk))` to avoid unnecessary allocation in for loop

db/sql/sql.go
giyyanan: else where the ESCAPE injunction is

db/sql/sql.go
giyyanan: else where the ESCAPE injunction is

db/sql/sql.go
cierpuchaw: I'm not asking to remove the entire `else` block. Let me explain using an example.

We have a similar situation to the following:
```go
if cond1 {
  return
} else {
  if cond2 {
    return
  } else {
    doStuff()
  }
}
```

This is equivalent to:
```go
if cond1 {
  return
}

if cond2 {
  return
}

doStuff()
```

Doing this transformation reduces unneeded indentation, which with deep nesting, makes it hard to follow the code.

db/sql/sql_test.go
```diff
@@ -422,6 +468,35 @@ var _ = Describe("Sql", func() {
 				checkAndStatement(resSql, "`test_string` = ?", "`test_number` = ?", param, []interface{}{42, "123"})
 			})
 		})
+		checkLikeDisjunctionAndConjunction := func(searchString string, numberString string, filterType string, convertedString string, convertedNumber string){
```
morrisson: I think we can consider one step more for generalized helper function and readable test cases not only for your added test cases. Does following function definition work for you?

```
checkTranslationFilterToWhereClause := func(filter map[string]interface{}, whereClause squirrel.Sqlizer) {
	res, err := AddFilterToQuery(testSchema, query, filter, false)

	resSql, param, err := res.ToSql()

	expectedQuery = expectedQuery.Where(whereClause)
	expectedSql, expectedParam, _ := expectedQuery.ToSql()
	Expect(resSql).To(Equal(expectedSql))
	Expect(param).To(Equal(expectedParam))
}
```

Maybe above helper function can be used by not only "Disjunction and conjunction" context but also "Basic select query".

db/sql/sql_test.go
```diff
@@ -448,6 +523,9 @@ var _ = Describe("Sql", func() {
 				Expect(resSql).To(Equal(expectedSql))
 				Expect(param).To(Equal(expectedParam))
 			})
+			It("should create disjunction for more than one parameter with like queries", func() {
+				checkLikeDisjunctionAndConjunction("123","42","__or__","%123%","%42%")
```
morrisson: `go fmt` should make a space after each comma.

cierpuchaw: Tip: set an after-save-hook in your editor to run `gofmt`, or even better, `goimports`.

giyyanan: thanks @cierpuchaw

morrisson:
Unfortunately, following files are changed unintentionally while you squash commits.

- .github/CODEOWNERS
- docs/extension.md
- server/api.go
- server/middleware/middleware.go

Please check your commits then fix to not include those files.

morrisson:
@giyyanan Looks good. Thanks :)

PR #736, opened 2019-04-04
[CW-1187] update go modules

jkmar:


zimnx:
Please sync this merge with update of versions in our go.mod

PR #737, opened 2019-04-05
CW-997 ip addres validation

kuba2:
Fixes validation in case when ip/cidr has redundant zeros
in the last octet. Ips/cidrs like 1.2.3.001, 1.2.3.023/24
won't be allowed.


kuba2:
Api tests have passed.

PR #739, opened 2019-04-25
Rename codegangsta to urfave

morrisson:
codegangsta/cli now moves to urfave/cli. Updating codegangsta to urfave
is recommended.

Also, cli has very old bug for cli.StringSliceFlag but it's modified
only in v2, unreleased version. Our code makes a workaround for scope
flag.


morrisson:
Let me merge this PR because it only contains very obvious changes.

PR #740, opened 2019-04-26
BL-383 Token scope support for openapi doc

morrisson:


morrisson:
Expects PR 739 come first

morrisson:
@cierpuchaw @jkmar @michalzurawski @zimnx Ping to reviewers.

cli/template.go
```diff
@@ -272,25 +274,62 @@ func doTemplate(c *cli.Context) {
...
-		BuildScopedToTenant()
+func calculateAllowedProperties(schemas []*SchemaWithPolicy, scopes []schema.Scope, role string, manager *schema.Manager, keystoneV2Compatible bool) {
+	if keystoneV2Compatible {
+		auth := schema.NewAuthorizationBuilder().
+			WithRoleIDs(role).
+			WithKeystoneV2Compatibility().
+			BuildScopedToTenant()
 
+		calculateAllowedPropertiesForAuth(schemas, auth, manager)
+	} else {
```
cierpuchaw: Nit: I'd prefer to return in the true branch, so that the else branch and on level of indentation are not needed.

go.mod
```diff
@@ -79,13 +79,13 @@ require (
 	github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2 // indirect
```
cierpuchaw: `go.mod` and `go.sum` don't need to be changed.

cli/template.go
morrisson: done

go.mod
morrisson: done

PR #741, opened 2019-05-01
OpenAPI template fix: PUT/POST response follows 'read' policy

morrisson:


morrisson:
'read' policy is applied in this line https://github.com/cloudwan/gohan/blob/master/server/resources/resource_management.go#L189

morrisson:
Will merge it because it's obvious fix.

PR #742, opened 2019-05-02
[BL-384] API Document Gen for LIST query parameters

morrisson:


morrisson:
It basically modifies template file fix for OpenAPI then it's confirmed in UI. Let me merge this.

PR #744, opened 2019-05-07
[CW-1339] add domain id to context

jkmar:


zimnx:
Documentation needs to be updated.

PR #745, opened 2019-05-13
[CW-1370] IsTenantIDValid available in goext.IAuth

zimnx:
This function checks whether provided ID is known to IdentityService


extension/goext/auth.go
```diff
@@ -23,4 +25,12 @@ type IAuth interface {
 	GetTenantName(context Context) string
 	// IsAdmin reports whether context belongs to admin
 	IsAdmin(context Context) bool
+	// IsTenantIDValid checks whether given tenant_id exists in backend.
+	// Returns goext.NotValidTenantIDErr if not valid or other unspecified error
+	// and nil when provided id is valid.
+	IsTenantIDValid(ctx Context, id string) error
```
jkmar: why `error` not `bool`

cierpuchaw: Maybe `ValidateTenantID` would be a better name.

extension/goplugin/auth.go
```diff
@@ -55,32 +57,75 @@ func (a *Auth) IsAdmin(context goext.Context) bool {
...
+func getIdentityServiceFromContext(ctx goext.Context) (middleware.IdentityService, error) {
+	rawIdentityService, err := getFromContext(ctx, "identity_service")
+	if err != nil {
+		return nil, err
+	}
+
+	identityService, ok := rawIdentityService.(middleware.IdentityService)
+	if !ok {
+		err := newInvalidTypeErr("identity_service")
+		log.Warning(err.Error())
```
jkmar: maybe add some context to this error

cloud/keystone_test.go
```diff
@@ -49,6 +49,29 @@ var _ = Describe("Keystone client", func() {
...
+		BeforeEach(func() {
+			server.AppendHandlers(
+				ghttp.RespondWithJSONEncoded(201, getV3TokensScopedToTenantResponse()),
+				ghttp.RespondWithJSONEncoded(200, getV3TenantsResponse()),
+			)
+			setupV3Client()
+		})
+
+		It("Returns NotValidTenantIDErr when tenant id is not known", func() {
+			err := client.IsTenantIDValid("some-random-id")
```
jkmar: `Expect(client.IsTenantIDValid("some-random-id")).To(MatchError(NotValidTenantIDErr))`

cloud/keystone_test.go
```diff
@@ -49,6 +49,29 @@ var _ = Describe("Keystone client", func() {
...
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(Equal(NotValidTenantIDErr))
+		})
+
+		It("Returns nil when tenant id is known", func() {
+			err := client.IsTenantIDValid("1234")
```
jkmar: `Expect(client.IsTenantIDValid("1234")).To(Succeed())`

cierpuchaw:
A high level question: what about domain ids?

server/middleware/middleware.go
```diff
@@ -295,6 +296,10 @@ func (i *NobodyIdentityService) GetServiceTokenID() string {
 	return ""
 }
 
+func (i *NobodyIdentityService) IsTenantIDValid(id string) error {
+	return fmt.Errorf("%s tenant_id is invalid", id)
```
jkmar: I'm not sure if this is correct - `NobodyIdentityService` returns "Nobody" as tenant id
https://github.com/cloudwan/gohan/pull/745/files#diff-f2c62ab5c95abc1469d5f5a41bf5321cL262

morrisson: At least, error message should be changed. If we do not want to expect `IsTenantIDValid` operation for nobody auth case, something like `invalid operation` error message should be returned.

extension/goext/auth.go
zimnx: It should be either `error` or `(bool, error)`. 
In first approach, package must expose special value to distinguish whether tenant id is invalid or whether something unusual happened. 
In second apprach, `bool` returned distinguish whether provided value is valid or not, which in my opinion better express logic.
I'll change it to second apprach.

extension/goplugin/auth.go
zimnx: MIO error itself carries enough context. I've refactored a little bit to remove duplication

cloud/keystone_test.go
zimnx: done

cloud/keystone_test.go
zimnx: done

cloud/keystone_test.go
```diff
@@ -49,6 +49,43 @@ var _ = Describe("Keystone client", func() {
...
+
+		setupKeystone := func() {
+			server.AppendHandlers(
+				ghttp.RespondWithJSONEncoded(201, getV3TokensScopedToTenantResponse()),
+				ghttp.RespondWithJSONEncoded(200, getV3TenantsResponse()),
+			)
+			setupV3Client()
+		}
+
+		BeforeEach(func() {
```
jkmar: Empty `BeforeEach`?

cloud/keystone_test.go
```diff
@@ -49,6 +49,43 @@ var _ = Describe("Keystone client", func() {
...
+		})
+
+		It("Returns false when tenant id is not known", func() {
+			setupKeystone()
+
+			valid, err := client.ValidateTenantID("some-random-id")
```
jkmar: nit: there is a short cut for that (in case error should be nil):
```go
Expect(client.ValidateTenantID("some-random-id")).To(BeFalse()))
```

zimnx:
> A high level question: what about domain ids?

I thought about same thing, but lets do it in separate ticket.

cloud/keystone_test.go
zimnx: Oops, thanks

cloud/keystone_test.go
zimnx: I prefer explicit check

PR #747, opened 2019-05-14
[CW-1408] ValidateDomainID available in goext.IAuth

zimnx:


cloud/keystone.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/cloudwan/gohan/schema"
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/domains"
```
cierpuchaw: How about aliasing this to `v3domains` for consistency with 2 imports below this one?

cloud/keystone.go
```diff
@@ -243,6 +245,32 @@ func (client *keystoneV3Client) getTenant(filter func(*v3tenants.Project) bool)
 	return result, nil
 }
 
+func (client *keystoneV3Client) getDomain(filter func(domain *domains.Domain) bool) (*domains.Domain, error) {
+	opts := domains.ListOpts{}
```
cierpuchaw: `ListOpts` has a `Enabled` filter. Shouldn't we be interested only in enabled domains?

cloud/keystone_test_utils.go
```diff
@@ -186,3 +186,16 @@ func getV3TenantsResponse() interface{} {
 		},
 	}
 }
+
+func getV3DomainsResponse() interface{} {
+	return map[string]interface{}{
+		"domains": []interface{}{
+			map[string]interface{}{
+				"id":          "Default",
+				"name":        "default",
```
cierpuchaw: This is test data so it doesn't really matter, but in reality it's the other way around: id is "default" and name is "Default".

cloud/keystone.go
zimnx: Good idea. Added to getTenants too.

cloud/keystone.go
zimnx: why not

cloud/keystone_test_utils.go
zimnx: done

PR #749, opened 2019-05-18
Circle ci changes

giyyanan:


extension/goplugin/schemas_test.go
```diff
@@ -623,9 +623,9 @@ var _ = Describe("Schemas", func() {
 		)
 
 		BeforeEach(func() {
-			if os.Getenv("MYSQL_TEST") != "true" {
+			//if os.Getenv("MYSQL_TEST") != "true" {
 				Skip("Locks are only valid in MySQL")
-			}
+			//}
```
morrisson: I made a comment offline that you do not need to fix failed tests in this context because these tests are not maintained for a long time. However, this way of commenting out is not good. You should delete those two lines, then make a comment something like `TODO: tests should be fixed`

extension/goplugin/schemas_test.go
```diff
@@ -270,12 +270,12 @@ var _ = Describe("Schemas", func() {
 			c, err := testSchema.Count(goext.Filter{}, context)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(c).To(Equal(uint64(0)))
-			for i := 0; i < 2; i++ {
+			for i := 10; i < 12; i++ {
```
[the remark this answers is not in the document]
giyyanan: just to rectify duplicate issues..the logic is not changed in the test cases..this wouldn't make any difference except for the start number

tests/test_schema.yaml
```diff
@@ -1284,6 +1292,7 @@ schemas:
         - string
         - "null"
         unique: false
+        sql: varchar(255) null
```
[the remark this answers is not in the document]
giyyanan: without this the tests fail with a foreign key reference error

https://circleci.com/gh/cloudwan/gohan/3081

tests/test_schema.yaml
```diff
@@ -699,6 +699,7 @@ schemas:
         type: string
         unique: false
         description: Network ID
+        sql: varchar(255) not null
```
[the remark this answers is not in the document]
giyyanan: without this the tests fail with a foreign key reference error

https://circleci.com/gh/cloudwan/gohan/3081

tests/test_schema.yaml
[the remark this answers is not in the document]
morrisson: I am now trying something on `sql-remove-circleCI-changes` branch from your branch but I found something weird. It looks as long as we use the same branch, DB is not cleared so it results in unstable testing. I am now trying to find such an information, but it looks some docker container is cached for performance.

Actually, our test will do `drop database` before testing https://github.com/cloudwan/gohan/blob/circleCI-changes/run_test.sh#L8 , but it will not be executed in circleCI because we do not have mysql-client on circle CI. This commit https://github.com/cloudwan/gohan/commit/e7566bd31f3f0a97b9833260fede527f75af1514 could such a fix so can you try that? Thanks.

.circleci/config.yml
```diff
@@ -1,24 +1,28 @@
...
-  build:
-    docker:
-      - image: circleci/golang:1.11.4
-    working_directory: /go/src/github.com/cloudwan/gohan
-    environment:
-        GO111MODULE: 'on'
-    steps:
-      - checkout
-      - run: go version
-      - run: make deps gen lint build
```
cierpuchaw: Why is this job removed? I know that target `test` depends on `build` and that depends on `deps`, so we aren't losing those steps. But we don't run linters anymore on CI with this change.

db/sql/sql.go
```diff
@@ -1343,7 +1343,7 @@ func (lk Like) ToSql() (sql string, args []interface{}, err error) {
 			return
 		}
 
-		expr = fmt.Sprintf("%s %s ? ESCAPE '\\'", key, opr)
+		expr = fmt.Sprintf("%s %s ? ESCAPE '\\\\'", key, opr)
```
cierpuchaw: Tiny nit: you can use a raw string literal here to reduce the amount of backslashes e.g. ``` `%s %s ? ESCAPE '\\'` ```

db/sql/sql_test.go
```diff
@@ -52,7 +52,7 @@ var _ = Describe("Sql", func() {
 	BeforeEach(func() {
 		var dbType string
 		if os.Getenv("MYSQL_TEST") == "true" {
-			conn = "gohan:gohan@/gohan_test"
+			conn = "root@/gohan_test"
```
cierpuchaw: Why sometimes we are forcing a TCP connection and sometimes (here) we're not?

extension/goplugin/schemas_test.go
```diff
@@ -623,9 +624,9 @@ var _ = Describe("Schemas", func() {
 		)
 
 		BeforeEach(func() {
-			if os.Getenv("MYSQL_TEST") != "true" {
-				Skip("Locks are only valid in MySQL")
-			}
+
+			//TODO: tests must be fixed for MYSQL DB as that is implemented by default in CircleCI
```
cierpuchaw: What has to be fixed to have those tests running?

morrisson:
@jkmar No, that means mysql container kept running 6 mins until this test process has been done. You also can see total execution time then you should be able to find total time and mysql container (running) time is almost the same.

extension/goplugin/schemas_test.go
morrisson: Not evaluated but it's not maintained long time. I think it's ok to be done after this PR because even now it's not running.

cierpuchaw: Ok, I was asking more out of curiosity if this is an easy thing or not. I have no problem with fixing that in a future PR.

db/sql/sql_test.go
giyyanan: changed connections to use tcp

db/sql/sql.go
giyyanan: done

PR #750, opened 2019-05-22
Fix CircleCI random result issue

morrisson:


.circleci/config.yml
```diff
@@ -1,6 +1,16 @@
...
+  build:
+    docker:
+      - image: circleci/golang:1.11.4
+    working_directory: /go/src/github.com/cloudwan/gohan
+    environment:
+        GO111MODULE: 'on'
+    steps:
+      - checkout
+      - run: go version
+      - run: make deps gen lint build
```
morrisson: As per @cierpuchaw 's comment, we can simply test lint and build here. I said that it's ok to remove, but let's revert this job .

.circleci/config.yml
```diff
@@ -25,7 +35,6 @@ jobs:
       - run: echo $MYSQL_TEST
       - run: curl -L https://github.com/coreos/etcd/releases/download/v3.1.7/etcd-v3.1.7-linux-amd64.tar.gz -o etcd-v3.1.7-linux-amd64.tar.gz
       - run: tar xzvf etcd-v3.1.7-linux-amd64.tar.gz
-      - run: sudo apt-get install mysql-client
```
morrisson: Modified `run_tests.sh` instead. Now, mysql client is no longer required for CircleCI env.

db/db_test.go
```diff
@@ -133,7 +133,7 @@ var _ = Describe("Database operation test", func() {
 			dataStore, err = dbutil.ConnectDB(dbType, conn, db.DefaultMaxOpenConn, options.Default())
 			Expect(err).ToNot(HaveOccurred())
 
-			for _, s := range manager.Schemas() {
+			for _, s := range manager.OrderedSchemas() {
```
morrisson: Actually, this is the biggest problem why this random failure happened. MySQL strictly checks table dependencies, so `OrderedSchemas` function is used in gohan server for this purpose. However, this test was not maintained for MySQL, so table creations had been randomly happened.

db/sql/sql_test.go
```diff
@@ -109,8 +109,8 @@ var _ = Describe("Sql", func() {
 		})
 
 		insertTwoRecords := func() {
-			Expect(tx.Exec(ctx, "INSERT INTO `tests` (`id`, `tenant_id`) values ('id1', 'tenant1')")).To(Succeed())
-			Expect(tx.Exec(ctx, "INSERT INTO `tests` (`id`, `tenant_id`) values ('id2', 'tenant2')")).To(Succeed())
+			Expect(tx.Exec(ctx, "INSERT INTO `tests` (`id`, `tenant_id`, `domain_id`) values ('id1', 'tenant1', 'domain')")).To(Succeed())
```
morrisson: With `test_schemas.yaml` fix, `domain_id` is now required property. Therefore, we need to add domain_id here. Without this, sqlite3 test failed.

extension/goplugin/schemas_test.go
```diff
@@ -270,13 +271,13 @@ var _ = Describe("Schemas", func() {
 			c, err := testSchema.Count(goext.Filter{}, context)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(c).To(Equal(uint64(0)))
-			for i := 10; i < 12; i++ {
-				createdResource.ID = string(i)
+			for i := 0; i < 2; i++ {
+				createdResource.ID = strconv.Itoa(i)
```
morrisson: `string(0)` becomes weird result. You can try it in golang playground. By using strconv, you do not hit any issue if `i` starts with 0.

run_test.sh
```diff
@@ -2,7 +2,7 @@
 
 # Run Unit Test for mysql
 
-if [[ $MYSQL_TEST == "true" ]]; then
+if [ $MYSQL_TEST == "true" ] && [ $CIRCLECI != "true" ]; then
```
morrisson: If this script is running on CircleCI env, this env value becomes true. If we run the test on CircleCI, we can get fresh mysql env everytime. Therefore, `drop database` is not required. In this sense, mysql-client is also not required.

run_test.sh
```diff
@@ -15,7 +15,7 @@ ETCD_PID=$!
 echo "mode: count" > profile.cov
 
 # Standard go tooling behavior is to ignore dirs with leading underscors
-for dir in $(find . -maxdepth 10 -not -path './.git*' -not -path '*/_*' -not -path './vendor/*' -not -path '*/test_data/*' -not -path '*/goext_example*' -type d);
+for dir in $(find . -maxdepth 10 -not -path './.git*' -not -path '*/_*' -not -path './vendor/*' -not -path '*/test_data/*' -not -path '*/goext_example*' -type d | sort);
```
morrisson: This is also the second reason why the unit tests are randomly picked. By using `sort` command, the order of running tests become the same everytime.

tests/test_schema.yaml
```diff
@@ -800,6 +800,9 @@ schemas:
     - id
     - tenant_id
     - domain_id
+    required:
+    - tenant_id
+    - domain_id
```
morrisson: `required` will append `not null`. From gohan perspective, this is more proper way to define the properties.

tests/test_schema.yaml
```diff
@@ -973,6 +976,9 @@ schemas:
...
```
morrisson: ditto

tests/test_schema.yaml
```diff
@@ -1358,7 +1361,6 @@ schemas:
         - "null"
         unique: false
         description: Attach point
-        sql: varchar(255) null
```
morrisson: Above four fixes should not be required.

extension/goplugin/schemas_test.go
cierpuchaw: Fun fact: `string(1)` is also not a useful ID value ;) This used one character strings with the character being `i`th ASCII character. For example `string(65)` results in `"A"`.

Thanks for this change, this was probably the intention of the original author ;)

run_test.sh
cierpuchaw: Are we depending somewhere on the order of test execution?

PR #751, opened 2019-05-23
[BL-389] ListAPI Query by "any_of"

jkmar:


cierpuchaw:
AFAIK @jkmar run into some problems, so maybe it would be better to start reviewing this once he green-lights it on Monday or later

jkmar:
@morrisson @cierpuchaw 
It wasn't even working last week (some issue with nested boolean values and also custom filters added directly to a filter), but now it should be working. I've updated docs and added tests for multiple cases with `is_public` and `tenant_id`

extension/goext/filter/filter.go
```diff
@@ -41,3 +41,23 @@ func False() FilterElem {
...
+	}
+	if len(filters) == 1 {
+		return filters[0]
+	}
+	return And(filters...)
+}
+
+func MaybeEmptyOrFilter(filters ...FilterElem) FilterElem {
+	if len(filters) == 0 {
+		return False()
```
cierpuchaw: Why `False()`? `MaybeEmptyAndFilter` returns a true filter in such case and alternative has a less restrictive contract than conjunction.

jkmar: This is actually correct - empty `OR` should be `false`. It may seem weird at first glance, but lets consider the following:
* In order for `AND` to be `false` at least one of the items must be `false`, in the empty set there are no `false` items, so `AND` on the empty set is `true`
* Similarly `OR` is true if and only if one of the items is `true`, so `OR` on the empty set is `false`, since there are no `true` items
This is also how `all` and `any` work in python:
```python
>>> all([])
True
>>> any([])
False
>>>
```

cierpuchaw: Ack, from wikipedia:
https://en.wikipedia.org/wiki/Logical_conjunction "the empty conjunction (AND-ing over an empty set of operands) is often defined as having the result 1"
https://en.wikipedia.org/wiki/Logical_disjunction "the empty disjunction (OR-ing over an empty set of operands) is generally defined as false. "

PR #752, opened 2019-06-07
Update github.com/pkg/errors to 0.8.1 and tidy go.mod/go.sum.

cierpuchaw:


cierpuchaw:
We should merge only when we're ready.

PR #754, opened 2019-06-11
[CW-1428] Expose tenant and domain id from auth

michalzurawski:


schema/policy.go
```diff
@@ -364,6 +373,10 @@ func (auth *AdminAuthorization) IsAdmin() bool {
 	return true
 }
 
+func (auth *AdminAuthorization) IsDomainAdmin() bool {
+	return true
```
michalzurawski: what do you think - should be true or false? in both cases we can distinguish admin from domain admin

jkmar: Doesn't this type represent admin scoped to admin?

michalzurawski: you mean `AdminAuthorization`? it does

jkmar: Then if think it should be `false`, since it's not a **domain** admin

michalzurawski: Ok, i just followed the logic that admin is a superset of domain admin so that's why i set it to true. Guys, could you vote with thumbs up if i should change it to `false` and thumbs down otherwise?

zimnx: Then it's better to call this function `HasDomainAdminPrivileges`. I agree with @jkmar.

michalzurawski: done in different manner

extension/goext/auth.go
```diff
@@ -21,6 +21,10 @@ type IAuth interface {
 	HasRole(context Context, role string) bool
 	// GetTenantName return name from the given context
 	GetTenantName(context Context) string
+	// GetTenantID return id of the tenant from the given context
+	GetTenantID(context Context) string
```
jkmar: Both `tenant_id` and `domain_id` (since [[CW-1339] add domain id to context](https://github.com/cloudwan/gohan/pull/744)) are included in context, what is the point of that functions then? Aren't they redundant?

PR #755, opened 2019-06-12
[CW-1453] Do no allow leading zero in cidr mask

michalzurawski:


schema/formats.go
```diff
@@ -95,13 +98,23 @@ func (f ipv4NetworkFormatChecker) IsFormat(input string) bool {
...
 		return false
 	}
+	if !maskIsValid(mask) {
+		return false
+	}
 	return hostIP.Equal(netIP)
 }
 
+func maskIsValid(mask string) bool {
+	if len(mask) == 2 && mask[0] == '0' {
```
jkmar: Is "001" valid mask?
According to this function it is.

schema/formats.go
```diff
@@ -120,7 +133,13 @@ func (f ipv4AddressWithCidrFormatChecker) IsFormat(input string) bool {
 }
 
 func extractIpFromCidrFormattedString(input string) string {
-	return strings.Split(input, "/")[0]
+	ip, _ := extractIpAndMaskFromCidrFormattedString(input)
+	return ip
+}
+
+func extractIpAndMaskFromCidrFormattedString(input string) (string, string) {
+	ipAndMask := strings.Split(input, "/")
```
jkmar: If input is "1.2.3.4/5/6", then it's not valid, but "1.2.3.4" and "5" values are returned.
Also it panics for "1.2.3.4".
It should rather be like this: https://github.com/coreswitch/netutil/blob/master/prefix.go#L71.

michalzurawski: we validate all that cases before - added tests for that

PR #756, opened 2019-06-27
[BL-396] Fix condition checks for custom actions

gryxon:
added checks of policy conditions during performing custom actions

### DIFF
--- a/server/resources/resource_management.go
+++ b/server/resources/resource_management.go
@@ -576,11 +576,12 @@ func getFilterFromPolicy(
 	resourceID string,
 	resourceSchema *schema.Schema,
 	policy *schema.Policy,
+	action string,
 ) transaction.Filter {
 	filter := transaction.IDFilter(resourceID)
 
 	currCond := policy.GetCurrentResourceCondition()
-	extendFilterByTenantAndDomain(resourceSchema, filter, schema.ActionUpdate, currCond, auth)
+	extendFilterByTenantAndDomain(resourceSchema, filter, action, currCond, auth)
 	currCond.AddCustomFilters(resourceSchema, filter, auth)
 
 	return filter
@@ -592,9 +593,10 @@ func checkIfResourceExistsForPolicy(
 	resourceID string,
 	resourceSchema *schema.Schema,
 	policy *schema.Policy,
+	action string,
 	preTransaction transaction.Transaction,
 ) (bool, error) {
-	filter := getFilterFromPolicy(auth, resourceID, resourceSchema, policy)
+	filter := getFilterFromPolicy(auth, resourceID, resourceSchema, policy, action)
 
 	return checkIfResourceExists(context, filter, resourceSchema, preTransaction)
 }
@@ -692,7 +694,7 @@ func CreateResource(
 		return err
 	}
 
-	if !applyFilterToResource(resource.Data(), getFilterFromPolicy(auth, resource.ID(), resourceSchema, policy)) {
+	if !applyFilterToResource(resource.Data(), getFilterFromPolicy(auth, resource.ID(), resourceSchema, policy, schema.ActionUpdate)) {
 		return ResourceError{err, "", Unauthorized}
 	}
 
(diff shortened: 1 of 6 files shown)


server/resources/resource_management.go
```diff
@@ -1166,6 +1167,66 @@ func ActionResource(context middleware.Context, resourceSchema *schema.Schema,
 	return fmt.Errorf("no response")
 }
 
+func checkIfActionIsAllowedForUser(context middleware.Context, dataStore db.DB, dataMap map[string]interface{},
```
cierpuchaw: If a user without access to a resource, but with knowledge of its ID, calls an action on it, we correctly reject such attempts, but instead of returning 404, we return 401. That betrays the fact that such resource exists.

tests/test_schema.yaml
```diff
@@ -99,6 +99,48 @@ extensions:
 networks: []
 
 policies:
+- action: singular_is_owner
```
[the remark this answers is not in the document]
cierpuchaw: What I've meant is that if we are defining policies for some actions in yaml we should also define those actions there.

server/resources/resource_management_test.go
```diff
@@ -2202,6 +2238,72 @@ var _ = Describe("Resource manager", func() {
...
+			context["domain_id"] = auth.DomainID()
+			context["auth"] = auth
+			context["response"] = ""
+		}
+
+		Context("Custom action isolation tests", func() {
+			var NoProblem resources.ResourceProblem = -1
+
+			DescribeTable("Performing action", func(auth schema.Authorization, resourceID string, action *schema.Action,
+				expectedResourceProblem resources.ResourceProblem, shouldFail bool) {
```
cierpuchaw: We don't really need `shouldFail`. It's true if and only if `expectedResourceProblem != NoProblem`.

server/resources/resource_management_test.go
```diff
@@ -2202,6 +2238,72 @@ var _ = Describe("Resource manager", func() {
...
+				Entry("Should not fail when performed action matched to policy without is_owner on own resource as member",
+					createAuthToken(tenantMemberAuthScope), memberResourceID, &singularPublicAction, NoProblem, false),
+				Entry("Should not fail when performed action matched to policy without is_owner on domain domain as domain admin",
+					createAuthToken(domainAdminAuthScope), adminResourceID, &singularPublicAction, NoProblem, false),
+				Entry("Should not fail when performed action on different domain resource as cloud admin",
+					createAuthToken(superAdminAuthScope), otherDomainResourceID, &singularPublicAction, NoProblem, false),
+				Entry("Should not fail when performed action matched to policy without is_owner on not own resource as member",
+					createAuthToken(tenantMemberAuthScope), adminResourceID, &singularPublicAction, NoProblem, false),
+				Entry("Should not fail when performed action matched to policy without is_owner_domain on not own resource nor domain as domain admin",
+					createAuthToken(domainAdminAuthScope), otherDomainResourceID, &singularPublicAction, NoProblem, false),
```
cierpuchaw: We only have one policy for `singular_public` and it's for `Member`s, why does this entry pass? Same goes for the other entry using domain admin.

tests/test_schema.yaml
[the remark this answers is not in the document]
gryxon: Added

server/resources/resource_management_test.go
gryxon: In the test_schema.yaml there is a policy:
```
- action: '*'
  effect: allow
  id: admin_statement
  principal: admin
  resource:
    path: .*
```

gryxon:
@jkmar @michalzurawski @morrisson kind reminder about this PR. Could you take a look, please?

server/resources/resource_management_test.go
cierpuchaw: Discussed offline, waiting for fix.

gryxon: I will not remove this policy to avoid breaking old tests, I have added extra policy for avoid confusing about this test case.

PR #758, opened 2019-07-03
[CW-1489] Regenerate ISchema mock

zimnx:
I forgot to commit one change to previous PR https://github.com/cloudwan/gohan/pull/757


zimnx:
@jkmar probably we could. Go ahead and create an issue for that. 
For now let's focus on merging this fix, because current master is broken.

PR #759, opened 2019-07-12
Performance improvement

michalzurawski:
Disable join by default for extensions.
Please note that this PR _**breaks backward compatibility**_ for extensions which fetch/list resources directly by [transcation interface](https://github.com/cloudwan/gohan/blob/master/extension/goext/transaction.go#L80) and pass `nil` to `listOptions`.
In any other case, compatibility is preserved including API requests (they still make join for any query) and locks (joins are managed explicitly by [lock policy](https://github.com/cloudwan/gohan/blob/master/db/sql/sql.go#L917)

### DIFF
--- a/extension/goplugin/transaction.go
+++ b/extension/goplugin/transaction.go
@@ -160,7 +160,14 @@ func (t *Transaction) List(ctx context.Context, schema goext.ISchema, filter goe
 	if err := ctx.Err(); err != nil {
 		return nil, 0, ctx.Err()
 	}
-	data, _, err := t.tx.List(context.Background(), t.findRawSchema(schemaID), transaction.Filter(filter), nil, (*pagination.Paginator)(paginator))
+	var o *transaction.ViewOptions
+	if listOptions != nil {
+		o = &transaction.ViewOptions{
+			Details: listOptions.Details,
+			Fields: listOptions.Fields,
+		}
+	}
+	data, _, err := t.tx.List(context.Background(), t.findRawSchema(schemaID), transaction.Filter(filter), o, (*pagination.Paginator)(paginator))
 	if err != nil {
 		return nil, 0, err
 	}
--- a/server/resources/resource_management.go
+++ b/server/resources/resource_management.go
@@ -239,7 +239,7 @@ func GetResourcesInTransaction(context middleware.Context, resourceSchema *schem
 		return err
 	}
 
-	var o *transaction.ViewOptions
+	o := &transaction.ViewOptions{Details: true}
 	r, ok := context["http_request"].(*http.Request)
 	if ok {
 		o = listOptionsFromQueryParameter(r.URL.Query())
@@ -480,7 +480,7 @@ func GetSingleResource(context middleware.Context, dataStore db.DB, resourceSche
 //GetSingleResourceInTransaction get resource in single transaction
 func GetSingleResourceInTransaction(context middleware.Context, resourceSchema *schema.Schema, resourceID string, tenantIDs []string, domainIDs []string) (err error) {
 	defer MeasureRequestTime(time.Now(), "get.single.in_tx", resourceSchema.ID)
-	var options *transaction.ViewOptions
+	options := &transaction.ViewOptions{Details: true}
 	r, ok := context["http_request"].(*http.Request)
 	if ok {
 		options = listOptionsFromQueryParameter(r.URL.Query())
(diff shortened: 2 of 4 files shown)


zimnx:
Do we have any performance measurements that proves such change helps? It would be silly to break backward compatibility without any benefit.

jkmar:
Can join still be used?

michalzurawski:
> Do we have any performance measurements that proves such change helps? It would be silly to break backward compatibility without any benefit.

Currently, we have just measured that UT runs ~1.5 time faster. What kind of tests would you like to run?

> Can join still be used?

Yes, though in GoExtensions one must use directly `Tx.List`. And as mentioned before it is still a default value for API queries.

cierpuchaw:
> > Do we have any performance measurements that proves such change helps? It would be silly to break backward compatibility without any benefit.
> 
> Currently, we have just measured that UT runs ~1.5 time faster. What kind of tests would you like to run?

On my dev machine it's more like a 25% time reduction instead of mentioned 33% (aka 1.5x faster), but that's still substantial. Great work!

server/resources/resource_management.go
```diff
@@ -480,7 +480,7 @@ func GetSingleResource(context middleware.Context, dataStore db.DB, resourceSche
 //GetSingleResourceInTransaction get resource in single transaction
 func GetSingleResourceInTransaction(context middleware.Context, resourceSchema *schema.Schema, resourceID string, tenantIDs []string, domainIDs []string) (err error) {
 	defer MeasureRequestTime(time.Now(), "get.single.in_tx", resourceSchema.ID)
-	var options *transaction.ViewOptions
+	options := &transaction.ViewOptions{Details: true}
```
jkmar: why join is used for get single resource?

server/resources/resource_management.go
```diff
@@ -480,7 +480,7 @@ func GetSingleResource(context middleware.Context, dataStore db.DB, resourceSche
...
 	r, ok := context["http_request"].(*http.Request)
 	if ok {
 		options = listOptionsFromQueryParameter(r.URL.Query())
```
jkmar: options are overwritten here

extension/goplugin/transaction.go
```diff
@@ -160,7 +160,14 @@ func (t *Transaction) List(ctx context.Context, schema goext.ISchema, filter goe
 	if err := ctx.Err(); err != nil {
 		return nil, 0, ctx.Err()
 	}
-	data, _, err := t.tx.List(context.Background(), t.findRawSchema(schemaID), transaction.Filter(filter), nil, (*pagination.Paginator)(paginator))
+	var o *transaction.ViewOptions
```
jkmar: why this is only used for `List` but not for `Fetch`?

server/resources/resource_management.go
michalzurawski: To preserve backward compatibility for API. Please note that currently UI depends on that.

jkmar: but in that case it's rewritten 2 lines below

server/resources/resource_management.go
michalzurawski: intended. again, we don't want to change behavior of the API which allows overriting Options and for which join is the default behavior

extension/goplugin/transaction.go
michalzurawski: i'm not sure why we have such interface for transaction but changing it will break compatibility even more.
if someone wants resource with all the related resources in response then he has to marshall it by himself - IMHO if that's not the problem then fetching one resource with List instead of Fetch also shouldn't be.

jkmar: Ok, it looks like `Fetch` [uses list internally](https://github.com/cloudwan/gohan/blob/master/db/sql/sql.go#L1060), so that is fine,
but there is an inconsistency there with `LockList` which [sets join to true](https://github.com/cloudwan/gohan/blob/master/db/sql/sql.go#L948), when options are nil (in contrast to list).

server/resources/resource_management.go
michalzurawski: > (not counting JS)

I'm afraid this sentence disqualifies the change as we still rely on JS extensions in tests. And there is no benefit in discarding compatibility with JS just to remove single `if` statement

server/resources/resource_management.go
michalzurawski: same as above

extension/goplugin/transaction.go
michalzurawski: that code is pretty confusing because as [we build sql query few lines above](https://github.com/cloudwan/gohan/blob/master/db/sql/sql.go#L935) - here we are using it only to parse data. These lines don't make any sense as if we wanted related resources we should use join also in query. Let me just remove them

PR #761, opened 2019-08-22
[BL-329] Add tenant and domain isolation for related resources

michalzurawski:


tests/test_schema.yaml
```diff
@@ -585,6 +585,27 @@ policies:
     - is_owner
   resource:
     path: /v2.0/owned_resource.*
+- action: '__attach__'
+  effect: allow
+  id: tenant_isolation_domain_admin_server_to_network
+  principal: admin
+  relation_property: network_id
+  target_condition:
```
jkmar: Why both `is_owner` and `skip_tenant_domain_check`?
It looks like `skip_tenant_domain_check` overrides `is_owner`.

tests/test_schema.yaml
```diff
@@ -585,6 +585,27 @@ policies:
...
+  - is_owner
+  - skip_tenant_domain_check
```
jkmar: Why both `is_owner` and `skip_tenant_domain_check`?
It looks like `skip_tenant_domain_check` overrides `is_owner`.

tests/test_schema.yaml
```diff
@@ -585,6 +585,27 @@ policies:
     - is_owner
   resource:
     path: /v2.0/owned_resource.*
+- action: '__attach__'
+  effect: allow
+  id: tenant_isolation_domain_admin_server_to_network
```
jkmar: For consistency wouldn't it be better to have `scope` in this policy as well?

server/server_test.go
```diff
@@ -559,6 +559,43 @@ var _ = Describe("Server package test", func() {
 					testURL("PUT", url, memberTokenID, testResource, http.StatusUnauthorized)
 				})
 			})
+
+			It("should not create resource in different tenant than relation", func() {
```
jkmar: Since these test are basically the same, I suggest using `DescribeTable`.

server/server_test.go
```diff
@@ -559,6 +559,43 @@ var _ = Describe("Server package test", func() {
...
+				}
+				testURL("POST", networkPluralURL, memberTokenID, network, http.StatusCreated)
+				server := map[string]interface{}{
+					"name":       "Server Red",
+					"network_id": networkID,
+					"status":     "ACTIVE",
+					"tenant_id":  powerUserTenantID,
+				}
+
+				testURL("POST", serverPluralURL, memberTokenID, server, http.StatusBadRequest)
```
jkmar: I'm not sure about that, since member should not be able to use other tenants id. Also does this fail for the same reason as the next statement?

server/server_test.go
```diff
@@ -559,6 +559,43 @@ var _ = Describe("Server package test", func() {
...
+				}
+				testURL("POST", networkPluralURL, adminTokenID, network, http.StatusCreated)
+				server := map[string]interface{}{
+					"name":       "Server Red",
+					"network_id": networkID,
+					"status":     "ACTIVE",
+					"tenant_id":  powerUserTenantID,
+				}
+
+				testURL("POST", serverPluralURL, adminTokenID, server, http.StatusCreated)
```
jkmar: Since `admin` can do everything, what would happen if this was done by the `Member`, but with different policy? Or maybe `admin` is restricted in these tests.

server/resources/resource_management.go
```diff
@@ -1358,12 +1361,21 @@ func validateAttachmentRelation(
 		}
 	}
 
+	rc := policy.GetOtherResourceCondition()
```
jkmar: maybe extract this to a function

server/resources/resource_management.go
```diff
@@ -1377,17 +1389,31 @@ func validateAttachmentRelation(
 	return nil
 }
 
+func checkRelatedResource(relatedResource *schema.Resource, data map[string]interface{}) bool {
+	if expectedTenantID, ok := data[tenantIDKey].(string); ok {
```
jkmar:
```go
func equalStringKeys(first, second map[string]interface{}, key string) bool {
    if firstValue, ok := first[key].(string); ok {
        if secondValue, ok := second[key].(string); ok {
            return false
        }
    }
    return true
}
```

tests/test_schema.yaml
michalzurawski: no i does not - it does not remove tenant whole tenant isolation it just allows to create resource in different tenant/domain than related resource providing that user has access to it. This is useful for some "public" resources

server/server_test.go
michalzurawski: `since member should not be able to use other tenants id` what do you mean? user might be able to use resource from other tenant if it's "public"

server/server_test.go
michalzurawski: even admin is not able to create resource in different tenant than related resource unless `skip_tenant_domain_check` flag is provided

server/server_test.go
michalzurawski: i don't like that idea however i might create Context and move common part to BeforeEach

server/server_test.go
```diff
@@ -559,6 +559,34 @@ var _ = Describe("Server package test", func() {
 					testURL("PUT", url, memberTokenID, testResource, http.StatusUnauthorized)
 				})
 			})
+
+			Context("Creating resource in different tenant", func() {
```
gryxon: I've checked that the validation is also applied in case of update but as the different piece of code is responsible for updating and creation resources I think it's good idea to cover the case of update also in the unittests.

gryxon: What do you think about add tests for checking case when related resource has no tenant assigned. I mean the resources which are tenant or domain independent

server/resources/resource_management.go
```diff
@@ -1377,17 +1388,35 @@ func validateAttachmentRelation(
...
+		equalStringKeys(data, relatedResource.Data(), domainIDKey)
+}
+
+func equalStringKeys(first, second map[string]interface{}, key string) bool {
+	if firstValue, ok := first[key].(string); ok {
+		if secondValue, ok := second[key].(string); ok {
+			return firstValue == secondValue
+		}
+	}
+	return true
```
gryxon: I am not sure but double check that in case when resource is not related to nor tenant nor domain there is no key tenant_in/domain_id in this dict. I think it can contain just empty strings in this case.

server/server_test.go
michalzurawski: in common case tenant/domain id are not properties for which update is possible - besides it's the same logic so i don't see any added value in this scenario

michalzurawski: that's good idea, will add that

server/resources/resource_management.go
michalzurawski: tenant_id/domain_id are relation to tenant schema so they can be either proper value of null - in the second case they won't be key in data

PR #768, opened 2019-10-08
[BL-408] Use maintained go-bindata

morrisson:


go.mod
```diff
@@ -1,39 +1,43 @@
 module github.com/cloudwan/gohan
 
 require (
+	github.com/Bowery/prompt v0.0.0-20190916142128-fa8279994f75 // indirect
```
tomasz-michalski: I see no reason to update `go.mod` (and `go.sum`) now. Moreover, some of added dependencies are removed after executing `go mod tidy` (`go1.13 linux/amd64`).

Please note that when dependencies in `gohan` are updated, other projects using `gohan` must be updated as well (go require packages to have exactly the same version in plugin and main program).

morrisson: Thanks. Will not add go.mod and go.sum .

morrisson:
> `public/webui/config.json` is not present in the repository, therefore it should not be present in `util/bindata.go`

My local env was somehow dirty. Regenerate `bindata.go` after cleaning up.

Also, I added `-nometadata` flag to `go-bindata`.

PR #769, opened 2019-10-23
Fix cover accuracy by including covering all packages

jkmar:
Fix cover accuracy by including covering all packages for tests instead of just tested package.

Previously coverage report was incorrect for some packages with only tests - for example server_error_test.go.
The only limitation are goplugins which cannot be tested this way, so for them old behavior is preserved.

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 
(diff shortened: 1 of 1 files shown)


jkmar:
Uses too much memory on CircleCI

PR #773, opened 2020-01-14
[CW-2021] Add descriptive logs about errors during verifying token

tomasz-michalski:


cloud/keystone.go
```diff
@@ -91,6 +96,8 @@ func (client *keystoneV3Client) VerifyToken(token string) (schema.Authorization,
 
 	tokenResult := v3tokens.Get(client.client, token)
 	if tokenResult.Err != nil {
+		// tokenResult.Err.Error() is stripped out of useful info
+		log.Warning("Error during verifying token: %#v", tokenResult.Err)
```
tomasz-michalski: We could also return `%#v` to the user but I'm afraid that it may expose too many details. WDYT?

jkmar: Agree that returning this in error exposes too much details, so better to keep it as log.
Why not just `+v` instead of `#v`?

cloud/keystone.go
```diff
@@ -31,6 +32,10 @@ import (
 	"github.com/gophercloud/gophercloud/pagination"
 )
 
+var (
+	log = l.NewLogger()
```
cierpuchaw: It's a bit terrifying that right now we have **0** log messages from this file.

cloud/keystone.go
cierpuchaw: I'm fine with returning a terse error and having a detailed log message. But I would log this on Error level.

tomasz-michalski: @jkmar `+v` prints result of `.Error()` instead of the struct fields. See https://play.golang.org/p/wE0OWAA9cYS

tomasz-michalski: @cierpuchaw I agree, changed to `Error`.

cloud/keystone.go
tomasz-michalski: I added one more log message about `Invalid token`, because we have seen such errors lately. Other errors are improbable or quite obvious so I will leave them for now.

cloud/keystone.go
```diff
@@ -100,6 +107,7 @@ func (client *keystoneV3Client) VerifyToken(token string) (schema.Authorization,
 	// If system token needed reauth and user token is invalid, err wont be propagated neither by return nor Err field,
 	// but response is nil
 	if err != nil || !ok {
+		log.Error("Invalid token, err: %s, result: %#v", err, tokenResult)
```
michalzurawski: so now even "true errors" about invalid token will be logged as error - isn't it excessive?

tomasz-michalski: If token is invalid it means that something went wrong - probably not in gohan, but somewhere else (e.g. UI). We don't expect the user to enter the token manually, therefore I think it's fine to log it as an error. If this will be troublesome it can always be changed.

michalzurawski: is expired token also invalid? that happens pretty often since expiration time is 1 hour

cierpuchaw: That is a valid point. We can differentiate using `err` and `ok` and either use different log levels or even different log messages for actual errors and reauth.

tomasz-michalski: I think that you misinterpreted what this code does. It handles the case when system token (the one that gohan itself uses) expired and user provided invalid token.

Let me describe `gophercloud` behaviour:
1. fetch info about user token - fails (401 unauthorized) because system token is expired
2. renew system token because 401 was received
3. fetch info about user token - fails with 404, but error is not propagated (`tokenResult.Err` is `nil` and `tokenResult.Body` is `nil`)

Scenario when system token is not expired:
1. fetch info about user token - fails with 404, error is propagated (`tokenResult.Err` is not `nil`)

So, both log messages introduced in this PR should have the same level because they are triggered by the same issue (user provided invalid token), which one is printed depends on the current system state. It was suggested before to use `Error` in such case, which I did.

---

Wow, that's a lot of text about one log message ;)

PR #779, opened 2020-05-13
[CW-2742] Update Dockerfile

tomasz-michalski:


Dockerfile
```diff
@@ -1,7 +1,15 @@
-FROM golang:1.9
-MAINTAINER Leif Madsen <leif@leifmadsen.com>
+ARG GO_VERSION=1.13
 
-RUN go get github.com/cloudwan/gohan
+
+FROM golang:$GO_VERSION-alpine
+RUN apk add --no-cache g++
```
cierpuchaw, line 5: I assume that this is needed for cgo sqlite, right? If only for that then probably gcc+libc-dev would suffice.

tomasz-michalski: Yes, you are right, done. Good suggestion, thanks!

PR #782, opened 2020-07-01
CW-2593: add support for prometheus metrics in addition to graphite.

cierpuchaw:


metrics/metrics.go
```diff
@@ -1,84 +1,63 @@
...
 	l "github.com/cloudwan/gohan/log"
 	"github.com/cloudwan/gohan/util"
-	graphite "github.com/cyberdelia/go-metrics-graphite"
-	"github.com/rcrowley/go-metrics"
+	metrics "github.com/rcrowley/go-metrics"
 )
 
 var (
-	graphiteConfigs   []graphite.Config
+	exporter          metricsExporter
```
jkmar, line 28: why global variable?

metrics/prometheus.go
```diff
@@ -0,0 +1,133 @@
...
+	pe.provider.WithTimerBuckets(pe.config.timerBuckets)
+
+	pe.config.serverPath = getString(config, "path", "/metrics")
+	pe.config.serverBackoff = getDuration(config, "backoff", 5*time.Second)
+
+	return nil
+}
+
+func (pe *prometheusExporter) setServerAddress(config *util.Config, mainAddress string) error {
+	mainSplit := strings.Split(mainAddress, ":")
```
jkmar, line 84: https://golang.org/pkg/net/#SplitHostPort

metrics/prometheus.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	return nil
+}
+
+func (pe *prometheusExporter) IsReady() bool {
+	return pe.provider != nil
+}
+
+func (pe *prometheusExporter) Start() {
```
jkmar, line 108: it would be nice to have this function testable, but I understand that this is not that easy

metrics/prometheus_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+	BeforeEach(func() {
+		pe = &prometheusExporter{}
+		emptyConfig = util.NewConfig(map[string]interface{}{})
+	})
+
+	It("Has reasonable defaults after setup", func() {
+		Expect(pe.Setup(emptyConfig, ":1234")).To(Succeed())
+
+		Expect(pe.config.flushInterval).To(Equal(10 * time.Second))
```
jkmar, line 40: I've seen this being repeated later, why not just `Expect(pe.config).To(Equal())` or `thenConfigIs(pe.config, ...)?`

metrics/prometheus.go
```diff
@@ -0,0 +1,133 @@
...
+	mainSplit := strings.Split(mainAddress, ":")
+	if len(mainSplit) != 2 {
+		return errors.Errorf("Incorrect main gohan address '%s', cannot deduct prometheus metrics address", mainAddress)
+	}
+	mainPort, err := strconv.Atoi(mainSplit[1])
+	if err != nil {
+		return errors.Errorf("Incorrect main gohan address '%s', port is not a number", mainAddress)
+	}
+
+	defaultAddress := fmt.Sprintf("%s:%d", mainSplit[0], mainPort+2) // healthcheck uses +1
```
jkmar, line 93: maybe lets create an enum with port offsets, to avoid magic numbers and comments

metrics/utils.go
```diff
@@ -0,0 +1,62 @@
...
+
+	switch exporterTag {
+	case graphiteTag:
+		exporter = &graphiteExporter{}
+	case prometheusTag:
+		exporter = &prometheusExporter{}
+	}
+
+	if exporter == nil {
+		return nil, errors.Errorf("Unsupported metrics exporter under metrics/exporter: %s, must be 'graphite' or 'prometheus'",
```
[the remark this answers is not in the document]
cierpuchaw: Done. Here and in prometheus.go.

metrics/utils.go
```diff
@@ -0,0 +1,62 @@
...
+	var exporter metricsExporter
+
+	switch exporterTag {
+	case graphiteTag:
+		exporter = &graphiteExporter{}
+	case prometheusTag:
+		exporter = &prometheusExporter{}
+	}
+
+	if exporter == nil {
```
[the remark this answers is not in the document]
cierpuchaw: Done.

metrics/utils_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	BeforeEach(func() {
+		emptyConfig = util.NewConfig(map[string]interface{}{})
+	})
+
+	Context("Percentiles", func() {
+		const key = "some_key"
+
+		thenPercentilesAre := func(config *util.Config, expectedPercentiles []float64) {
+			percentiles, err := getPercentilesFrom(config, key)
```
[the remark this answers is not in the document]
cierpuchaw: Done. I've left the `thenPercentileAre` function instead of inlining it (it's only 1 line now), because it reads better IMHO.

metrics/utils.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+	"strconv"
+
+	"github.com/cloudwan/gohan/util"
+	"github.com/pkg/errors"
+)
+
+func getPercentilesFrom(config *util.Config, path string) (percentiles []float64, err error) {
+	defaultPercentiles := []string{"0.5", "0.75", "0.95", "0.99", "0.999"}
+	percentilesStr := config.GetStringList(path, defaultPercentiles)
```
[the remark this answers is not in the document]
cierpuchaw: I'm not sure I understand. Config does not return any error in case of problems, just the default value provided.

metrics/graphite.go
```diff
@@ -0,0 +1,86 @@
...
+func (ge *graphiteExporter) Start() {
+	log.Debug("Starting sending runtime metrics to graphite, config %+v", ge.configs)
+
+	for _, config := range ge.configs {
+		go graphite.WithConfig(config)
+	}
+}
+
+func (ge *graphiteExporter) getGraphiteConfig(config *util.Config) (graphiteConfigs []graphite.Config, err error) {
+	graphiteEndpoints := config.GetStringList("metrics/graphite/endpoints", []string{})
```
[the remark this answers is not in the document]
cierpuchaw: Done.
BTW I've left graphite code mostly intact, only extracted it into a new file and renamed functions to methods implementing a common interface.

metrics/graphite.go
```diff
@@ -0,0 +1,86 @@
...
+		config := baseconfig
+		config.Addr = addr
+		graphiteConfigs = append(graphiteConfigs, config)
+	}
+
+	return graphiteConfigs, nil
+}
+
+func (ge *graphiteExporter) GetFlushInterval() time.Duration {
+	return ge.configs[0].FlushInterval
```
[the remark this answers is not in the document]
cierpuchaw: Then it'll panic ;)
Actually it won't be called at all because `IsReady()` is checked first.

metrics/graphite.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	baseconfig.FlushInterval = time.Duration(config.GetInt("metrics/graphite/flush_interval_sec", 60)) * time.Second
+	baseconfig.Prefix = config.GetString("metrics/graphite/prefix", "gohan")
+	baseconfig.DurationUnit = time.Nanosecond
+	baseconfig.Registry = metrics.DefaultRegistry
+
+	for _, endpoint := range graphiteEndpoints {
+		addr, err := net.ResolveTCPAddr("tcp", endpoint)
+		if err != nil {
+			return nil, fmt.Errorf("Can't resolve graphite endpoint %s: %s", endpoint, err)
```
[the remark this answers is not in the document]
cierpuchaw: Done. Also for all returned custom errors.

metrics/graphite.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (ge *graphiteExporter) IsReady() bool {
+	return len(ge.configs) > 0
+}
+
+func (ge *graphiteExporter) Start() {
+	log.Debug("Starting sending runtime metrics to graphite, config %+v", ge.configs)
+
+	for _, config := range ge.configs {
+		go graphite.WithConfig(config)
```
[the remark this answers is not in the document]
cierpuchaw: It shouldn't. I have not seen any leaks and that's the recommended way to call this. Also it's old code (I've only moved it).

metrics/metrics.go
cierpuchaw: The public interface for metrics is free functions (e.g. `StartMetricsProcess()`) so they have to operate on some kind of singleton. I have not changed the public interface at all, notice that previously `graphiteConfigs` were global - I've simply abstracted that using an interface.

metrics/prometheus.go
cierpuchaw: Done.
Thanks, I didn't know that. It's a pity it returns port as string ;).

metrics/prometheus.go
cierpuchaw: I agree. The only thing to actually test here is the http server restarting loop. And testing that one thing is IMHO too much work for very little benefit.

metrics/prometheus.go
cierpuchaw: Done.

metrics/prometheus_test.go
cierpuchaw: There is a small catch. This test case does not check the value of serverAddress as I have other test cases dedicated for that, because there is some additional logic behind it.

If you really want, I can merge them.

metrics/prometheus.go
jkmar: It doesn't look done to me

metrics/metrics.go
jkmar: I see, can we group those global variables to some struct, or is it too much work for now?

PR #785, opened 2020-07-31
Validate Keystone V3 roles for admin authorization.

cierpuchaw:
Our intention is for admin auth to be allowed to do anything. That's
accomplished through the built-in `admin_statement` policy.
Unfortunately if we assign multiple roles to it, it's possible that a
policy with "effect: deny" for one of those roles will block some
operation, possibly making it impossible to use admin token to revert
some changes. This commit rejects such Keystone V3 tokens, effectively
forcing a desired configuration of Keystone users and role assignments.

### DIFF
--- a/cloud/keystone.go
+++ b/cloud/keystone.go
@@ -144,6 +144,10 @@ func (client *keystoneV3Client) VerifyToken(token string) (schema.Authorization,
 			WithRoleIDs(roleIDs...)
 
 		if isTokenScopedToAdminProject(&tokenResult) {
+			if err := client.validateRolesForAdmin(roleIDs); err != nil {
+				return nil, err
+			}
+
 			return builder.BuildAdmin(), nil
 		}
 		return builder.BuildScopedToTenant(), nil
@@ -226,6 +230,15 @@ func isTokenScopedToAdminProject(result *v3tokens.GetResult) bool {
 	}
 	err := result.ExtractInto(&s)
 	return (err == nil) && s.IsAdminProject
+}
+
+func (client *keystoneV3Client) validateRolesForAdmin(roleIDs []string) error {
+	if len(roleIDs) != 1 || roleIDs[0] != schema.AdminRole {
+		client.updateCounter(1, "error.token.mismatched_role_admin")
+		return fmt.Errorf("Admin token should have only a single admin role instead of '%v'", roleIDs)
+	}
+
+	return nil
 }
 
 func (client *keystoneV3Client) getTenant(filter func(*v3tenants.Project) bool) (*v3tenants.Project, error) {
--- a/cloud/keystone_test.go
+++ b/cloud/keystone_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/cloudwan/gohan/schema"
 
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/ghttp"
 )
@@ -327,7 +328,7 @@ var _ = Describe("Keystone client", func() {
 					Expect(auth.TenantName()).To(Equal("acme"))
 					Expect(auth.DomainID()).To(Equal("domain-id"))
 					Expect(auth.DomainName()).To(Equal("domain"))
-					Expect(auth.Roles()).To(Equal([]*schema.Role{{Name: "member"}}))
+					Expect(auth.Roles()).To(Equal([]*schema.Role{{Name: "Member"}}))
 				})
 			})
 		})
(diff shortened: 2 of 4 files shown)


cloud/keystone_test.go
```diff
@@ -368,19 +373,52 @@ var _ = Describe("Keystone client", func() {
...
 				Expect(auth.TenantName()).To(Equal("admin-project"))
 				Expect(auth.DomainID()).To(Equal("default"))
 				Expect(auth.DomainName()).To(Equal("default"))
 				Expect(auth.IsAdmin()).To(BeTrue())
 			})
+
+			It("Should reject admin token with multiple roles", func() {
+				multipleRoles := []interface{}{
+					map[string]interface{}{
+						"id":   "51cc68287d524c759f47c811e6463340",
```
jkmar, line 392: I've seen those roles in `keystone_test_utils.go`.
Can they be extracted to a function, for example:
```go
func getMemberRole() map[string]interface{} {
    return map[string]interface{
        "id": "51cc68287d524c759f47c811e6463340",
        "name": "Member",
    }
}
func getAdminRole() map[string]interface{} {
    return map[string]interface{
        "id": "7f0ea059b6d84029b60c18169d3c1d9a",
        "name": "admin",
    }
}
```
This would also be helpful to give context about the ids, since currently it's tricky to tell why they have such specific values.

cloud/keystone.go
```diff
@@ -228,6 +232,15 @@ func isTokenScopedToAdminProject(result *v3tokens.GetResult) bool {
 	return (err == nil) && s.IsAdminProject
 }
 
+func (client *keystoneV3Client) validateRolesForAdmin(roleIDs []string) error {
+	if len(roleIDs) != 1 || roleIDs[0] != schema.AdminRole {
+		client.updateCounter(1, "error.token.mismatched_role_admin")
+		return fmt.Errorf("Admin token should have only a single admin role instead of '%v'", roleIDs)
```
jkmar, line 238: nit: errors should start from small letter

cloud/keystone.go
```diff
@@ -228,6 +232,15 @@ func isTokenScopedToAdminProject(result *v3tokens.GetResult) bool {
 	return (err == nil) && s.IsAdminProject
 }
 
+func (client *keystoneV3Client) validateRolesForAdmin(roleIDs []string) error {
+	if len(roleIDs) != 1 || roleIDs[0] != schema.AdminRole {
+		client.updateCounter(1, "error.token.mismatched_role_admin")
```
jkmar, line 237: Is it useful to have metric for such a specific event?

cloud/keystone_test.go
```diff
@@ -368,19 +373,52 @@ var _ = Describe("Keystone client", func() {
...
+
 				Expect(err).To(BeNil())
 				Expect(auth.TenantID()).To(Equal("admin-project-id"))
 				Expect(auth.TenantName()).To(Equal("admin-project"))
 				Expect(auth.DomainID()).To(Equal("default"))
 				Expect(auth.DomainName()).To(Equal("default"))
 				Expect(auth.IsAdmin()).To(BeTrue())
 			})
+
+			It("Should reject admin token with multiple roles", func() {
```
jkmar, line 389: nit: the following 2 tests can be written as `DescribeTable`

cloud/keystone_test.go
```diff
@@ -368,19 +373,52 @@ var _ = Describe("Keystone client", func() {
...
+					},
+					map[string]interface{}{
+						"id":   "7f0ea059b6d84029b60c18169d3c1d9a",
+						"name": "admin",
+					},
+				}
+				setResponses(getV3TokensAdminResponseWithRoles(multipleRoles))
+				setupV3Client()
+
+				_, err := client.VerifyToken(token)
```
jkmar, line 403: nit: this could be one lined:
```go
Expect(client.VerifyToken(token)).NotTo(Succeed())
```

cierpuchaw: Unfortunately it cannot: `Unexpected non-nil/non-zero extra argument at index 1:`

cloud/keystone_test.go
cierpuchaw: I actually agree., especially since I added a third one :) Done.

cloud/keystone.go
cierpuchaw: It doesn't hurt. There are already multiple different metrics for different failure cases.

cloud/keystone.go
cierpuchaw: Probably they should but this file already uses capital letters for them, I wanted to be consistent.

cloud/keystone_test.go
cierpuchaw: I actually taken then from there but didn't really finish cleaning up. Done.

Regarding their IDs: I have no idea, they were added way before. I suspect they were simply copied from some Keystone installation to serve as examples.

cloud/keystone.go
```diff
@@ -228,6 +232,15 @@ func isTokenScopedToAdminProject(result *v3tokens.GetResult) bool {
 	return (err == nil) && s.IsAdminProject
 }
 
+func (client *keystoneV3Client) validateRolesForAdmin(roleIDs []string) error {
+	if len(roleIDs) != 1 || roleIDs[0] != schema.AdminRole {
```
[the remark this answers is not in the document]
cierpuchaw: I was sure that golang has short-circuit logical expression evaluation (and thankfully it does) but apparently that's not really well know :/

https://kuree.gitbooks.io/the-go-programming-language-report/content/22/text.html

 BTW I added a test case to prove this works.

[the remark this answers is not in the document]
jkmar: Sorry, about that comment, I've apparently got confused,
if you look at the statement I've proposed it is equivalent to the current one,
but when I was writing that comment I thought it was different,
maybe it was easier for me to understand it with just one negation.
> golang has short-circuit logical expression evaluation

I knew that and actually this is true for most of the languages.

cierpuchaw:
@yoichi-cloudwan -san a gentle reminder about this PR.

cierpuchaw:
@yoichi-cloudwan a gentle reminder about this PR.

yoichi-cloudwan (approved):

cloudwanbot (approved):
STATUS: OK Please check the result of code analysis: http://10.135.224.43:9000/dashboard?id=gohan_validate_roles_for_admin_token

PR #787, opened 2020-10-08
CW-1782: fully recursively populate default values.

cierpuchaw:
Middle level objects do not always have a default value but their
children might. Use DefaultMask (as it's fully computed) instead of
only Default in order to handle such cases.

### DIFF
--- a/schema/resource_test.go
+++ b/schema/resource_test.go
@@ -133,6 +133,10 @@ var _ = Describe("Resources", func() {
 		networkRed := NewResource(networkSchema, networkRedObj)
 		Expect(networkRed.PopulateDefaults()).To(Succeed())
 		expectedConfig := map[string]interface{}{
+			"default_vlan": map[string]interface{}{
+				"name":    "default_vlan",
+				"vlan_id": float64(1),
+			},
 			"empty_vlan": map[string]interface{}{},
 			"vpn_vlan": map[string]interface{}{
 				"name": "vpn_vlan",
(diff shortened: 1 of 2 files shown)


cloudwanbot (approved):
STATUS: OK Please check the result of code analysis: http://10.135.224.43:9000/dashboard?id=gohan_CW-1782_fixed_recursive_default_value_handling

cierpuchaw:
This is probably the most important piece of context for this PR: https://github.com/cloudwan/gohan/blob/master/schema/property.go#L231

If schema subobject contains a "default:" clause then it's used for the value of DefaultMask. Otherwise, if the subobject is an object (instead of an array or a "primitive" type) we recursively cache default values all the way down.

schema/resource_test.go
```diff
@@ -133,6 +133,10 @@ var _ = Describe("Resources", func() {
 		networkRed := NewResource(networkSchema, networkRedObj)
 		Expect(networkRed.PopulateDefaults()).To(Succeed())
 		expectedConfig := map[string]interface{}{
+			"default_vlan": map[string]interface{}{
```
jkmar, line 136: nit: this map appears twice in this pr, so maybe create a `var` for it

jkmar (approved):

schema/resource_test.go
cierpuchaw: Unfortunately both are not exactly the same. I could extract this small part into a function, but I'm not sure it would be an improvement.

yoichi-cloudwan (approved):

